PR #13, opened 2025-05-27
Allow RWX page mappings

vapourismo:
Closes RV-646

Migrated from [this GitLab MR](https://gitlab.com/tezos/tezos/-/merge_requests/18052).

# What

This allows RWX page mappings.

# Why

We previously disallowed them because writable pages are hard to track in terms of block cache validity. However, Jstz now needs them.

# How

Running instructions from writable and executable memory must not make it into the block cache. We might want to remove the block cache from the state at some point (tbc). Therefore, items whose validity is hard to track (like instructions from writable memory) can't be in caches like these.

# Manually Testing

```
make -C src/riscv all
```

# Benchmarking

Removing the conditional configuration attribute came with a -1.8% performance hit. I needed to change the representation of `Permissions` into a struct with three booleans. With that, the performance is unchanged.

# Tasks for the Author

- [x] Put in reasonable effort to ensure that CI will pass.
  - `make -C src/riscv`
  - `dune build src/lib_riscv`
  - `dune build src/rust_deps`
- [x] Benchmark performance and populate the table above if needed.
- [x] Explain changes to regression test captures when applicable.


felixp72 (dismissed):

src/riscv/lib/src/machine_state.rs
```diff
@@ -963,196 +968,101 @@ mod tests {
...
-                    rs2: nz::ra.into(),
-                    ..TaggedArgs::DEFAULT
-                },
-            })
-            .unwrap(),
+    // This test exercises the "partial block" feature of the block cache which is needed to ensure
+    // divergence of the state machine does not occur when the instructions in memory are changed
+    // but the state is rebound.
+    backend_test!(test_block_cache_partial_determinism, F, {
+        const CODE1: [u32; 4] = [
```
kurtisc, line 975: Why can't we use the TaggedInstruction mechanism any more?

kurtisc: My thought is that you've had to add comments to explain something that was previously tied to the code (what are the instructions and their arguments).

Distraction from the objective of the test is a good point, but that leads me to wonder how we could have both a distraction-free test and having the information about the instructions as part of the code. That leads me right back to where it was before: the function isn't made to be tested, it's made to take an opcode and arguments and produce an instruction. It's already a minimal interface for what I want. The mechanism for reducing distraction is just a function.

Presumably these literals came out of either copying output of our own code or from a calculation involving the RISC-V spec and hexadecimal arithmetic. Well, I have machine that's really good at that, but I'm not.

Ideally the test wouldn't actually *be* checking the parsed block, it would be using the output of something done at compile-time since we know it should succeed, but not enough of this is `const` and Rust doesn't have much explicit differentiation for the human anyway. Also, I'm not sure if it's necessarily `TaggedInstruction` that I want, `Instruction` might suffice.

vapourismo: I understand what you mean now. You don't trust the comments because they cannot be tested. 

I have added the instruction comparison [here](https://github.com/tezos/riscv-pvm/compare/9134747d34e5f72321aaa4f2a788d125df9e53af..22795397a3dad33e0713d19877e434b2307952ff). I needed to use `TaggedInstruction`s for comparison, though.

Additionally, I added a few more lines to explain the test [here](https://github.com/tezos/riscv-pvm/compare/22795397a3dad33e0713d19877e434b2307952ff..fb9b7f13618e18cfae7c4b96843be7c3b77db8eb). Hopefully, that makes it a bit more obvious that the instructions of each block aren't that important. What's important is that they differ - this is tested.

kurtisc (approved):

PR #17, opened 2025-05-27
Notify team members directly instead of using the team only

vapourismo:
Small fix to ensure everybody is notified on Pull Requests.

Adding only @tezos/risc-v-pvm-developers is interpreted as "one of this team". So, unfortunately, we need to mention everybody individually.


felixp72:
Username(s) checkout 🚀

PR #18, opened 2025-05-27
Remove unused artefacts

vapourismo:
Removes all of the Octez source tree that we don't need. This is effectively everything minus the `src/riscv` plus `src/riscv/api`.

This doesn't move the `src/riscv` workspace just yet.

For your convenience, the changes are split into two commits:
1. Remove API crate from RISC-V workspace
2. Remove everything but the RISC-V workspace


jobjo (approved):
Spring cleaning!

kurtisc (approved):

PR #20, opened 2025-05-28
Verify proof deserialisation for streaming bytes

felixp72:
Part of RV-631

Based on #22 & #15 

# What

Adds 2 more tests similar to `test_jstz_proofs` to check the deserialisation of the streaming implementation of the new deserialiser from #22 

# Why

Validates the deserialisation of proofs.

# How

Parametrised the `jstz_proofs` tests by the verification function as well. Added `verify_proof_by_`

# Manually Testing

```
cargo test --test test_proofs -- --nocapture --ignored
```

# Benchmarking

N/A

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/stepper/pvm.rs
+++ b/src/riscv/lib/src/stepper/pvm.rs
@@ -36,6 +36,7 @@ use crate::state_backend::FnManagerIdent;
 use crate::state_backend::ManagerBase;
 use crate::state_backend::ManagerReadWrite;
 use crate::state_backend::ProofLayout;
+use crate::state_backend::ProofPart;
 use crate::state_backend::ProofTree;
 use crate::state_backend::Ref;
 use crate::state_backend::hash::Hash;
@@ -43,6 +44,8 @@ use crate::state_backend::owned_backend::Owned;
 use crate::state_backend::proof_backend::ProofGen;
 use crate::state_backend::proof_backend::proof::Proof;
 use crate::state_backend::proof_backend::proof::deserialise_owned;
+use crate::state_backend::proof_backend::proof::deserialise_stream::{self};
+use crate::state_backend::proof_backend::proof::serialise_merkle_tree;
 use crate::state_backend::verify_backend::ProofVerificationFailure;
 use crate::state_backend::verify_backend::Verifier;
 use crate::state_backend::verify_backend::handle_stepper_panics;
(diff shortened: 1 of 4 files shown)


src/riscv/lib/src/stepper/pvm.rs
```diff
@@ -272,6 +275,22 @@ impl<'hooks, MC: MemoryConfig, CL: CacheLayouts, M: ManagerReadWrite>
         }
     }
 
+    /// Similar to [`PvmStepper::verify_proof`] but constructs the allocated space by using the raw deserialisation.
+    ///
+    /// Useful for testing the stream deserialisation.
+    /// Similar to [`PvmStepper::verify_proof`] but constructs the allocated space by using the raw deserialisation
```
victor-dumitrescu, line 281:
```suggestion
```

victor-dumitrescu:
I think the test for constructing a `Verifier` backend from proof bytes is a bit contrived. A proof is generated and serialised, then we continue to verify the initial proof in both the `proof -> Verifier` case and the `bytes -> Verifier` case. In the latter, the proof is then serialised again. What do you think about exposing a `verify_serialised_proof` function instead in the PVM stepper which takes the proof bytes directly? This would also correspond more neatly to the API which will be offered through `deserialise_proof`.

victor-dumitrescu (changes requested):

kurtisc (approved):

vapourismo:
There is a CI failure that looks awefully like an out-of-memory situation :(

vapourismo:
> There is a CI failure that looks awefully like an out-of-memory situation :(

Indeed, running `make test-long` peaks at 17.3081207275 GiB of residential memory on my machine. Compare that to 10.5374298096 GiB in the target branch.

vapourismo:
> > Indeed, running make test-long peaks at 17.3081207275 GiB of residential memory on my machine. Compare that to 10.5374298096 GiB in the target branch.
> 
> And this PR does not even work with the new `to_verifier_alloc` which also constructs a `MerkleProof` as a side effect. Not looking good...

Could you check if it is perhaps just a matter of more tests that could potentially run in parallel? Check what happens when you run only 1 test at a time.

felixp72:
> I think the test for constructing a Verifier backend from proof bytes is a bit contrived. A proof is generated and serialised, then we continue to verify the initial proof in both the proof -> Verifier case and the bytes -> Verifier case. In the latter, the proof is then serialised again. What do you think about exposing a verify_serialised_proof function instead in the PVM stepper which takes the proof bytes directly? This would also correspond more neatly to the API which will be offered through deserialise_proof.

I'd like to do that, but If I do not have a `Proof` type, then I can not call `verify_proof_internal` which uses a `ProofTree` - which in turn currently not obtainable from the raw bytes. I have that functionality as a part of #48 , I'm in the process of creating a separate PR for it.

victor-dumitrescu (approved):

PR #39, opened 2025-06-02
Fix Restore/Save operations of Rustup and Cargo cache

vapourismo:
I made a mistake in the configuration for the "Rustup and Cargo cache" portion of the CI workflow. `lookup-only` doesn't do what I thought it did.

This leads to mistakenly creating too many large Rustup and Cargo cache entries. GitHub repositories have a 10 GiB cache limit. With a cache created for each branch, we're significantly above the limit, which results in GitHub gradually deleting some caches. This may occasionally delete the `main` branch cache.

This PR aims to address this by splitting the workflow steps into "restore" and "save" and only enabling the save step on the `main` branch.

See [this workflow run](https://github.com/tezos/riscv-pvm/actions/runs/15403851177/job/43342449600?pr=39) for a demonstration that the cache is no longer produced for PR runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,9 +70,9 @@ jobs:
             nix-${{ runner.os }}-${{ runner.arch }}-
           save: ${{ github.ref_name == 'main' }}
 
-      - name: Rustup and Cargo cache
+      - name: Restore Rustup and Cargo cache
         id: rust-cache
-        uses: actions/cache@v4.2.3
+        uses: actions/cache/restore@v4.2.3
         with:
           key: rust-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('**/Cargo.lock', '**/rust-toolchain*') }}-v1
           restore-keys: |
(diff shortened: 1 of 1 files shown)


victor-dumitrescu (approved):

.github/workflows/ci.yml
```diff
@@ -95,7 +94,7 @@ jobs:
       # All downloads should be cached. If there is a cache miss, we should ensure that the cache
       # contains all downloads.
       - name: Prefetch Cargo downloads
-        if: steps.rust-cache.outputs.cache-hit != 'true' && github.ref_name == 'main' 
+        if: github.ref_name == 'main' && steps.rust-cache.outputs.cache-hit != 'true'
```
NSant215, line 97: is this for readability or does it change something more about execution?

NSant215 (approved):

.github/workflows/ci.yml
vapourismo: Oh! I missed that. I undid it in [this change](https://github.com/tezos/riscv-pvm/compare/4b65200181d9a38a5f0ac9a2b2396fb3272f0607..ffc2cdf6942726581855c72bbd5e83d0a88edd1f).

PR #42, opened 2025-06-04
RISC-V: Implement `handle_getpid`

kurtisc:
Closes RV-666 😈

# What
Implements `handle_getpid`. As there is only one process, this always returns `Ok(1)`.

# Why

# How
There's only one process, so we always return 1
# Manually Testing

```
make -C src/riscv all
```

# Benchmarking
No expected impact.

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/pvm/linux.rs
+++ b/src/riscv/lib/src/pvm/linux.rs
@@ -89,6 +89,9 @@ const RT_SIGACTION: u64 = 134;
 /// System call number for `rt_sigprocmask` on RISC-V
 const RT_SIGPROCMASK: u64 = 135;
 
+/// System call number for `getpid` on RISC-V
+const GETPID: u64 = 172;
+
 /// System call number for `brk` on RISC-V
 const BRK: u64 = 214;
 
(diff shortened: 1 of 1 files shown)


src/riscv/lib/src/pvm/linux.rs
```diff
@@ -872,6 +876,14 @@ impl<M: ManagerBase> SupervisorState<M> {
         })
     }
 
+    /// Handle `getpid` system call.
+    ///
+    /// See: <https://man7.org/linux/man-pages/man2/getpid.2.html>
+    pub(super) fn handle_getpid(&mut self) -> Result<u64, Infallible> {
```
vapourismo, line 882:
```suggestion
    pub(super) fn handle_getpid(&self) -> Result<u64, Infallible> {
```

kurtisc: Good call, [done](https://github.com/tezos/riscv-pvm/compare/92f8519bf199e08d08a48ebda3ff5b767d585446..e0bd475d83f2082c3a278a2d9b460ad290069df0)

PR #43, opened 2025-06-04
RISC-V: Implement `close` and `read` system call handlers

kurtisc:
Closes RV-667
# What
Implements system call handlers for `close` and `read`
 - `close`
   - We don't allow opening files, so this always results in an error
 - `read`
   - We disallow this unless the length is zero.

# Why
These are unimplemented system call handlers that jstz uses with V8

# How
For now, just return error for the reasons given.

I'm currently trying to see which file(s) jstz is trying to read from. It may be necessary to mock some data. But building it takes _hours_

# Manually Testing

```
make -C src/riscv all
```

# Benchmarking
No change needed

# Tasks for the Author


src/riscv/lib/src/pvm/linux/fs.rs
```diff
@@ -23,6 +23,32 @@ impl<M: ManagerBase> SupervisorState<M> {
         Err(Error::Access)
     }
 
+    /// Handle `close` system call. All access to the file system is denied.
+    ///
+    /// See <https://man7.org/linux/man-pages/man2/close.2.html>
+    pub(super) fn handle_close(&mut self) -> Result<u64, Error>
```
vapourismo, line 29:
```suggestion
    pub(super) fn handle_close(&self) -> Result<u64, Error>
```

src/riscv/lib/src/pvm/linux/fs.rs
```diff
@@ -23,6 +23,32 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+        M: ManagerReadWrite,
+    {
+        // None of the file descriptors we allow (stdout/stderr) can sensibly be closed.
+        Err(Error::Access)
+    }
+
+    /// Handle `read` system call. All access to the file system is denied.
+    ///
+    /// See <https://man7.org/linux/man-pages/man2/read.2.html>
+    pub(super) fn handle_read(&mut self, length: u64) -> Result<u64, Error>
```
vapourismo, line 40:
```suggestion
    pub(super) fn handle_read(&self, length: u64) -> Result<u64, Error>
```

src/riscv/lib/src/pvm/linux/fs.rs
```diff
@@ -23,6 +23,32 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+    where
+        M: ManagerReadWrite,
+    {
+        if length == 0 {
+            // If the length is zero then the standard allows returning zero without reading.
+            return Ok(0);
+        }
```
vapourismo, line 47: This may be surprising to some, despite the reference to the man page. Could you add a small note explaining that POSIX allows omission of error checking if the length is zero?

vapourismo (changes requested):
Overall looks good to me. Just two minor things:
- Remove unnecessary `mut`-ness
- Include a reference to POSIX, allowing the omission of error checking

NSant215 (approved):

src/riscv/lib/src/pvm/linux/fs.rs
kurtisc: [done](https://github.com/tezos/riscv-pvm/compare/3d161914190de68abd7b0c09061834fd5b3b2d47..5330cc7bbc96be075f90f28a79d5ecbf275c5f32)

src/riscv/lib/src/pvm/linux/fs.rs
kurtisc: [done](https://github.com/tezos/riscv-pvm/compare/3d161914190de68abd7b0c09061834fd5b3b2d47..5330cc7bbc96be075f90f28a79d5ecbf275c5f32)

PR #44, opened 2025-06-04
RISC-V: Fix reservation set implementation

NSant215:
Closes RV-674.

# What   

Changes reservation-set alignment to align-down rather than up. 

# Why

Currently, our implementation of recording the starting address of the reservation set is incorrect because we 'align up' any addresses rather than 'aligning down'. For instance, our reservation set size is 8 bytes, so an LR.W rd, 4 should reserve bytes 0 to 7, but the starting address recorded would be 8. This means that an SC.D rd, 0 would fail, when it should pass.

# How

# Manually Testing

```
make -C src/riscv all
```

# Benchmarking

| $MyMachine | X.XXX TPS | X.XXX TPS | XX.XX% |
| Benchmark Machine | Y.YYY TPS | Y.YYY TPS | YY.YY% |

# Regressions

# Tasks for the Author

- [ ] Benchmark performance and [populate the section above](#benchmarking) if needed.


src/riscv/lib/src/machine_state/reservation_set.rs
```diff
@@ -36,16 +36,13 @@ pub type ReservationSetLayout = backend::Atom<u64>;
 /// data word or doubleword. [...] The Unix platform is expected to require of
 /// main memory that the reservation set be of fixed size, contiguous, naturally
 /// aligned, and no greater than the virtual memory page size."
-const SIZE: u64 = 8;
+pub(crate) const RESERVATION_SET_SIZE: u64 = 8;
```
vapourismo, line 39: I think you can simplify a few things by expressing the reservation set size in terms of a bit mask. This means you only need to bitwise-and the mask with the address to achieve alignment.

vapourismo (changes requested):

src/riscv/lib/src/machine_state/reservation_set.rs
NSant215: cool idea :D - done [here](https://github.com/tezos/riscv-pvm/compare/e8d832f..f20e881fef5faf9a76529e41738122070d34db14)

src/riscv/lib/src/machine_state/reservation_set.rs
```diff
@@ -36,18 +38,13 @@ pub type ReservationSetLayout = backend::Atom<u64>;
 /// data word or doubleword. [...] The Unix platform is expected to require of
 /// main memory that the reservation set be of fixed size, contiguous, naturally
 /// aligned, and no greater than the virtual memory page size."
-const SIZE: u64 = 8;
+///
+/// With an 8-byte reservation set, we can align addresses by bitwise ANDing with
+/// the bitmask.
+pub(crate) const RES_SET_BITMASK: u64 = 0xFFFFFFF8;
```
vapourismo, line 44: Nit: Could you use an expression that makes it easier to spot what the bits are that have not been set? For example `!0b111`.

vapourismo (approved):

src/riscv/lib/src/machine_state/reservation_set.rs
vapourismo: Feel free to resolve this

NSant215: ah cool done [here](https://github.com/tezos/riscv-pvm/compare/f20e881fef5faf9a76529e41738122070d34db14..7372b73d287d315593c242a68222074531296094)

kurtisc (approved):

PR #45, opened 2025-06-04
RISC-V: Implement `sched_getaffinity` system call handler

kurtisc:
Closes RV-672

# What
Implements `sched_getaffinity`. As there is only one hart, this always returns or sets a single bit in the out parameter.

# Why

# How

We only have one hart, so there is only one bit that can be set. We can't control the arguments this is called with, so we still need support for dealing with interesting arguments.

# Manually Testing

```
make -C src/riscv all
```

# Benchmarking

No change expected

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/pvm/linux.rs
+++ b/src/riscv/lib/src/pvm/linux.rs
@@ -110,6 +110,9 @@ const GETRANDOM: u64 = 278;
 /// System call number for `clock_gettime` on RISC-V
 const CLOCK_GETTIME: u64 = 113;
 
+/// System call number for `sched_getaffinity` on RISC-V
+const SCHED_GETAFFINITY: u64 = 123;
+
 /// System call number for `gettimeofday` on RISC-V
 const GETTIMEOFDAY: u64 = 169;
 
@@ -720,6 +723,7 @@ impl<M: ManagerBase> SupervisorState<M> {
             MADVISE => dispatch0!(madvise),
             GETRANDOM => dispatch2!(getrandom, core),
             CLOCK_GETTIME => dispatch2!(clock_gettime, core),
+            SCHED_GETAFFINITY => dispatch3!(sched_getaffinity, core),
             GETTIMEOFDAY => dispatch2!(gettimeofday, core),
             SBI_FIRMWARE_TEZOS => return on_tezos(core),
             _ => Err(Error::NoSystemCall),
@@ -896,6 +900,37 @@ impl<M: ManagerBase> SupervisorState<M> {
 
         // Return 0 as an indicator of success
         Ok(0)
+    }
+
+    /// Handle `sched_getaffinity` system call. We only support one hart.
+    ///
+    /// See: <https://man7.org/linux/man-pages/man2/sched_getaffinity.2.html>
+    fn handle_sched_getaffinity(
+        &self,
+        core: &mut MachineCoreState<impl MemoryConfig, M>,
+        _pid: parameters::ProcessId,
+        cpusetsize: parameters::CpuSetSize,
+        mask: VirtAddr,
+    ) -> Result<u64, Error>
+    where
+        M: ManagerReadWrite,
+    {
+        const SINGLE_PROCESS_PID_AFFINITY: u8 = 0b1_u8;
+
+        let other_bytes_size = cpusetsize.0.get() - 1;
+
+        for i in 0..other_bytes_size {
+            let address = mask + i;
+            core.main_memory.write(address.to_machine_address(), 0u8)?;
+        }
+
+        core.main_memory.write(
+            (mask + other_bytes_size).to_machine_address(),
+            SINGLE_PROCESS_PID_AFFINITY,
+        )?;
+
+        // Return bytes written
+        Ok(cpusetsize.0.get())
     }
 
     /// Handle `gettimeofday` system call. Fills the timeval and timezone structures with zeros.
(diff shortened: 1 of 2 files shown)


src/riscv/lib/src/pvm/linux.rs
```diff
@@ -898,6 +903,41 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+
+        if cpusetsize.get() > 1 {
+            let other_bytes = vec![0_u8; cpusetsize.get() as usize - 1];
+
+            core.main_memory
+                .write_all(mask.0.to_machine_address(), &other_bytes)?;
+        }
+
+        core.main_memory.write(
+            mask.0.to_machine_address() + cpusetsize.get() - 1,
```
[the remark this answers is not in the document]
felixp72: Maybe-a-nit: This whole address computation (`mask.0.to_machine_address() + cpusetsize.get() - 1,`) can be moved to a method on `CpuSet::to_machine_address(cpusetsize: NonZeroU64)`. This would also eliminate the need of the `pub` on the mask's `0` field. Bonus: It's used in two places.

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -898,6 +903,41 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+
+        if pid.0 == 0 {
+            return Ok(SINGLE_PROCESS_PID_AFFINITY as u64);
+        }
+
+        if cpusetsize.get() > 1 {
+            let other_bytes = vec![0_u8; cpusetsize.get() as usize - 1];
+
+            core.main_memory
+                .write_all(mask.0.to_machine_address(), &other_bytes)?;
```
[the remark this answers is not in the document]
kurtisc: - I get your point about the proof size, but I'm not sure if this is everything required to bound what we write

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -898,6 +902,50 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+                remaining -= 8;
+                let address = mask.0 + remaining;
+                core.main_memory.write(address.to_machine_address(), 0u64)?;
+            }
+
+            for i in 0..remaining {
+                let address = mask.0 + i;
+                core.main_memory.write(address.to_machine_address(), 0u8)?;
+            }
+        }
```
vapourismo, line 938: I think zero-ing code can be simplified. This might be a performance loss, though, this function isn't called often. Hence simplicity might be preferrable.

```suggestion
        for i in 1..cpusetsize.0.get() {
            let address = mask.0 + i;
            core.main_memory.write(address.to_machine_address(), 0u8)?;
        }
```

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -75,6 +76,57 @@ impl TryFrom<u64> for MainThreadId {
...
+
+impl TryFrom<u64> for NullableProcessId {
+    type Error = Error;
+
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        // We only support the single hart or a null value
+        match value {
+            0 | 1 => Ok(NullableProcessId(value)),
+            _ => Err(Error::Search),
+        }
```
vapourismo, line 92: I think `NullableProcessId` deserves to be an enum.

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -75,6 +76,57 @@ impl TryFrom<u64> for MainThreadId {
...
+    type Error = Error;
+
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        /// <https://github.com/torvalds/linux/blob/0ff41df1cb268fc69e703a08a57ee14ae967d0ca/arch/x86/Kconfig#L1033C58-L1034C27>
+        const MAX_CPU_SET_SIZE: u64 = 8192 / u8::BITS as u64;
+        if value > MAX_CPU_SET_SIZE {
+            return Err(Error::Search);
+        }
+
+        Ok(CpuSetSize(NonZeroU64::new(value).ok_or(Error::Search)?))
```
vapourismo, line 112: Why `Search` and not `Invalid`?

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -898,6 +902,50 @@ impl<M: ManagerBase> SupervisorState<M> {
...
 
+    /// Handle `sched_getaffinity` system call. We only support one hart.
+    ///
+    /// See: <https://man7.org/linux/man-pages/man2/sched_getaffinity.2.html>
+    fn handle_sched_getaffinity(
+        &self,
+        core: &mut MachineCoreState<impl MemoryConfig, M>,
+        pid: parameters::NullableProcessId,
+        cpusetsize: parameters::CpuSetSize,
+        mask: parameters::CpuSet,
```
vapourismo, line 913: This could be `VirtAddr` instead. I don't see what `CpuSet` adds here as a type.

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -898,6 +902,50 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+            }
+
+            for i in 0..remaining {
+                let address = mask.0 + i;
+                core.main_memory.write(address.to_machine_address(), 0u8)?;
+            }
+        }
+
+        core.main_memory.write(
+            mask.0.to_machine_address() + other_bytes_size,
```
vapourismo, line 941: Please compute the virtual address before converting to machine address in order to respect the virtual address arithmetic rules.

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -898,6 +902,50 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+            SINGLE_PROCESS_PID_AFFINITY,
+        )?;
+
+        // Return 0 as an indicator of success
+        Ok(0)
```
vapourismo, line 946: From the man page, I understand that the kernel would return the number of bytes written to mask. What is the reason for returning 0 for success here?

vapourismo (changes requested):

src/riscv/lib/src/pvm/linux/parameters.rs
kurtisc: [done](https://github.com/tezos/riscv-pvm/compare/342c2aa0cfe34c47632e4a8f1407fd90cfbe17fa..188f0093e0763ee7079b0d9ccbb30ab8a30b3cc1)

src/riscv/lib/src/pvm/linux/parameters.rs
kurtisc: You're right, the manpage says this should be EINVAL, [done](https://github.com/tezos/riscv-pvm/compare/342c2aa0cfe34c47632e4a8f1407fd90cfbe17fa..188f0093e0763ee7079b0d9ccbb30ab8a30b3cc1)

vapourismo (changes requested):

src/riscv/lib/src/pvm/linux/parameters.rs
kurtisc: whoops, [also changed](https://github.com/tezos/riscv-pvm/compare/188f0093e0763ee7079b0d9ccbb30ab8a30b3cc1..4d73374abff739a521315ebcb62aee93feecff74)

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -75,6 +75,49 @@ impl TryFrom<u64> for MainThreadId {
     }
 }
 
+/// Known to be a valid process ID. As we currently only support one hart, this will be that hart,
+/// or zero to represent no hart.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum NullableProcessId {
+    Null,
+    One,
```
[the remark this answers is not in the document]
kurtisc: Makes sense, [done](https://github.com/tezos/riscv-pvm/compare/188f0093e0763ee7079b0d9ccbb30ab8a30b3cc1..4d73374abff739a521315ebcb62aee93feecff74)

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -898,6 +902,41 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+        mask: VirtAddr,
+    ) -> Result<u64, Error>
+    where
+        M: ManagerReadWrite,
+    {
+        const SINGLE_PROCESS_PID_AFFINITY: u8 = 0b1_u8;
+
+        if pid == parameters::NullableProcessId::CallingThread {
+            return Ok(SINGLE_PROCESS_PID_AFFINITY as u64);
+        }
```
vapourismo, line 922: I can't find this behaviour in the man page. Doesn't this also need to write to the mask?

kurtisc: Yes, this was a misreading of something to do with the *set* counterpart

[changes](
https://github.com/tezos/riscv-pvm/compare/4d73374abff739a521315ebcb62aee93feecff74..bc23aa06342228d003a6ed77d11f2f41da73e233)

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1139,6 +1178,220 @@ mod tests {
...
+        machine_state
+            .hart
+            .xregisters
+            .write(registers::a2, mask_address.to_machine_address());
+
+        // Set mask address
+        machine_state
+            .hart
+            .xregisters
+            .write(registers::a2, mask_address.to_machine_address());
```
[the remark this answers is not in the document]
kurtisc: [No](https://github.com/tezos/riscv-pvm/compare/bc23aa06342228d003a6ed77d11f2f41da73e233..ba80669e8315a7afcffcaaf34c9ec862e14c6530), nice catch!

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -75,6 +75,45 @@ impl TryFrom<u64> for MainThreadId {
...
+#[derive(Debug, Clone, Copy)]
+pub struct ProcessId;
+
+impl TryFrom<u64> for ProcessId {
+    type Error = Error;
+
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        // We only support the single hart or a null value
+        match value {
+            1 => Ok(ProcessId),
```
vapourismo, line 89: The type `ProcessId` is used in places where `0` is a valid argument. I suppose in our current implementation, `0` and `1` both mean the "main" process/thread.

kurtisc: I see what you mean. [Updated](https://github.com/tezos/riscv-pvm/compare/ba80669e8315a7afcffcaaf34c9ec862e14c6530..41c9aba2990d12b0d10ef41443c5ad21430ac01e), with a cheeky method of testing it without duplicating tests.

vapourismo (approved):

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1139,6 +1174,182 @@ mod tests {
...
+
+                assert_eq!(leading_mask, 0u8);
+            }
+
+            let end_mask = machine_state
+                .main_memory
+                .read::<u8>((mask_address + i - 1).to_machine_address())
+                .unwrap();
+
+            assert_eq!(end_mask, 0b1_u8);
```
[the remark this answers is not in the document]
kurtisc: [Conceded](https://github.com/tezos/riscv-pvm/compare/41c9aba2990d12b0d10ef41443c5ad21430ac01e..10b256b8fb3ccd2cca3531397b257abe067e801c)

victor-dumitrescu (approved):

PR #52, opened 2025-06-05
Lower Divu, Divw, Divuw

HantangSun:
Closes RV-675

# What

Lowers Opcode Divu, Divw, Divuw

# Why

Part of the effort to enable internal opcodes to be runnable in both interpreted mode and jitted mode

# How

- Added suppport for unsigned division in ICB
- Move implementation of Divu, Divw, Divuw to integer.rs
- Renamed the opcodes according to the [opcode redesign](https://docs.google.com/spreadsheets/d/1ehIkV6O3i0mIcAgsU_P8qHVjWqdGwv34f3IWWJhYC9Y/edit?gid=772829846#gid=772829846)
- Enabled the opcode for jit by adding its case to fn to_lowering

# Manually Testing

```
make -C src/riscv all
```

# Benchmarking

| M2 MBP| 15.841 TPS | 15.666 TPS | -1.10% |
| Benchmark Machine | 11.283 TPS | 11.186 TPS | -0.86% |

# Regressions

Regression caused by changing the Div* opcodes to to take NonZeroXRegisters instead of XRegister by splitting on the zero case. This changes the argument shape and therefore changes the state hash.

# Tasks for the Author


src/riscv/lib/src/machine_state/instruction/constructors.rs
```diff
@@ -2257,6 +2311,36 @@ impl Instruction {
         }
     }
 
+    pub(super) fn from_ic_divu(args: &RTypeArgs) -> Instruction {
```
NSant215, line 2320: as you have done correctly here, can you fix the splitting above as well for `Div` to only split on rd, as well as copy the comment there about `non-trapping` for all cases? It is useful info to know that we can be assured that `rd=x0` are no-ops due to all of these being non-trapping.

NSant215 (approved):

HantangSun:
> > I observed consistent slowdown on the benchmark machine. I am looking into potential reasons for it
> 
> Had you checked if the master you tested against was the same as the one this branch was rebased off?

Yes, and the performance regression persisted through the rebase

kurtisc (changes requested):
@HantangSun 
>I observed consistent slowdown on the benchmark machine. I am looking into potential reasons for it

Did you find anything?

HantangSun:
> @HantangSun
> 
> > I observed consistent slowdown on the benchmark machine. I am looking into potential reasons for it
> 
> Did you find anything?

The div* opcodes are actually not used frequently enough to appear in the flame graph. I guess we need to accept that sometimes minor performance regression does happen

kurtisc (approved):

PR #54, opened 2025-06-06
RISC-V: Implement `set_robust_list` system call handler

kurtisc:
Closes RV-673

# What

Implements `set_robust_list`. As there is currently only one thread,
this does nothing but check the size parameter, then returns error or
success.

# Why

# How
We currently only support one thread, so this only checks parameters and returns error or success.

# Manually Testing

```
make -C src/riscv all
```

# Benchmarking

No change expected

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/pvm/linux.rs
+++ b/src/riscv/lib/src/pvm/linux.rs
@@ -790,17 +790,15 @@ impl<M: ManagerBase> SupervisorState<M> {
     /// Handle `set_robust_list` system call.
     ///
     /// See: <https://www.man7.org/linux/man-pages/man2/set_robust_list.2.html>
-    fn handle_set_robust_list(&self, _head: VirtAddr, size: usize) -> Result<u64, Error> {
+    fn handle_set_robust_list(
+        &self,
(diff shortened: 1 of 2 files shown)


src/riscv/lib/src/pvm/linux.rs
```diff
@@ -775,6 +779,22 @@ impl<M: ManagerBase> SupervisorState<M> {
         Ok(MAIN_THREAD_ID)
     }
 
+    /// Handle `set_robust_list` system call.
+    ///
+    /// See: <https://www.man7.org/linux/man-pages/man2/set_robust_list.2.html>
+    fn handle_set_robust_list(&self, _head: VirtAddr, _size: usize) -> Result<u64, Error>
+    where
+        M: ManagerRead + ManagerWrite,
```
felixp72, line 787: Do you need these bounds?

felixp72 (changes requested):

src/riscv/lib/src/pvm/linux.rs
kurtisc: No, good call

PR #62, opened 2025-06-09
RISC-V: Implement `prlimit64` system call handle_prlimit64

kurtisc:
Closes RV-671

# What

Implements `prlimit64`. Following the limits set with this system call is undesirable and non-mandatory, so this is just a stub.

# Why

# How

jstz only uses this to set the limits for the stack size. Privileged processes (read: we) don't need to follow these limits - and we don't want to. So this is just a stub implementation.

jstz calls:
```
prlimit64(0, RLIMIT_STACK, NULL, {rlim_cur=8192*1024, rlim_max=RLIM64_INFINITY}) = 0
``` 

The manpage notes:
>[A privileged process (under Linux: one with the CAP_SYS_RESOURCE capability) may make arbitrary changes to either limit value. ](https://man7.org/linux/man-pages/man2/prlimit.2.html)

# Manually Testing

```
make -C src/riscv all
```

# Benchmarking

No change expected

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/pvm/linux.rs
+++ b/src/riscv/lib/src/pvm/linux.rs
@@ -131,6 +131,9 @@ const SCHED_GETAFFINITY: u64 = 123;
 /// System call number for `gettimeofday` on RISC-V
 const GETTIMEOFDAY: u64 = 169;
 
+/// System call number for `prlimit64` on RISC-V
+const PRLIMIT64: u64 = 261;
+
 /// Key into the auxiliary vector which informs supervised processes of auxiliary information
 #[derive(Clone, Copy)]
 #[repr(u64)]
@@ -749,6 +752,7 @@ impl<M: ManagerBase> SupervisorState<M> {
             CLOCK_GETTIME => dispatch2!(clock_gettime, &mut machine.core),
             SCHED_GETAFFINITY => dispatch3!(sched_getaffinity, &mut machine.core),
             GETTIMEOFDAY => dispatch2!(gettimeofday, &mut machine.core),
+            PRLIMIT64 => dispatch4!(prlimit64, &mut machine.core),
             SBI_FIRMWARE_TEZOS => return on_tezos(&mut machine.core),
             _ => Err(Error::NoSystemCall),
         };
@@ -1011,6 +1015,88 @@ impl<M: ManagerBase> SupervisorState<M> {
         if tz != 0 {
             core.main_memory.write(tz, [0u8; TIMEZONE_SIZE])?;
         }
+
+        // Return 0 as an indicator of success
+        Ok(0)
+    }
+
+    /// Handle `prlimit64` system call.
+    ///
+    /// See: <https://www.man7.org/linux/man-pages/man2/prlimit64.2.html>
+    fn handle_prlimit64<MC: MemoryConfig>(
+        &self,
+        core: &mut MachineCoreState<MC, M>,
+        _pid: parameters::ProcessId,
+        resource: parameters::Rlimit,
+        new_limit: VirtAddr,
+        old_limit: VirtAddr,
+    ) -> Result<u64, Error>
+    where
+        M: ManagerReadWrite,
+    {
+        // In jstz this is only used to set the maximum size of the process stack to infinity. In
+        // other programs, it can be used to set limits for system resources. Allowing programs to
+        // set these values to extreme values would inflate proof size, so is undesirable.
+        // Futhermore, a privileged process (such as the supervisor implemented in this PVM) has
+        // the ability to set these limits to arbitrary values.
+        // So it does! By using its own predefined limits and returning EPERM.
+
+        struct MemorySize<MC>(MC);
+        impl<MC: MemoryConfig> MemorySize<MC> {
+            const SIZE: u64 = MC::TOTAL_BYTES as u64;
+
+            const UPPER_BOUND: () = assert!(MC::TOTAL_BYTES <= u64::MAX as usize);
+
+            // Hard limit on the size of the data segment
+            const RLIMIT_DATA: u64 = Self::SIZE;
+
+            // Hard limit on resident set size. Not used.
+            const RLIMIT_RSS: u64 = Self::SIZE;
+
+            // Hard limit on the memory that may be locked into RAM (B)
+            const RLIMIT_MEMLOCK: u64 = Self::SIZE;
+
+            // Hard limit on the size of a process's virtual memory (B)
+            const RLIMIT_AS: u64 = Self::SIZE;
+        }
+
+        let () = MemorySize::<MC>::UPPER_BOUND;
+
+        // If new_limit is not NULL, the system call is being used to write new limits, so ignore
+        // it and return a permissions error
+        if new_limit != 0 {
+            return Err(Error::NotPermitted);
+        }
+
+        // If the system call is used to read, return any current limit
+        let limit = match resource {
+            parameters::Rlimit::Cpu => parameters::RLIMIT_CPU,
+            parameters::Rlimit::Fsize => parameters::RLIMIT_FSIZE,
+            parameters::Rlimit::Data => MemorySize::<MC>::RLIMIT_DATA,
+
+            // Hard limit on the size of the process stack
+            parameters::Rlimit::Stack => STACK_SIZE,
+
+            parameters::Rlimit::Core => parameters::RLIMIT_CORE,
+            parameters::Rlimit::Rss => MemorySize::<MC>::RLIMIT_RSS,
+            parameters::Rlimit::Nproc => parameters::RLIMIT_NPROC,
+            parameters::Rlimit::Nofile => parameters::RLIMIT_NOFILE,
+            parameters::Rlimit::Memlock => MemorySize::<MC>::RLIMIT_MEMLOCK,
+            parameters::Rlimit::As => MemorySize::<MC>::RLIMIT_AS,
+            parameters::Rlimit::Locks => parameters::RLIMIT_LOCKS,
+            parameters::Rlimit::Sigpending => parameters::RLIMIT_SIGPENDING,
+            parameters::Rlimit::Msgqueue => parameters::RLIMIT_MSGQUEUE,
+            parameters::Rlimit::Nice => parameters::RLIMIT_NICE,
+            parameters::Rlimit::Rtprio => parameters::RLIMIT_RTPRIO,
+            parameters::Rlimit::Rttime => parameters::RLIMIT_RTTIME,
+        };
+
+        const RLIM_T_SIZE: u64 = 8;
+
+        core.main_memory
+            .write(old_limit.to_machine_address(), limit)?;
+        core.main_memory
+            .write((old_limit + RLIM_T_SIZE).to_machine_address(), limit)?;
 
         // Return 0 as an indicator of success
         Ok(0)
--- a/src/riscv/lib/src/pvm/linux/parameters.rs
+++ b/src/riscv/lib/src/pvm/linux/parameters.rs
@@ -92,6 +92,142 @@ impl TryFrom<u64> for ProcessId {
     }
 }
 
+/// Hard limit on CPU time (s)
+pub(crate) const RLIMIT_CPU: u64 = u64::MAX;
+
+/// Hard limit for maximum file size (we don't support writing to a filesystem)
+pub(crate) const RLIMIT_FSIZE: u64 = 0;
+
+/// Hard limit for core dumps (we don't support core dumps)
+pub(crate) const RLIMIT_CORE: u64 = 0;
+
+/// Hard limit on processes (we only support one)
+pub(crate) const RLIMIT_NPROC: u64 = 1;
+
+/// Hard limit on the number of file descriptors that a system call can work with
+///
+/// We also use this constant to implictly limit how much memory can be associated with a system
+/// call. For example, `ppoll` takes a pointer to an array of `struct pollfd`. If we don't limit
+/// the length of that array, then we might read an arbitrary amount of memory. This impacts the
+/// proof size dramatically as everything read would also be in the proof.
+pub const RLIMIT_NOFILE: u64 = 512;
+
+/// Hard limit on the number of flock locks plus fcntl locks. Unused by us (and we only support one hart and no filesystem)
+pub(crate) const RLIMIT_LOCKS: u64 = 0;
+
+/// Hard limit on the number of signals for a user ID. Signals are not currently implemented.
+pub(crate) const RLIMIT_SIGPENDING: u64 = 0;
+
+/// Hard limit on the number of bytes that may be allocated for message queues. Message queues are not currently implemented.
+pub(crate) const RLIMIT_MSGQUEUE: u64 = 0;
+
+/// Hard limit on the niceness of a process. This is inverted and the actual niceness is 20 - the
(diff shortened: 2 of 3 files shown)


src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1009,6 +1013,21 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+    /// Handle `prlimit64` system call.
+    ///
+    /// See: <https://www.man7.org/linux/man-pages/man2/prlimit64.2.html>
+    fn handle_prlimit64(&self) -> Result<u64, Error> {
+        // In jstz this is only used to set the maximum size of the process stack to infinity. In
+        // other programs, it can be used to set limits for system resources. Allowing programs to
+        // set these values to extreme values would inflate proof size, so is undesirable.
+        // Futhermore, a privileged process (such as the supervisor implemented in this PVM) has
+        // the ability to set these limits to arbitrary values.
+        // So it does! By ignoring this system call, and using its own predefined limits.
```
vapourismo, line 1026: What about other uses of `prlimit64`? What if this system call is used for reading?

It is not enough to assume that all kernels will only ever do what one version of Jstz does. So either you filter and allow only one use, or you try to support all uses. Since there are only 16 resource types, the latter should be viable as well.

vapourismo (changes requested):

kurtisc:
I've changed this to allow reading from some constant limits.

This needed some changes for the dispatch macros, and I may split that work out into a separate PR. The limits will presumably have to vary based on the `MemoryConfig`, so this will be necessary work either way.

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -77,16 +79,155 @@ impl TryFrom<u64> for MainThreadId {
...
+        }
+
+        let () = MemorySize::<MC>::LOWER_BOUND;
+        let () = MemorySize::<MC>::UPPER_BOUND;
+
+        match value {
+            0..=15 => {
+                let limit = match value {
+                    // Hard limit on CPU time (s)
+                    0 => RLIMIT_CPU,
```
felixp72, line 181: I feel like you could do away with the nested match statements and get rid of the `unreachable`, unless I'm missing a code path from your initial logic?

```suggestion
        let limit = match value {
            // Hard limit on CPU time (s)
            0 => RLIMIT_CPU,
            // other 1..15 cases
            _ => Err(Error::Search)
        };
        Ok(Rlimit::<MC>(limit, PhantomData)
```

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -77,16 +79,155 @@ impl TryFrom<u64> for MainThreadId {
...
+            const RLIMIT_STACK: u32 = Self::SIZE;
+
+            // Hard limit on resident set size. Not used.
+            const RLIMIT_RSS: u32 = Self::SIZE;
+
+            // Hard limit on the memory that may be locked into RAM (B)
+            const RLIMIT_MEMLOCK: u32 = Self::SIZE;
+
+            // Hard limit on the size of a process's virtual memory (B)
+            const RLIMIT_AS: u32 = Self::SIZE;
```
felixp72, line 171: a `u32` will mean a process will only have ~4GB. Will we expect in the future to support more than 4GB? - even though this is per process so the total physical memory will/can be bigger anyway.

felixp72 (changes requested):

src/riscv/lib/src/pvm/linux/parameters.rs
kurtisc: Done but the diff includes a rebase

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -77,21 +79,155 @@ impl TryFrom<u64> for MainThreadId {
...
+
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        let value = value.try_into()?;
+
+        struct MemorySize<MC>(MC);
+        impl<MC: MemoryConfig> MemorySize<MC> {
+            const SIZE: u64 = MC::TOTAL_BYTES as u64;
+
+            const LOWER_BOUND: () = assert!(MC::TOTAL_BYTES >= u64::MIN as usize);
+            const UPPER_BOUND: () = assert!(MC::TOTAL_BYTES <= u64::MAX as usize);
```
vapourismo, line 156: Out of curiousity, do these actually evaluate at compile time, or are the expressions simply inlined?

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -77,21 +79,155 @@ impl TryFrom<u64> for MainThreadId {
...
+pub struct ProcessId(pub u64);
 
 impl TryFrom<u64> for ProcessId {
     type Error = Error;
 
     fn try_from(value: u64) -> Result<Self, Self::Error> {
         // We only support the single hart or a null value
         match value {
-            0 | 1 => Ok(ProcessId),
+            id @ 0 | id @ 1 => Ok(ProcessId(id)),
```
vapourismo, line 91: Does it make sense to make `ProcessId` into an enum?

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1009,6 +1078,60 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+    }
+
+    /// Handle `prlimit64` system call.
+    ///
+    /// See: <https://www.man7.org/linux/man-pages/man2/prlimit64.2.html>
+    fn handle_prlimit64(
+        &self,
+        core: &mut MachineCoreState<impl MemoryConfig, M>,
+        pid: parameters::ProcessId,
+        resource: parameters::Rlimit<impl MemoryConfig>,
```
vapourismo, line 1105: The two memory configs are not required to be the same here. Bind them to a dedicated single type variable `MC: MemoryConfig`.

vapourismo (changes requested):

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1009,6 +1078,60 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+        // In jstz this is only used to set the maximum size of the process stack to infinity. In
+        // other programs, it can be used to set limits for system resources. Allowing programs to
+        // set these values to extreme values would inflate proof size, so is undesirable.
+        // Futhermore, a privileged process (such as the supervisor implemented in this PVM) has
+        // the ability to set these limits to arbitrary values.
+        // So it does! By ignoring this system call, and using its own predefined limits.
+
+        if pid.0 != 0 {
+            return Err(Error::Access);
+        }
```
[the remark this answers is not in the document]
kurtisc: >The pid argument specifies the ID of the process on which the call
       is to operate.  If pid is 0, then the call applies to the calling
       process.  To set or get the resources of a process other than
       itself, the caller must have the CAP_SYS_RESOURCE capability in
       the user namespace of the process whose resource limits are being
       changed, or the real, effective, and saved set user IDs of the
       target process must match the real user ID of the caller and the
       real, effective, and saved set group IDs of the target process
       must match the real group ID of the caller.
       
I'm interpreting this as correct usage being that non zero PIDs are for other processes. Having said that, I think the second clause can be assumed to be true (without having to implement users/groups) as the target process will be itself. So I've opted to roll this back

src/riscv/lib/src/pvm/linux/parameters.rs
kurtisc: [Rolled back](https://github.com/tezos/riscv-pvm/compare/2e0b4ff35d5ef07bcabf4ec7f16a29f50089669c..8d2bd931cc69b3f764ffed9995a2097c5c222662#diff-5fe78c11847d325ddd58c78d62b53e62c43efe6e61c8789761c223e2d6f725b7L83)

src/riscv/lib/src/pvm/linux.rs
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/2e0b4ff35d5ef07bcabf4ec7f16a29f50089669c..8d2bd931cc69b3f764ffed9995a2097c5c222662#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7R1017)

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -77,21 +79,155 @@ impl TryFrom<u64> for MainThreadId {
...
+/// rlimit. We only support one hart
+const RLIMIT_NICE: u64 = 20;
+
+/// Hard limit on the real time priority of a process. We only support one hart.
+const RLIMIT_RTPRIO: u64 = 0;
+
+/// Hard limit on CPU time without making a blocking system call. We only support one hart.
+const RLIMIT_RTTIME: u64 = u64::MAX;
+
+pub struct Rlimit<MC: MemoryConfig>(pub u64, PhantomData<MC>);
```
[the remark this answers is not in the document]
kurtisc: This makes sense, thanks. MC in parameters felt like a code smell.

[Done](https://github.com/tezos/riscv-pvm/compare/2e0b4ff35d5ef07bcabf4ec7f16a29f50089669c..8d2bd931cc69b3f764ffed9995a2097c5c222662#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7R1017)

felixp72 (approved):

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1009,6 +1013,101 @@ impl<M: ManagerBase> SupervisorState<M> {
         // Return 0 as an indicator of success
         Ok(0)
     }
+
+    fn write_rlimit<MC: MemoryConfig>(
+        core: &mut MachineCoreState<impl MemoryConfig, M>,
```
vapourismo, line 1018:
```suggestion
        core: &mut MachineCoreState<MC, M>,
```

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1009,6 +1013,101 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+        resource: parameters::Rlimit,
+    ) -> Result<(), Error>
+    where
+        M: ManagerReadWrite,
+    {
+        struct MemorySize<MC>(MC);
+        impl<MC: MemoryConfig> MemorySize<MC> {
+            const SIZE: u64 = MC::TOTAL_BYTES as u64;
+
+            const LOWER_BOUND: () = assert!(MC::TOTAL_BYTES >= u64::MIN as usize);
```
vapourismo, line 1029: What is the purpose of the lower bound? It cannot be false regardless of the value you pick for `TOTAL_BYTES`.

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1009,6 +1013,101 @@ impl<M: ManagerBase> SupervisorState<M> {
         // Return 0 as an indicator of success
         Ok(0)
     }
+
+    fn write_rlimit<MC: MemoryConfig>(
```
vapourismo, line 1017: Usually having unassociated functions in an `impl` block is because you want to share type variables. This isn't really leveraged in this case though.

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1009,6 +1013,101 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+            parameters::Rlimit::Sigpending => parameters::RLIMIT_SIGPENDING,
+            parameters::Rlimit::Msgqueue => parameters::RLIMIT_MSGQUEUE,
+            parameters::Rlimit::Nice => parameters::RLIMIT_NICE,
+            parameters::Rlimit::Rtprio => parameters::RLIMIT_RTPRIO,
+            parameters::Rlimit::Rttime => parameters::RLIMIT_RTTIME,
+        };
+
+        core.main_memory.write(addr.to_machine_address(), limit)?;
+        core.main_memory.write(
+            (addr + size_of_val(&limit) as u64).to_machine_address(),
```
vapourismo, line 1072: Do you mind using 8 instead? Or a constant for the size of `rlim_t` like some of the other system call handlers. I would like to avoid using host constants in PVM semantics.

```suggestion
            (addr + 8).to_machine_address(),
```

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1009,6 +1013,101 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+        // set these values to extreme values would inflate proof size, so is undesirable.
+        // Futhermore, a privileged process (such as the supervisor implemented in this PVM) has
+        // the ability to set these limits to arbitrary values.
+        // So it does! By ignoring this system call, and using its own predefined limits.
+
+        // If new_limit is not NULL, the system call is being used to write new limits, so ignore
+        // it
+        if new_limit != 0 {
+            return Ok(0);
+        }
```
vapourismo, line 1103: Instead of silently ignoring it, we should probably fail here to avoid entering an inconsistent state. Perhaps you can return `EPERM`?

vapourismo (changes requested):

src/riscv/lib/src/pvm/linux.rs
kurtisc: >The calling process did not have permission set limits for the process specified by pid.

Sure, no problem. [Done](https://github.com/tezos/riscv-pvm/compare/8d2bd931cc69b3f764ffed9995a2097c5c222662..562da1975bb4d5e68f80708247922f9c7ff56912#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7R1065).

src/riscv/lib/src/pvm/linux.rs
vapourismo: That's also good.

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -1009,6 +1013,88 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+        if new_limit != 0 {
+            return Err(Error::NotPermitted);
+        }
+
+        // If the system call is used to read, return any current limit
+        let limit = match resource {
+            parameters::Rlimit::Cpu => parameters::RLIMIT_CPU,
+            parameters::Rlimit::Fsize => parameters::RLIMIT_FSIZE,
+            parameters::Rlimit::Data => MemorySize::<MC>::RLIMIT_DATA,
+            parameters::Rlimit::Stack => MemorySize::<MC>::RLIMIT_STACK,
```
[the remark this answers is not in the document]
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/562da1975bb4d5e68f80708247922f9c7ff56912..ead640d1584f2721aaabc2ba77b35b316b75c627), but I've made it assigned to the `RLIMIT_STACK` constant just to keep the documentation consistent.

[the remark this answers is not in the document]
vapourismo: `RLIMIT_STACK` is only accessible with a `MC` now.

vapourismo (approved):

kurtisc:
Base changed until `ole/rv-686` is merged

vapourismo:
@kurtisc Don't forget about this one :)

PR #65, opened 2025-06-09
Organise the block cache into configuration and state

vapourismo:
# What

Refactors the block cache into separate configuration and state modules.

# Why

My largest motivation behind this change is the separation of concerns. It makes several things easier going forward regarding how the block cache will evolve.

Refactoring the block cache layout trait into a configuration trait allows us to reuse the underlying state layout more flexibly. Currently, a dedicated `Layout` implementation is required, which reimplements an existing layout. We achieve reuse by associating the layout with the trait implementation instead of making the trait implementation be the layout as well.

Before this change, the `BlockCacheLayout` was used to describe both the layout and the underlying functionality. That meant all operational code would work with the layout trait instead of an operation-focused trait. The new `BlockCache` trait addresses this - it separates its concerns from the layout/configuration trait and focuses entirely on what the block cache can do.

The refactor now provides an abstract interface for the block cache, rather than its entries.

Extracting a configuration instance from the block cache makes two things easier:
- Bundling the PVM configuration as one object (e.g. a type)
- Moving away from a configuration altogether

The reason we have these single-implementation traits in the first place is the lack of const-generic evaluation. For example, we can't deal with the `BITS` and `SIZE` constants effectively.

Turning an `impl<...> ... for BlockCache` into an `impl<...> BlockCache` is the change we would need to make to achieve the transition off the single-implementation traits. This may happen for other reasons, not just the const-generic evaluation becoming stable.

Some of these points may seem contradictory. Here are some scenarios to consider:
- The block cache is decoupled from the PVM state → No more `Layout` dependency
- The block cache becomes dynamically sized → Singular or no `Layout`, and no configuration or state trait needed
- The block cache becomes a permanent part of PVM → Need solid configurability of the PVM, hence requires good integration of the block cache configuration into the PVM-level configuration

# How

Unfortunately, this PR is quite big. Seemingly, the GH embedded diff viewer is good at detecting moved chunks within a file, but not across files 😢 

For reviewing convenience, I have split it into 3 separate changes:

1. Moving of the block cache implementations. This moves items from the `block_cache` into `block_cache::state` and `block_cache::config`. The implementations of items are unchanged. However, namespaces, imports and documentation needed to be adjusted. The same commit also introduces the `BlockCacheConfig` and `BlockCache` traits, both are defined in `block_cache`.
2. Rename all the type variables and block cache configuration references. Generally, `CL` -> `BCC` and `*CacheLayout` -> `*CacheConfig`.
3. Delete now redundant code.

# Manually Testing

```
make -C src/riscv all
```

# Benchmarking

This does not impact runtime performance.

# Tasks for the Author


src/riscv/lib/src/machine_state/block_cache.rs
```diff
@@ -175,722 +164,6 @@ impl<'a, MC: MemoryConfig, M: ManagerReadWrite> From<&'a Instruction> for ICall<
     }
 }
 
-/// Layout for an address cell
-pub struct AddressCellLayout;
```
NSant215, line 179: how come we can remove `AddressCellLayout` and replace it with just the allocated type `Cell<Address, M>` in `Cache`

vapourismo: `AddressCellLayout` was previously used to control the initial address value (`!0`). However, since then we have decoupled initialisation from the layout. It is now done via `NewState`. That means `AddressCellLayout` and `Atom<Address>` (allocates as `Cell<Address, M>`) are equivalent layouts.

src/riscv/lib/src/machine_state/block_cache/config.rs
```diff
@@ -0,0 +1,140 @@
...
+use crate::state_backend::AllocatedOf;
+use crate::state_backend::Atom;
+use crate::state_backend::FnManager;
+use crate::state_backend::ManagerBase;
+use crate::state_backend::ManagerReadWrite;
+use crate::state_backend::Many;
+use crate::state_backend::Ref;
+
+/// Configuration for a block cache
+pub struct BlockCacheConfig<const BITS: usize, const SIZE: usize>;
```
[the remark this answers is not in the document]
vapourismo: Good point! Change [here](https://github.com/tezos/riscv-pvm/compare/6c9dc7ce4f1a85d765287b5721a47902a690d036..4c62177b1ecd8921d107bd674edb924286875cbf).

victor-dumitrescu (approved):

PR #71, opened 2025-06-11
RISC-V: Optimize cranelift configuration

HantangSun:
Closes RV-681

# What

Modify configuration for cranelift 

# Why

Configuration for cranelift used in jit compilation can be optimised. This MR brings performance improvement of 1.5%

# Manually Testing

```
make -C src/riscv all
```

# Benchmarking

| M2 MBP | 16.517TPS | 16.867 TPS | 2.12% |
| Benchmark Machine | 11.748TPS | 11.894 TPS | 1.24% |

# Regressions

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/jit.rs
+++ b/src/riscv/lib/src/jit.rs
@@ -115,8 +115,10 @@ impl<MC: MemoryConfig, M: JitStateAccess> JIT<MC, M> {
         }
 
         let mut flag_builder = settings::builder();
-        flag_builder.set("use_colocated_libcalls", "false")?;
-        flag_builder.set("is_pic", "false")?;
+
+        // Optimization level for generated code. Configured to Generate the fastest possible code
+        // https://docs.wasmtime.dev/api/cranelift/prelude/settings/struct.Flags.html#method.opt_level
+        flag_builder.set("opt_level", "speed")?;
 
         let isa_builder = cranelift_native::builder().map_err(JitError::UnsupportedPlatform)?;
         let isa = isa_builder.finish(settings::Flags::new(flag_builder))?;


src/riscv/lib/src/jit.rs
```diff
@@ -115,8 +115,9 @@ impl<MC: MemoryConfig, M: JitStateAccess> JIT<MC, M> {
         }
 
         let mut flag_builder = settings::builder();
-        flag_builder.set("use_colocated_libcalls", "false")?;
-        flag_builder.set("is_pic", "false")?;
+        flag_builder.set("opt_level", "speed")?;
+        flag_builder.set("enable_verifier", "false")?;
```
vapourismo, line 119: Does turning off the verifier have negative consequences?

src/riscv/lib/src/jit.rs
```diff
@@ -115,8 +115,9 @@ impl<MC: MemoryConfig, M: JitStateAccess> JIT<MC, M> {
...
+        flag_builder.set("enable_heap_access_spectre_mitigation", "false")?;
```
vapourismo, line 120: Please leave this on.

src/riscv/lib/src/jit.rs
```diff
@@ -115,8 +115,9 @@ impl<MC: MemoryConfig, M: JitStateAccess> JIT<MC, M> {
         }
 
         let mut flag_builder = settings::builder();
-        flag_builder.set("use_colocated_libcalls", "false")?;
-        flag_builder.set("is_pic", "false")?;
+        flag_builder.set("opt_level", "speed")?;
```
vapourismo, line 118: Are there potentially fewer guarantees on the semantics of instructions with this?

vapourismo (changes requested):

vapourismo:
Thanks for looking into tweaking these knobs. Any performance improvement is always welcome, of course. I wonder whether you have looked into the consequences of these changes, apart from the improved performance, because I see no evidence of this in the description or the proposed changes.

src/riscv/lib/src/jit.rs
HantangSun: The documentation says "pub fn [enable_verifier] (&self) -> [bool]
Run the Cranelift IR verifier at strategic times during compilation.

This makes compilation slower but catches many bugs. The verifier is always enabled by default, which is useful during development."

I think given that our jit IR is thoroughly tested, this can be disabled

src/riscv/lib/src/jit.rs
HantangSun: Changed [here](https://github.com/tezos/riscv-pvm/compare/bf19327c8056090f8dae6edba0870ba54c20b606..47cd52831251bebec2ea9f14f6a64434cdeb993f)

NSant215 (commented):
I think these changes look fine but I think it would be good to document each one with a reason why and possibly with a link to the flag docs. I also can't seem to find what the default setting is for each flag? could you point me to this?

src/riscv/lib/src/jit.rs
vapourismo: Could you expand on this? How does PIC have fewer guarantees? `opt_level=speed` means none of the IR semantics change, everything will be complete safe and there are no surprises, no pitfalls that we expose ourselves to?

src/riscv/lib/src/jit.rs
```diff
@@ -115,8 +115,8 @@ impl<MC: MemoryConfig, M: JitStateAccess> JIT<MC, M> {
...
+        flag_builder.set("enable_verifier", "false")?;
```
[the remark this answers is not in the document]
HantangSun: The method to configure setting for cranelift codegen is ```fn set(&mut self, name: &str, value: &str) -> SetResult<()>``` so the values actually has to be string literals

src/riscv/lib/src/jit.rs
HantangSun: So "PIC" (position-independent code) means that the code can be executed regardless of its address, which would have more guarantee. This MR removes the code that disabled this flag. And actually https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/src/settings.rs#L512 shows that this flag is set to false by default(The deleted line wasn't needed in the first place).

For opt_level=speed, there aren't any documentation on exactly what optimisations are performed. We would need to really look into cranelift source code to better understand. 

I feel that the regression test with outline-jit should provide a good enough safety net for us to make the change

PR #93, opened 2025-06-17
RISC-V: host crypto for ethereum

edwin1729:
# What

Add system calls for secp256k1 based verification and keccak hashing, used by ethereum

# Why

Massively improve performance of the revm kernel (which lives in this [fork](https://github.com/edwin1729/riscv-pvm/pull/2))

# How
Implementation follows the example set by ed25519, using crates libsecp256k1 and sha3 (same as etherlink)

# Manually Testing

There are no tests for this change (there are no regressions when it's used in my fork though)\
See https://github.com/edwin1729/riscv-pvm/pull/2
```
make -C src/riscv all
```

# Benchmarking

| 13th Gen Intel(R) Core(TM) i7-13700HX | 11.621 TPS | 11.416 TPS | - |
| Benchmark Machine | 11.523 TPS | 11.851 TPS | - |

# Tasks for the Author


src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -20,6 +23,14 @@ use tezos_smart_rollup_constants::riscv::SBI_TEZOS_INBOX_NEXT;
 use tezos_smart_rollup_constants::riscv::SBI_TEZOS_REVEAL;
 use tezos_smart_rollup_constants::riscv::SbiError;
 
+// temporary till added kernel_sdk constants
+
+/// Function ID for `sbi_tezos_secp256k1_verify`
+pub const SBI_TEZOS_SECP256K1_VERIFY: u64 = 0x0a;
+
+/// Function ID for `sbi_tezos_keccak_hash256`
+pub const SBI_TEZOS_KECCAK_HASH256: u64 = 0x0b;
```
victor-dumitrescu, line 32: Will the host Keccak-256 function be used other than for hashing a message in order to verify its signature? If not, I'd argue the host verify function could take the message bytes directly and hash and verify on the host side.

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -20,6 +23,14 @@ use tezos_smart_rollup_constants::riscv::SBI_TEZOS_INBOX_NEXT;
 use tezos_smart_rollup_constants::riscv::SBI_TEZOS_REVEAL;
 use tezos_smart_rollup_constants::riscv::SbiError;
 
+// temporary till added kernel_sdk constants
+
+/// Function ID for `sbi_tezos_secp256k1_verify`
+pub const SBI_TEZOS_SECP256K1_VERIFY: u64 = 0x0a;
+
+/// Function ID for `sbi_tezos_keccak_hash256`
```
victor-dumitrescu, line 31: The function is called Keccak-256 so I think the names/comments should reflect that here (e.g., `SBI_TEZOS_KECCAK256_HASH`) and in the rest of the code.

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -20,6 +23,14 @@ use tezos_smart_rollup_constants::riscv::SBI_TEZOS_INBOX_NEXT;
 use tezos_smart_rollup_constants::riscv::SBI_TEZOS_REVEAL;
 use tezos_smart_rollup_constants::riscv::SbiError;
 
+// temporary till added kernel_sdk constants
```
victor-dumitrescu, line 26: Turn this into TODO comments for each of the constants.

vapourismo: In fact, raise Linear issues for all and mention them in the `TODO` comment.

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -249,6 +260,59 @@ where
...
+where
+    MC: MemoryConfig,
+    M: ManagerReadWrite,
+{
+    let arg_out_addr = machine.hart.xregisters.read(a0);
+    let arg_msg_addr = machine.hart.xregisters.read(a1);
+    let arg_msg_len = machine.hart.xregisters.read(a2);
+
+    let mut msg_bytes = vec![0u8; arg_msg_len as usize];
+    machine.main_memory.read_all(arg_msg_addr, &mut msg_bytes)?;
```
vapourismo, line 307: In a PVM you must not read arbitrary number of bytes. This is because if you read X bytes in a step, the Merkle proof for that step will contain at least X bytes as well.

Seems like the Blake2B handler also doesn't respect this either - so we need to fix that there too.

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -10,6 +10,9 @@ use ed25519_dalek::Signature;
 use ed25519_dalek::Signer;
 use ed25519_dalek::SigningKey;
 use ed25519_dalek::VerifyingKey;
+
+use libsecp256k1::{Message, PublicKey, Signature as SecpSig};
+use sha3::{Digest, Keccak256};
```
vapourismo, line 15: Looks like [you need to run the formatter](https://github.com/tezos/riscv-pvm/actions/runs/15705636743/job/44252813366?pr=93#step:10:17). Try `scripts/format.sh --all`.

src/riscv/lib/src/pvm/tezos.rs
edwin1729: I don't see how else to make a hashing function. Is there some upper limit for the maximum number of bytes that we could decide? For my specific use case this `msg_bytes` would be constant length other than when it encodes a contract creating transaction, which can grow really large.

I can also just remove hashing (hash in kernel) for a performance loss of 5%.

vapourismo: If you limit it to 4096 bytes of payload, that would be fine. However, arbitrary size is not allowed.

src/riscv/lib/src/pvm/tezos.rs
victor-dumitrescu: I think it should not be included for now if it's not needed.

src/riscv/lib/src/pvm/tezos.rs
edwin1729: This should be resolved now. Revm kernel messages are never larger than 4096 bytes, so the limit doesn't affect the benchmark

src/riscv/lib/src/pvm/tezos.rs
edwin1729: It ended up being needed, since I only hash conditionally on the message length being less than 4096 bytes now. This is to limit proof sizes.

HantangSun (approved):

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -249,6 +267,58 @@ where
...
+) -> Result<u64, SbiError>
+where
+    MC: MemoryConfig,
+    M: ManagerReadWrite,
+{
+    let arg_out_addr = machine.hart.xregisters.read(a0);
+    let arg_msg_addr = machine.hart.xregisters.read(a1);
+    let arg_msg_len = machine.hart.xregisters.read(a2);
+
+    if (arg_msg_len > MAX_PVM_MEMORY_ACCESS as u64) {
```
vapourismo, line 308: Use the potentially larger integer to avoid truncation. Or alternatively, define `MAX_PVM_MEMORY_ACCESS` as a `u64`.

```suggestion
    if ((arg_msg_len as usize) > MAX_PVM_MEMORY_ACCESS) {
```

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -249,6 +267,58 @@ where
...
+        return Err(SbiError::InvalidParam);
+    }
+
+    let mut msg_bytes = vec![0u8; arg_msg_len as usize];
+    machine.main_memory.read_all(arg_msg_addr, &mut msg_bytes)?;
+    let hash: [u8; 32] = Keccak256::digest(&msg_bytes).into();
+    machine
+        .main_memory
+        .write_all(arg_out_addr, hash.as_slice())?;
```
vapourismo, line 317: Write the hash directly if possible.

```suggestion
        .write(arg_out_addr, hash)?;
```

vapourismo (changes requested):

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -249,6 +260,59 @@ where
...
+where
+    MC: MemoryConfig,
+    M: ManagerReadWrite,
+{
+    let arg_pk_addr = machine.hart.xregisters.read(a0);
+    let arg_sig_addr = machine.hart.xregisters.read(a1);
+    let arg_msg_addr = machine.hart.xregisters.read(a2);
+
+    let mut pk_bytes = [0u8; 65];
+    machine.main_memory.read_all(arg_pk_addr, &mut pk_bytes)?;
```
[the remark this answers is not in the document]
edwin1729: This and similar changes have been made

vapourismo (approved):

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -249,6 +267,58 @@ where
...
+
+    let pk = PublicKey::parse(&pk_bytes).map_err(|_| SbiError::Failed)?;
+    let sig = SecpSig::parse_standard(&sig_bytes).map_err(|_| SbiError::Failed)?;
+    let msg = Message::parse(&msg_bytes);
+    let valid = libsecp256k1::verify(&msg, &sig, &pk);
+
+    Ok(valid as u64)
+}
+
+/// Compute a Keccak 256-bit digest.
```
victor-dumitrescu, line 295:
```suggestion
/// Compute a Keccak-256 digest.
```

victor-dumitrescu (changes requested):
LGTM, but could you please squash the review suggestion commit?

edwin1729:
I'm not sure what [this](https://github.com/tezos/riscv-pvm/actions/runs/15761943800/job/44465719893?pr=93) error means, about the ssh keys

PR #94, opened 2025-06-17
Etherlink: Run kernel in RISC-V sandbox

victor-dumitrescu:
Closes RV-626.

# What

The kernel writes the values Etherlink expects to find in its durable storage in order to correctly run.

# How

The values are hardcoded and written to durable storage upon first entry. I considered using the kernel installer for this step in order to derive a configuration program from a YAML file but since one of the entries is too large to be set directly and would need to be revealed instead, adding the need to generate the preimages, I opted for this more straightforward approach.

# Manually Testing

```
make -C etherlink
cargo run -- run --input etherlink/target/riscv64gc-unknown-linux-musl/release/etherlink --address sr163Lv22CdE8QagCwf48PWDTquk6isQwv57 --inbox-file assets/etherlink-sample-inbox.json
```

A sample inbox is included in order to test correct execution. This will be removed and replaced with an inbox generator program and a benchmark in the following PR.

# Benchmarking

Not applicable.

# Tasks for the Author


src/riscv/etherlink/kernel/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
+hex = "0.4.3"
```
vapourismo, line 7: Would this be better suited in the workspace Cargo manifest?

src/riscv/etherlink/kernel/src/main.rs
```diff
@@ -3,10 +3,48 @@
...
+    ),
+    (
+        "/evm/world_state/eth_accounts/0000000000000000000000000000000000000000/ticket_table/fc2a0f0107d0e2d1b32bd9d99dd3da5bbda4edf016e64d5eec8d64310d3eeb88/f0affc80a5f69f4a9a3ee01a640873b6ba53e539",
+        "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF7F",
+    ),
+    (
+        "/tezlink/context/contracts/index/000002298c03ed7d454a101eb7022bc95f7e5f41ac78/balance",
+        "00",
+    ),
+];
```
vapourismo, line 35: Out of curiosity, have you had a look into using the installer config directly?

victor-dumitrescu: Yes, I first tried using the installer with a YAML file. The issue is that the `/evm/ticketer` value is longer than what the installer can set directly in the storage so it wants to reveal it instead, meaning we also need to generate and pass reveal preimages, which didn't seem worth it at this stage.

PR #104, opened 2025-06-20
Add support for reads/writes that skip checks using static information

kurtisc:
Closes RV-694

Reviewing each commit separately is recommended.

# What
Currently, reads and writes are performed with the assumption that an address could be out of alignment and out of bounds. In some cases, we can statically know that they're aligned or in-bounds, even without knowing the opaque values of the X64 values.

This adds support for:
 - Reading and writing using APIs that take aligned values
 - Skipping bounds checks for values known to be in bounds
 
It also changes `dyn_region_read_all` in the owned backend to use a method that assumes the addresses are non-overlapping. I think we can (and could always) assume this, since it's copying from a distinct area of memory. Unfortunately we can't skip the ub checks here, but they [always were assuming it was aligned!](https://doc.rust-lang.org/beta/src/core/intrinsics/mod.rs.html#3840)

A downside to this is that the X64 type, which implements Copy, has been extended with two enums.

# Why

Faster :)

# How

 - Adds Alignment and Bounded types to the X64 type. 
 - Splits the memory load C interface into distinct interfaces for unaligned + unbounded/aligned/bounded/aligned + bounded
 - Extends the Elem type
 - Splits execution of reading/writing based on constant information

# Manually Testing

```
make -C src/riscv all
```

# Regressions

None

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/instruction_context/value.rs
+++ b/src/riscv/lib/src/instruction_context/value.rs
@@ -133,7 +133,7 @@ impl PhiValue for XValue {
         params: &[ir::Value],
         _: &mut jit::builder::Builder<'_, MC, JSA>,
     ) -> Self::IcbValue<jit::builder::Builder<'a, MC, JSA>> {
-        jit::builder::X64(params[0])
+        jit::builder::X64(params[0], Default::default())
     }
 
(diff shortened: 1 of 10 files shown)


kurtisc:
Don't think this is working out on x64 unfortunately

PR #109, opened 2025-06-23
Use a variable to store the step counter in a block

vapourismo:
Closes RV-697

# What

Moves the step counter within a JIT function into a local variable.

# Why

This enables JIT functions to perform intra-block jumps.

# Manually Testing

```
make -C src/riscv all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/jit/builder.rs
+++ b/src/riscv/lib/src/jit/builder.rs
@@ -100,12 +100,14 @@ impl<'a, MC: MemoryConfig, JSA: JitStateAccess> Builder<'a, MC, JSA> {
         let result_ptr_val = builder.block_params(entry_block)[3];
         // last param ignored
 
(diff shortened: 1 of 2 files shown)


vapourismo:
> So this is functionally doing the same thing as before but is preparing for later changing to using a Variable instead of a static counter. Think change is fine but wondering if we will actually use an 'offset' when we move to instruction block land, as each instruction will interact with the variable directly I would think?

The concept of "offset" indeed changes a bit. It is basically just the number of steps accumulated since the last time we updated the steps variable.

Each instruction can't interact with the steps variable directly. They lack context to determine whether the instruction is a join point.

Consider the following example. Instruction I2 is the join point. However, it only knows this when I4 declared that it may conditionally branch back to I2.

image

This means we need an extra pass after the instruction blocks have been generated.

kurtisc (approved):

PR #120, opened 2025-06-26
Simplify Control and State Registers

vapourismo:
# What

This change simplifies the control and state registers structure by reducing it to the two fields needed: `fflags` and `frm`.

Other CSRs are not writable at the moment, so we return static values for them.

# Why

Removes unnecessary complexity.

# Manually Testing

```
make -C src/riscv all
```

# Regressions

The state layout for the PVM has changed due to the `CSRRegisterValues` structure being removed. This impacts state hashes and proof representations alike.

# Tasks for the Author


kurtisc (approved):

src/riscv/lib/src/pvm/common.rs
```diff
@@ -261,8 +261,10 @@ impl<MC: MemoryConfig, BCC: BlockCacheConfig, B: block::Block<MC, M>, M: state_b
         // We want to just slightly modify the state without interfering with normal execution.
         // Increasing the number of instructions executed is a good candidate.
```
victor-dumitrescu, line 262: This comment is now out of date.

src/riscv/lib/src/pvm/common.rs
```diff
@@ -261,8 +261,10 @@ impl<MC: MemoryConfig, BCC: BlockCacheConfig, B: block::Block<MC, M>, M: state_b
...
         let csregs = &mut self.machine_state.core.hart.csregisters;
-        let instrs_retired = csregs.read::<u64>(CSRegister::instret);
-        csregs.write(CSRegister::instret, instrs_retired + 1);
+
+        // `fflags` is a writable CSR
+        let fflags = csregs.read::<u64>(CSRegister::fflags);
+        csregs.write(CSRegister::fflags, fflags + 1);
```
victor-dumitrescu, line 267: Won't this potentially affect execution?

vapourismo: Changing any CSR can affect execution if a program reads the registers. Not sure how we can avoid that.

src/riscv/lib/src/pvm/common.rs
vapourismo: I removed the outdated line [here](https://github.com/tezos/riscv-pvm/compare/9ffdf80c46928fc0403a793ff14a611d4923ac49..872acbb30f8a00d6ceec85df641d883c47787ee4).

src/riscv/lib/src/pvm/common.rs
victor-dumitrescu: I thought the value of `fflags` might have an impact on floating-point instruction execution, but if that's not the case it should be OK.

PR #125, opened 2025-06-30
Restore inline annotations for CSR operations

vapourismo:
# What

Restores 2 `#[inline]` annotations that should not have been deleted.

# Why

Improves performance as those functions benefit a lot from inlining.

# Manually Testing

```
make -C src/riscv all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/machine_state/csregisters.rs
+++ b/src/riscv/lib/src/machine_state/csregisters.rs
@@ -177,6 +177,7 @@ pub struct CSRegisters<M: backend::ManagerBase> {
 
 impl<M: backend::ManagerBase> CSRegisters<M> {
     /// Write to a CSR.
+    #[inline]
     pub fn write<V: Bits64>(&mut self, reg: CSRegister, value: V)
     where
         M: backend::ManagerWrite,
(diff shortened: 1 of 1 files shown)


kurtisc:
I don't think this is really showing up in the benchmark. The mean is slightly better but it's less than the standard deviation and, if you look at the raw results, it seems like it's mostly due to the best run being an outlier.

A few thoughts:

- There's recently been a regression that has dropped the mean benchmark back under 12 TPS
- The variance and flakiness of the benchmark for 15 iterations is becoming larger than some small but meaningful improvements

Maybe it's worth increasing the number of iterations we use as the standard, especially now it's a mostly asynchronous thing we leave to CI (when it's working...), which would reduce the impact of the outliers and variance. I don't think we want to change to median, because we still want pathologically bad iterations to be a blocker.

- I've always been surprised by how many inline annotations we have and my prior assumption was that the compiler would be better at judging this than people. It also keeps its heuristics up-to-date with any changes in the codebase, whereas we could be working on a change where these annotations are a pessimisation and won't know unless we compare results for different inline annotations for each change. I also think it's something that could lead to bigger performance differences between Mac and the reference machine, since the instruction caching will be totally different. Is it really that bad that it's worth us (you) having to do all of this manually?

kurtisc (approved):

PR #138, opened 2025-07-07
Clean up ordering of trait implementations

vapourismo:
# What

Reorganise the trait implementations such that the module roughly follows this ordering:
- type definition
- `impl _` associated
- `impl _ for _` for the defined typed

# Why

There were some stray implementations grouped with unrelated types.

# Manually Testing

```
make -C src/riscv all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/state_backend/region.rs
+++ b/src/riscv/lib/src/state_backend/region.rs
@@ -143,6 +143,45 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


victor-dumitrescu (commented):
The changes to region.rs look good.

Since there's no mention otherwise, I assumed the scope of this PR would be the whole project? If that's the case, there are other places that could benefit from a trait implementation reordering, for example the proof deserialisation code.

PR #150, opened 2025-07-10
Add `Verifier` to backend_test! macro

felixp72:
Closes RV-699

# What

Add `Verify` backend to the `backend_test!` macro.

# Why

Better test coverage for the `Verifier` backend.

# How

* Implemented `ManagerAlloc` for `Verifier` and updated the `backend_test!` macro.
* Had to change the size of dynamic regions in 2 tests in order to pass the compile time check for number of bytes being a multiple of the leaf size of a merkle tree. This check is done for the dynamic regions of the `Verify` backend [here](https://github.com/tezos/riscv-pvm/blob/main/src/riscv/lib/src/state_backend/verify_backend.rs#L446).

# Manually Testing

```
make -C src/riscv all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/state_backend/verify_backend.rs
+++ b/src/riscv/lib/src/state_backend/verify_backend.rs
@@ -96,6 +96,39 @@ impl ManagerBase for Verifier {
     }
 }
 
+#[cfg(test)]
+mod test_helpers {
+    use crate::state_backend::ManagerAlloc;
+    use crate::state_backend::verify_backend::DynRegion;
+    use crate::state_backend::verify_backend::PageId;
+    use crate::state_backend::verify_backend::Region;
+    use crate::state_backend::verify_backend::Verifier;
+
+    impl<const LEAF_SIZE: usize, const LEN: usize> DynRegion<LEAF_SIZE, LEN> {
+        /// Construct a zero-initialized dynamic region.
+        fn zero_initialized() -> Self {
+            let nr_pages = LEN.div_ceil(LEAF_SIZE);
+
+            Self::from_pages((0..nr_pages).map(|page_id| {
+                let page_index = PageId::<LEAF_SIZE>::from_address(page_id * LEAF_SIZE);
+                (page_index, Box::new([0; LEAF_SIZE]))
+            }))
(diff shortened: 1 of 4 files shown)


vapourismo (changes requested):
Rebase please 🙏

src/riscv/lib/src/state_backend/verify_backend.rs
```diff
@@ -96,6 +96,51 @@ impl ManagerBase for Verifier {
...
+    use crate::state_backend::verify_backend::Verifier;
+
+    /// Trait for zero-initialized dynamic regions.
+    ///
+    /// Zero-initialising a dynamic region requires computing the  constant `NR_PAGES` which can not
+    /// be evaluated in a function. As a workaround we add it as an associated constant to this
+    /// trait.
+    trait ZeroInit<const LEAF_SIZE: usize, const LEN: usize> {
+        /// Number of pages in the dynamic region.
+        const NR_PAGES: usize = LEN.div_euclid(LEAF_SIZE);
```
vapourismo, line 114: You need to round up.

```suggestion
        const NR_PAGES: usize = LEN.div_ceil(LEAF_SIZE);
```

vapourismo: Also, where is this constant used through the trait?

vapourismo (changes requested):

src/riscv/lib/src/state_backend/verify_backend.rs
felixp72: It's used in the `zero_initialized` function. I can not use its expression inside the function body sadly, so I have to use an associated constant.

felixp72: I tried creating a `const` and that's why it wasn't working 🤦 . With just a `let` I can get rid of the trait. [diff](https://github.com/tezos/riscv-pvm/compare/5cbff6e72fde3382f86ff8290f9f7e5eaf37cb34..81849a835d4dd4987e573148bd9a6eed34d0d195)

PR #151, opened 2025-07-11
Track initialisation status of stack slots and values

vapourismo:
Closes RV-735

# What

This change tweaks the stack slot mechanism to track whether it has been initialised or not.

You can use `Slot::init` to initialise a slot proactively. Or, you can use `Slot::assume_init` to declare a slot as initialised after it has been written to by an external function. There is also `Pointer::assume_init` to obtain pointers to uninitialised slots and interpret them as initialised later - i.e. in case of branches where only one side may treat the pointee as initialised.

# Why

Tighter integration with typed IR values. Memory safety when calling Rust functions (via C ABI) that deal with references or pointers to un-/initialised memory.

I have prepared a follow-up PR that introduces some tweaks to the external function call framework, which harnesses this to enforce initialisation contracts across the external function barrier.

# Manually Testing

```
make -C src/riscv all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/jit/state_access/stack.rs
+++ b/src/riscv/lib/src/jit/state_access/stack.rs
@@ -38,6 +38,7 @@
 //! [`handle_exception`]: super::handle_exception
 
 use std::marker::PhantomData;
+use std::mem::MaybeUninit;
 use std::ptr::NonNull;
 
 use cranelift::codegen::ir::InstBuilder;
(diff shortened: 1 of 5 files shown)


src/riscv/lib/src/jit/state_access/stack.rs
```diff
@@ -156,7 +161,34 @@ impl<T: StackAddressable> Slot<T> {
...
+        Slot {
+            slot: self.slot,
+            ptr_type: self.ptr_type,
+            _pd: PhantomData,
+        }
+    }
+}
+
+impl<T: StackAddressable> Slot<T> {
+    /// Get a pointer to the memory of the slot.
```
NSant215, line 191: I seemingly can't comment on lines not in the diff - but line 196 ` unsafe { Value::<NonNull<T>>::from_raw(raw_value) }` - how do we ensure that whenever this is called, the value is `NonNull` - as it looks like we could call `new` to create it and then use `ptr` to get the pointer to a null value?

PR #154, opened 2025-07-14
Handle exceptions using external function framework

vapourismo:
Part of RV-738

# What

Introduces a framework that allows you to call external functions while ensuring that the typed IR values match.

## Duplication and Code Deletion

Some items introduced in this PR resemble the existing `abi` module. I chose to recreate the functionality in `typed` to avoid impacting the diffs with distracting refactors. In the last change related to RV-738, I am deleting the "old" way. [See tip of the branch](https://github.com/tezos/riscv-pvm/commits/ole/rv-738/).

# Why

There is no type safety when it comes to the values passed through the FFI barrier.

Some type mismatches may be easy to detect. However, some are more subtle and require a dedicated framework. Such as initialisation guarantee mismatches (e.g. `T` vs `MaybeUninit<T>`).

# Manually Testing

```
make -C src/riscv all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/jit/builder/ext_calls.rs
+++ b/src/riscv/lib/src/jit/builder/ext_calls.rs
@@ -1,0 +1,139 @@
+// SPDX-FileCopyrightText: 2025 TriliTech <contact@trili.tech>
+//
+// SPDX-License-Identifier: MIT
+
+//! Framework for calling external functions from Cranelift IR
+
+use cranelift::codegen::ir;
+use cranelift::prelude::FunctionBuilder;
+use cranelift::prelude::InstBuilder;
+use cranelift::prelude::Signature;
+use cranelift::prelude::isa::TargetFrontendConfig;
+
+use crate::jit::builder::typed::Type;
+use crate::jit::builder::typed::Typed;
+use crate::jit::builder::typed::Value;
+
+/// Trait for types that can be returned from external function calls
+pub trait ReturnTyped {
+    /// IR types of the function return values
+    const RETURN_TYPES: &[Type];
+
+    /// IR [`self::Value`] representation of the return values
+    type Value;
+
+    /// Construct the IR [`self::Value`] from the return values.
+    fn from_return_values(values: &[ir::Value]) -> Self::Value;
+}
+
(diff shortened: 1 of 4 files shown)


NSant215 (commented):
generally very keen for this change :D - the namings I am finding a bit confusing but also might just be getting used to them - it seems like you have introduced a non-cranelift 'IR' concept to use in-between cranelift and the interpreter that is the new `Type`? the structure makes sense I think but I'm wondering if there is a more descriptive name we could use for this ... intermediate IR? XD

vapourismo:
> generally very keen for this change :D - the namings I am finding a bit confusing but also might just be getting used to them - it seems like you have introduced a non-cranelift 'IR' concept to use in-between cranelift and the interpreter that is the new `Type`? the structure makes sense I think but I'm wondering if there is a more descriptive name we could use for this ... intermediate IR? XD

Maybe it is easier to think of those as abstractions to be used in place of their Cranelift variants.

NSant215 (approved):

src/riscv/lib/src/jit/builder/ext_calls.rs
```diff
@@ -0,0 +1,139 @@
...
+use cranelift::prelude::isa::TargetFrontendConfig;
+
+use crate::jit::builder::typed::Type;
+use crate::jit::builder::typed::Typed;
+use crate::jit::builder::typed::Value;
+
+/// Trait for types that can be returned from external function calls
+pub trait ReturnTyped {
+    /// IR types of the function return values
+    const RETURN_TYPES: &[Type];
```
victor-dumitrescu, line 20: I don't quite get how `RETURN_TYPES` works. In what cases would this be >1 ? What does it mean for a function to have multiple return types?

vapourismo: There are a few cases where you can have multiple return values. Ignoring that Cranelift itself allows you to return multiple values from a function, you can have multiple return values from external functions per the C ABI. Struct(ured) returns can be used to return up to two values, I think. Generally, Cranelift doesn't constrain the number of return values, though.

victor-dumitrescu (approved):

PR #156, opened 2025-07-14
Pass through the correct destination register for X32Mul/mulw

vapourismo:
Closes RV-741

# What

Changes the `union Register` into a struct. Additionally, fixes a bug in the `mulw` implementation that gets exposed from change to `Register`.

# Why

Some subtle bugs have arisen in the past related to the register being a union. The memory representation of the register enums is the same (`u8`).

A `NonZeroXRegister` can be used in place of an `XRegister`. In the case of `mulw`, the `Args::rd` field got constructed using an `XRegister`, but the instruction implementation expects a `NonZeroXRegister`.

This made it possible to call `run_x32_mul` with `rd = x0` despite it expecting type `NonZeroXRegister`.

This is super dangerous because `x0` is represented as `255` at runtime, and we use that value as an index into the X register values array. That means if you run the RISC-V instruction `mulw x0, _, _`, you will inevitably write to index 255 of an array with 31 items.  Even worse, there is no explicit runtime error for this, so we just write to random memory.

# How

This PR makes the instruction dispatch memory safe. You can still construct `Args` incorrectly, but at least that won't result in random writes to memory any more.

# Manually Testing

Check out the first commit and run the test suite.

```
cargo test --lib
```

This should fail on tests related to `mulw`. Once you check out the branch tip, all tests should pass again.

Alternatively, you can run tests without optimisations `RUSTFLAGS="-C opt-level=0" make test`.

# Regressions

The state shape has changed due to the `Register` union becoming a struct. This results in a different state hash as the serialisation output changes.

# Tasks for the Author

  - There are existing tests that fail after the migration to `Register` as a struct


src/riscv/lib/src/machine_state/instruction/constructors.rs
```diff
@@ -2712,6 +2712,17 @@ impl Instruction {
         }
     }
 
+    /// TODO
+    pub(crate) fn from_ic_mulw(args: &RTypeArgs) -> Instruction {
```
vapourismo, line 2716: @claude Write a better documentation comment for this function.

src/riscv/lib/src/machine_state/instruction.rs
vapourismo, line 1: @claude Remove the MIRI tests in this file as they are no longer needed. Also check if the `test-miri` Makefile target would still run any tests.

kurtisc (approved):

vapourismo:
It seems there is a divergence issue with this PR. I'll look into it.

vapourismo:
> lgtm - can we remove all tagged instruction stuff with this change? should be able to serialise/deserialise Instruction directly?

Yes! This was, in fact, necessary to avoid state divergences. The new top commit removes it.

src/riscv/lib/src/machine_state/instruction/shape.rs
```diff
@@ -0,0 +1,125 @@
...
+    XSrcFSrc,
+    // rd, rs1, rs2 => NZX
+    NZXSrcNZXDest,
+    // rd => NZX | rs1, rs2 => X
+    XSrcNZXDest,
+}
+
+/// This function maps each opcode to the corresponding ArgsShape that the opcode uses
+/// i.e. the types of registers used.
+/// This is used to determine how to serialise and deserialise the Args for each opcode.
```
NSant215, line 30: I don't think this function is used for serde any more? so we can remove that part of the docstring

src/riscv/lib/src/machine_state/instruction.rs
```diff
@@ -2736,4 +2742,102 @@ mod test {
             );
         }
     });
+
+    #[test]
+    fn test_miri_instruction_serde() {
```
NSant215, line 2747: are these tests still valuable? Just want to confirm as in my head they were only because of the union stuff

src/riscv/lib/src/machine_state/instruction/shape.rs
NSant215: in fact it only seems used for debug - wondering if we could make this cleaner / or just include all reg types in the debug statement - this might affect some tests though so we could also push this to later

vapourismo: Ah yes! I think using conventional `Debug` would help here. I removed the Shape-related stuff and derived the `impl Debug` instead [here](https://github.com/tezos/riscv-pvm/compare/1c4f907286b4871c5b5af8d32b0e2accf27105b9..2758286accba59c38a2265011de8efcf0d86410a).

src/riscv/lib/src/machine_state/instruction.rs
vapourismo: Seems like they can be [removed](https://github.com/tezos/riscv-pvm/pull/156/commits/ea04f202754e5cb0ba410c7da980cce34270faa7).

NSant215 (approved):

src/riscv/lib/src/machine_state/instruction/constructors.rs
```diff
@@ -25,7 +25,7 @@ use crate::parser::instruction::UJTypeArgs;
 use crate::parser::split_x0;
 
 impl Instruction {
-    /// Create a new [`Instruction`] with the appropriate [`super::ArgsShape`] for the [`OpCode::Add`].
+    /// Create a new [`Instruction`] for the [`OpCode::Add`].
```
victor-dumitrescu, line 28:
```suggestion
    /// Create a new [`Instruction`] for [`OpCode::Add`].
```

src/riscv/lib/src/machine_state/instruction/constructors.rs
```diff
@@ -82,7 +82,7 @@ impl Instruction {
         }
     }
 
-    /// Create a new [`Instruction`] with the appropriate [`super::ArgsShape`] for the [`OpCode::Sub`].
+    /// Create a new [`Instruction`] for the [`OpCode::Sub`].
```
victor-dumitrescu, line 85:
```suggestion
    /// Create a new [`Instruction`] for [`OpCode::Sub`].
```

src/riscv/lib/src/machine_state/instruction/constructors.rs
```diff
@@ -120,7 +120,7 @@ impl Instruction {
         }
     }
 
-    /// Create a new [`Instruction`] with the appropriate [`super::ArgsShape`] for the [`OpCode::Neg`].
+    /// Create a new [`Instruction`] for the [`OpCode::Neg`].
```
victor-dumitrescu, line 123:
```suggestion
    /// Create a new [`Instruction`] for [`OpCode::Neg`].
```

src/riscv/lib/src/machine_state/instruction/constructors.rs
```diff
@@ -159,7 +159,7 @@ impl Instruction {
         }
     }
 
-    /// Create a new [`Instruction`] with the appropriate [`super::ArgsShape`] for  [`OpCode::Nop`].
+    /// Create a new [`Instruction`] for  [`OpCode::Nop`].
```
victor-dumitrescu, line 162:
```suggestion
    /// Create a new [`Instruction`] for [`OpCode::Nop`].
```

victor-dumitrescu (approved):

src/riscv/lib/src/machine_state/instruction/constructors.rs
vapourismo: Fixed [here](https://github.com/tezos/riscv-pvm/compare/7761e794af17d2e6cb7eb977a39caa04ae1d67a6..6b1bffa5a9f2dd78219fcf83ea0a7a6715051d44).

src/riscv/lib/src/machine_state/instruction/constructors.rs
vapourismo: Remove the word in [this change](https://github.com/tezos/riscv-pvm/compare/7761e794af17d2e6cb7eb977a39caa04ae1d67a6..6b1bffa5a9f2dd78219fcf83ea0a7a6715051d44).

src/riscv/lib/src/machine_state/instruction/constructors.rs
vapourismo: Fixed [here](https://github.com/tezos/riscv-pvm/compare/7761e794af17d2e6cb7eb977a39caa04ae1d67a6..6b1bffa5a9f2dd78219fcf83ea0a7a6715051d44).

vapourismo:
@claude, find typos and fix them.

PR #167, opened 2025-07-16
Raise Store AMO access fault exception using external function framework

vapourismo:
Part of RV-738

# What

# Why

# How

# Manually Testing

```
make -C src/riscv all
```

# Regressions

# Tasks for the Author


NSant215 (commented):
One thing I failed to register at first with this new framework - we no longer store the reference to the extern 'C' function in JsaCalls like we did before - I don't quite follow the affect of this. Does it mean that for situations where we use a particular external function a lot, we would be spending more time fetching the function reference?

NSant215 (approved):

PR #178, opened 2025-07-17
Update CI benchmark to run 20 transactions and 20 iterations

vapourismo:
Update CI benchmark to run 20 transactions and 20 iterations. Hopefully we can reign in the increased flakiness of the benchmark.

### DIFF
--- a/scripts/ci-bench.sh
+++ b/scripts/ci-bench.sh
@@ -2,6 +2,9 @@
 
 set -e
 
+NUM_TXS=20
+NUM_ITERS=20
+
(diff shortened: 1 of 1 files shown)


vapourismo:
> Somewhat unrelated but more relevant with the increased iterations, but perhaps only posting the stats table to the PR instead of the results of every iteration would be enough? They'd still be available in the job log but it would declutter the PR a bit.

I like to look at the results from time to time. Perhaps it can be made into a collapsible section.

victor-dumitrescu (approved):

PR #180, opened 2025-07-17
RISC-V: Support doc comments for `struct_layout`

kurtisc:
Closes RV-743

# What
Documentation comments are syntactic sugar for a `#[doc]` attribute. By passing on any attributes in the `struct_layout` macro, we can pass on documentation comments - and potentially any other attribute.

I don't think it's necessary to update the example.

See also:

https://rust-lang.github.io/api-guidelines/macros.html#item-macros-compose-well-with-attributes-c-macro-attr

# Why
I'm trying to write documentation for something using `struct_layout` and don't want to use `//`

# How

Reproducing any attributes inside the macro

# Manually Testing

```
make -C src/riscv all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/state_backend/layout.rs
+++ b/src/riscv/lib/src/state_backend/layout.rs
@@ -67,6 +67,9 @@ impl<const LEN: usize> Layout for DynArray<LEN> {
 #[macro_export]
 macro_rules! struct_layout {
     (
+        $(
+            #[$attributes:meta]
+        )*
         $vis:vis struct $layout_t:ident $(< $($param:ident),+ >)? {
             $($field_vis:vis $field_name:ident: $cell_repr:ty),+
             $(,)?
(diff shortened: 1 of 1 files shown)


src/riscv/lib/src/state_backend/layout.rs
```diff
@@ -67,13 +67,19 @@ impl<const LEN: usize> Layout for DynArray<LEN> {
...
+        $(
+            #[$attributes:meta]
+        )*
         $vis:vis struct $layout_t:ident $(< $($param:ident),+ >)? {
             $($field_vis:vis $field_name:ident: $cell_repr:ty),+
             $(,)?
         }
     ) => {
         paste::paste! {
             #[derive(serde::Deserialize, serde::Serialize, Debug, Clone, PartialEq, Eq)]
```
felixp72, line 79: Nit: Should these be taken out and be added at use-site now that we support attributes?

felixp72 (approved):

src/riscv/lib/src/state_backend/layout.rs
kurtisc: I don't think so, any struct using this macro will need these derivations.

vapourismo (approved):

PR #186, opened 2025-07-22
Split `FromProofError` type

felixp72:
Closes RV-629

# What

# Why

# How

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author


src/riscv/lib/src/state_backend/proof_backend/proof/deserialise_owned.rs
```diff
@@ -103,12 +109,12 @@ impl ProofTreeDeserialiser<'_> {
 
 /// Suspended computation combinator for [`ProofTreeDeserialiser`] deserialiser.
 pub struct OwnedParserComb<'t, R> {
-    result: R,
+    result: ParseProofResult<R>,
```
vapourismo, line 112: What's the motivation behind moving the fallibility into the struct?

felixp72: Otherwise, the separation between error types wouldn't be a separation, but rather a nested doll type of approach. Now, `DeserResult` & `ProofParseResult` have distinct error types with distinct variants (whereas otherwise, `DeserResult` would have a variant for `ProofParseResult`)

 I would say this is an improvement overall - having the minimal error type needed in each place & making both implementations align better with eachother.

vapourismo: Do you think this may cause problems with suspended computations that can't fail? Do those even exist?

vapourismo:
I've had multiple attempts at reviewing this PR. However, every time I abandon it because I am confused as to what is going on and why. I don't see the simplification for the purpose of readability and avoidance of confusion. I struggle to connect the dots for most changes as to why they're needed and how they make it more readable or simpler. Do you think it is possible to break this change down into smaller digestible bits? It's not that the PR is huge, but it seems complex.

src/riscv/lib/src/state_backend/proof_backend/proof.rs
```diff
@@ -288,6 +289,7 @@ pub fn serialise_merkle_tree(tree: &MerkleProof) -> impl Iterator<Item = u8> + '
     tags_encoding.chain(nodes_encoding)
 }
 
+/// The raw byte representation of a tag is invalid.
```
NSant215: not a big deal but you get rid of this error type later anyway so not sure if adding the comment has value XD

src/riscv/lib/src/state_backend/proof_backend/proof/deserialise_owned.rs
```diff
@@ -44,13 +44,13 @@ impl<'t> Deserialiser for ProofTreeDeserialiser<'t> {
                         })?;
                 Ok(bytes)
             })
-            .map(OwnedParserComb::new)
+            .map(|data| OwnedParserComb::new(Ok(data)))
```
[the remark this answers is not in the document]
felixp72: The result type on `OwnedParserComb` will end up actually be `ProofParseResult<R>` - you will see this is actually used [later](https://github.com/tezos/riscv-pvm/pull/186/commits/9902bf2088980a2b4cd3b42dbca17ae916289f80#diff-5fd08e43fe0a7d5c20bcbe888108fee830351bdb3b3d3dce9706c1ada6be99e7R58) when parsing the leaf - you can have an error there, and it is a parsing error. TLDR I wanted to separate errors from parsing the data and errors from parsing the layout. Relevant thread [here](https://github.com/tezos/riscv-pvm/pull/186#discussion_r2231589456)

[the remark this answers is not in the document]
felixp72: typo fixed btw, thanks

felixp72:
I made names for result/error types more consistent [here](https://github.com/tezos/riscv-pvm/compare/844c4a1de335ea3b59693d7c3426ab15576da07b..5c29b402cb081515524b4b5e878991792ff7d093)

NSant215 (approved):

vapourismo (approved):

PR #200, opened 2025-07-28
Use Blake3 instead of Blake2b when hashing

vapourismo:
Closes RV-250

# What

Use the Blake3 hash function instead of Blake2b through Tezos crypto.

# Why

image

It is faster. Also, the `blake3` crate comes with a nice API. It doesn't introduce redundant error cases (errors, not exceptions).

# Manually Testing

```
make all
```

# Regressions

Changing the hash function results in different state hashes being produced. That results in regression tests captures changing.

# Tasks for the Author


kurtisc (approved):

HantangSun:
There seems to be no meaningful performance impact when I run the benchmarks locally. 

Also, would non-cryptographic hash functions(e.g [xxhash](https://xxhash.com/)) be more suitable for the use case here? Since we are not making use of the security property of the hash function.

vapourismo:
> There seems to be no meaningful performance impact when I run the benchmarks locally.

Which benchmark are you running that exercises the hash function from this PR?

> Also, would non-cryptographic hash functions(e.g [xxhash](https://xxhash.com/)) be more suitable for the use case here? Since we are not making use of the security property of the hash function.

We're relying on the improbable collision properties for arbitrary input. So we need a cryptographic hash function.

HantangSun:
> > There seems to be no meaningful performance impact when I run the benchmarks locally.
> 
> Which benchmark are you running that exercises the hash function from this PR?
> 
> > Also, would non-cryptographic hash functions(e.g [xxhash](https://xxhash.com/)) be more suitable for the use case here? Since we are not making use of the security property of the hash function.
> 
> We're relying on the improbable collision properties for arbitrary input. So we need a cryptographic hash function.

Ok, the improbable collision property makes sense. 

For the benchmark, I am running the jstz with outline jit on macos and performance change is quite tiny.

PR #220, opened 2025-08-01
Remove page constraint for block

HantangSun:
Closes RV-749

# What

Remove page constraint for block

# Why

Previously blocks cannot span across 2 memory pages, so at page boundaries we need to stop the current block and then create a new block. This is no longer needed.

# How

Removes page boundary checks from block logic

# Regressions

Regression caused because block can now continue when reaching page boundary

# Tasks for the Author


src/riscv/lib/src/machine_state/block_cache/state.rs
```diff
@@ -17,7 +17,6 @@ use crate::machine_state::instruction::Instruction;
 use crate::machine_state::memory::Address;
 use crate::machine_state::memory::MemoryConfig;
 use crate::machine_state::memory::OFFSET_MASK;
```
vapourismo, line 19: If this import still exists, I think there might still be a page boundary check.

PR #222, opened 2025-08-04
projection: eliminate potential negative offsets

emturner:
# What

Eliminate _potential_ source of undefined/incorrect behaviour w.r.t. offsets:
- offsets incorrectly use `wrapping` arithmetic, rather than saturating/checked

# Why

Offset usage in the JIT must be correct and point to the exact place in memory. Since all offsets are part of _projections_, they can only ever be positive (ie looking _inside_ a larger structure) and we should ensure this is the case.

While this currently is not encountered by existing code, it would be a potential 'gotcha' when dealing with larger structures in memory (e.g. indexing into main memory, which can be extremely large depending on the configuration)

# How

Use `usize` for calculating all offsets, with checked arithmetic to avoid overflow. When using an offset, care can be taken at the _usage_ site to ensure the offset is indeed valid.

While not strictly necessary for this PR, switching to `usize` emphasises the non-negative requirement, and will be needed when indexing into potentially large arrays in future (particularly main memory)

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/state_context/projection.rs
+++ b/src/riscv/lib/src/state_context/projection.rs
@@ -68,6 +68,64 @@ impl TypeCons for MachineCoreCons {
     type Applied<MC: MemoryConfig, M: ManagerBase> = MachineCoreState<MC, M>;
 }
 
+/// Offset from a base pointer to a [projection's] subject, within the owned backend.
+///
+/// Additional offsets may be added to an existing one, to build up offsets within
+/// layered projections. All such additions will panic on overflowing the `i32` range.
+///
+/// [projection's]: Projection
+#[derive(Debug, Clone, Copy, Default)]
+pub struct ProjectionOffset {
+    offset: i32,
+}
+
+impl ProjectionOffset {
+    /// Create a new projection offset from the given offset.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the offset overflows the `i32` range.
+    pub fn new(offset: usize) -> Self {
+        Self {
+            offset: offset.try_into().expect("offset overflows i32 range"),
+        }
+    }
+}
+
+impl std::ops::Add<usize> for ProjectionOffset {
+    type Output = Self;
+
+    fn add(self, other: usize) -> Self {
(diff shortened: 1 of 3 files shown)


src/riscv/lib/src/state_context/projection.rs
```diff
@@ -108,7 +108,11 @@ pub trait Projection {
     /// offset to an address of the subject value that would give you the address of the target
     /// value. This is exclusive to the [`crate::state_backend::owned_backend::Owned`] state
     /// backend.
-    fn owned_pointer_offset<MC: MemoryConfig>(param: Self::Parameter) -> i32;
+    ///
+    /// When using this value, care should be taken when converting to signed offsets.
+    /// In particular, the offset represents a _positive_ value and this must be upheld
+    /// when converting to signed values.
+    fn owned_pointer_offset<MC: MemoryConfig>(param: Self::Parameter) -> usize;
```
vapourismo, line 115: Could you introduce a dedicated type to encode the right semantics? 
Or perhaps you could re-use [`Offset32`](https://docs.rs/cranelift-codegen/latest/cranelift_codegen/ir/immediates/struct.Offset32.html).

vapourismo (changes requested):

src/riscv/lib/src/state_context/projection.rs
emturner: absolutely, wasn't sure if it would be overly complex or not. But makes sense to indeed

vapourismo:
FYI, I'll be extending projection offset capabilities at some point. [This](https://github.com/tezos/riscv-pvm/commit/6df8e51b1e9bff038fc8ad782e5126618c55440d#diff-aa531cabdbc0311e9619a67ef333f10ca3f2a92f36a991c25bbe947032010299R80-R91) is what I ended up with last week.

src/riscv/lib/src/state_context/projection.rs
emturner: have introduced `ProjectionOffset` now, which can add itself/usize.

vapourismo (approved):

PR #228, opened 2025-08-05
Move exception handling code out of JIT function

vapourismo:
Closes RV-750

# What

This PR moves exception handling out of JIT-produced functions. The generated code (both IR and final assembly) for JIT-compiled sequences of instructions becomes simpler that way.

I had to change a few JIT scenario tests which now perform one fewer step, as the exception handling, which counted as the completion of a step in those cases, is no longer done. This has been split out into a separate commit for your reviewing convenience.

# Why

The changes result in a slight reduction in complexity of the JIT builder. It also results in a ~2.12% performance boost.

# How

Since exceptions are no longer handled, there is no mechanism to update the `Result<(), EnvironmentException>` parameter via its mutable reference.

Strictly speaking, the exception kind is no longer an environment exception, but a regular exception. Changing the out parameter type to `Result<(), Exception>` would require an external function call to convert the `ExceptionCode` as it is raised into an `Exception` and then update the result parameter. To avoid this additional external function call, we change the result parameter to `ExceptionCode`. This means we can directly update the mutable reference as an exception is raised; no conversion needed.

Exception handling is now done in a single place: `Pvm::step_max_handle`.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/jit/state_access.rs
+++ b/src/riscv/lib/src/jit/state_access.rs
@@ -42,7 +42,6 @@ use crate::machine_state::registers::FValue;
 use crate::machine_state::registers::XValue;
 use crate::state_backend::Elem;
 use crate::state_backend::owned_backend::Owned;
-use crate::traps::EnvironException;
 use crate::traps::Exception;
 
 /// Exception codes used for efficient exception handling in JIT-compiled code
@@ -50,7 +49,7 @@ use crate::traps::Exception;
 /// This enum represents different types of exceptions that can occur during
 /// instruction execution, encoded as i64 values for easy transmission between
 /// JIT-compiled code and runtime handlers.
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(i64)]
 pub enum ExceptionCode {
     NoException = 0,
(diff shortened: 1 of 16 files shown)


src/riscv/lib/src/jit/state_access.rs
```diff
@@ -82,22 +81,18 @@ impl ExceptionCode {
     }
 
     /// Convert the exception code back to an [`Exception`].
```
emturner, line 83: `... back to the corresponding [`Exception`], if applicable.`?

src/riscv/lib/src/machine_state.rs
```diff
@@ -543,16 +504,25 @@ impl<MC: memory::MemoryConfig, BCC: BlockCacheConfig, B: Block<MC, M>, M: backen
                     steps = steps.saturating_add(1);
                     step_bounds = bound_saturating_sub(step_bounds, 1);
 
-                    match handle(self, cause) {
-                        Ok(may_continue) => {
-                            if !may_continue {
-                                break None;
+                    match cause {
```
emturner, line 507: side note - this MR change does indeed make the Fence.I update much simpler 🙏🏻

emturner (approved):

src/riscv/lib/src/jit/state_access.rs
emturner: (resolve at will)

NSant215 (approved):

PR #248, opened 2025-08-12
RISC-V: Supervisor synchronous signal support

kurtisc:
Closes RV-726
Closes RV-653

# What
Adds support for synchronous signal handlers for exceptions

# Why

This is the minimum support needed for the first milestone of signal support.

See [signal(7)](https://man7.org/linux/man-pages/man7/signal.7.html) (Execution of signal handlers) and [rt_sigreturn(2)](https://man7.org/linux/man-pages/man2/sigreturn.2.html)

The control flow goes like this:

Setting a signal handler:
 - A signal handler function pointer is passed using `sigaction(2)` (`signal(2)` is currently unsupported)
 - The `sa_restorer` is saved into memory
   - This is a function pointer to a function provided by the libc. It just [calls `rt_sigreturn`](https://github.com/kraj/musl/blob/1b06420abdf46f7d06ab4067e7c51b8b63731852/src/signal/riscv64/restore.s#L9)

Raising a signal:
 - In places where we previously wrote 0 to the program counter, we now call a signal handler
   - In practice, all of these have a TERM or CORE disposition, so go to 0 afterwards anyway
 - The program counter is set to the signal handler
 - The return address is set to the old program counter
 - Other context is saved to the main memory stack

Returning from a signal
 - The context is restored
 - The old program counter (in practice, currently always 0) is restored from the return address
 
 Consider this an implementation of existing POSIX/Linux logic for the supervising PVM, rather than a bespoke system.

# How
 - Implements a `sigreturn` handler
 - Updates `step_max_handle` to optionally set the program counter to a signal handler
 - Adds the logic for returning from a signal handler (tested but not used)
 
 Next steps:
 - Signals that don't TERM will need to continue execution, needing the data in the registers to be restored.
 - The data stored on the stack should be changed so that signal handlers can read it using the `siginfo_t` struct

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/pvm/linux.rs
+++ b/src/riscv/lib/src/pvm/linux.rs
@@ -35,6 +35,7 @@ use crate::machine_state::memory::Permissions;
 use crate::machine_state::registers;
 use crate::program::Program;
 use crate::pvm::hooks::PvmHooks;
+use crate::pvm::linux::signals::Signal;
 use crate::state::NewState;
 use crate::state_backend::AllocatedOf;
 use crate::state_backend::Atom;
@@ -785,7 +786,13 @@ impl<M: ManagerBase> SupervisorState<M> {
                     .hart
                     .xregisters
                     .write_system_call_error(Error::NoSystemCall);
-                false
+
+                if machine.core.dispatch_signal(Signal::Sigsys).is_err() {
+                    // If handling the signal fails, set the program counter to zero as if there is
+                    // no handler.
+                    machine.core.hart.pc.write(0);
+                }
+                true
             }
 
             Err(error) => {


src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -215,12 +229,63 @@ struct_layout! {
...
+    /// Get the program counter for handling a given signal, if a handler exists, and prepare to
+    /// jump to it.
+    pub fn handle_signal(&mut self, signal: Signal) -> Option<Address> {
+        // We only currently support the `action` field. Supporting the `handler` field requires
+        // parsing `flags`.
+        let handler = self.signal_actions.read_action(signal);
+
+        if handler == NO_HANDLER {
+            None
+        } else {
```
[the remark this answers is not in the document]
kurtisc: I'm still in my Rust expressions honeymoon phase :)

[Less indentation](https://github.com/tezos/riscv-pvm/compare/8d45b8066cc65fd6ed23391d4d714efc0fe03353..5385940bce1629fac4bba84bd77c44024bd081c9), same link as above

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -599,4 +599,22 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+    ///
+    /// See: <https://www.man7.org/linux/man-pages/man2/rt_sigreturn.2.html>
+    pub(super) fn handle_rt_sigreturn(
+        &self,
+        core: &mut MachineCoreState<impl MemoryConfig, M>,
+    ) -> Result<u64, Error>
+    where
+        M: ManagerReadWrite,
+    {
+        let pc = core.pop_signal_context().map_err(|_| Error::Fault)?;
```
victor-dumitrescu, line 673: According to the man page, I think this should be `ENOSYS`.

src/riscv/lib/src/machine_state.rs
```diff
@@ -492,9 +494,37 @@ impl<MC: memory::MemoryConfig, BCC: BlockCacheConfig, B: Block<MC, M>, M: backen
...
+                        Exception::IllegalInstruction => {
+                            // We only currently support the `action` field.
+                            // TODO RV-732 Parse `sa_flags` subset.
+                            let handler = self.core.signal_actions.read_action(Signal::Sigill);
+                            if handler != NO_HANDLER {
+                                self.core.dispatch_signal(Signal::Sigill, handler);
+                                break None;
+                            } else {
+                                self.core.hart.pc.write(0);
+                            }
```
victor-dumitrescu, line 506: Since this is identical to the code below except for the argument to `read_action`, perhaps it would be worth pulling it out into a function, unless you know that these blocks will diverge when implementing the missing functionality.

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -215,12 +229,54 @@ struct_layout! {
 }
 
 impl<MC: MemoryConfig, M: ManagerReadWrite> MachineCoreState<MC, M> {
+    /// Set the hart state to a signal handler
+    pub fn dispatch_signal(&mut self, signal: Signal, handler: VirtAddr) {
+        let restorer = self
+            .push_signal_context(signal)
+            .expect("Signal handling should succeed");
```
vapourismo, line 236: I'm not sure if you have any control over `push_signal_context` failing or not at this point. Probably best to forward the error.

src/riscv/lib/src/machine_state.rs
kurtisc: [Updated](https://github.com/tezos/riscv-pvm/compare/29fe17914ec413857d4d7702c9d15bffa1e89b97..4d6d3e6f2f873d30b4831c7b320154b85380156e)

src/riscv/lib/src/pvm/linux/signals.rs
victor-dumitrescu: Oh, I misread this as the error being returned by calling `sigreturn` but raising it would mean the signal context was not correctly set up, right?

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -215,12 +229,57 @@ struct_layout! {
...
+
+        self.hart
+            .xregisters
+            .write_nz(nz::ra, restorer.to_machine_address());
+
+        // Write handler arguments
+        // Signal number
+        self.hart.xregisters.write_nz(nz::a0, signal as u64);
+
+        // TODO RV-754
```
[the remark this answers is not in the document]
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/4d6d3e6f2f873d30b4831c7b320154b85380156e..938fa7d8d628d1a958df0786f566060630958eba#diff-4293ba17c6d4b17be9dd006e0ca50f51760cc8ee0bc78d11aaaaef4f81bf74baR250)

victor-dumitrescu (approved):

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -781,6 +790,9 @@ impl<M: ManagerBase> SupervisorState<M> {
                     .hart
                     .xregisters
                     .write_system_call_error(Error::NoSystemCall);
```
vapourismo, line 792: Should we be writing the return registers when the signal gets dispatched?

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -781,6 +790,9 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+
+                // Execution will end regardless
+                let _ = machine.core.dispatch_signal(Signal::Sigsys);
                 false
```
vapourismo, line 796: If a signal handler is trigger we should probably continue.

src/riscv/lib/src/machine_state.rs
```diff
@@ -534,6 +548,19 @@ impl<MC: memory::MemoryConfig, BCC: BlockCacheConfig, B: Block<MC, M>, M: backen
...
+
+    fn dispatch_signal_or_trap(&mut self, signal: Signal)
+    where
+        M: ManagerReadWrite,
+    {
+        // We only currently support the `action` field.
+        // TODO RV-732 Parse `sa_flags` subset.
+        let handler = self.core.signal_actions.read_action(signal);
+
+        if handler == NO_HANDLER || self.core.dispatch_signal(signal).is_err() {
```
vapourismo, line 560: Nit: Could `dispatch_signal` return `Err` when no handler has been configured? That would simplify the logic a little.

Resolve at will.

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -599,4 +660,21 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+    pub(super) fn handle_rt_sigreturn(
+        &self,
+        core: &mut MachineCoreState<impl MemoryConfig, M>,
+    ) -> Result<(), Error>
+    where
+        M: ManagerReadWrite,
+    {
+        let pc = core.pop_signal_context().map_err(|_| Error::Fault)?;
+        core.hart.pc.write(pc);
+        Ok(())
```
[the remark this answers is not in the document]
vapourismo: What if the user kernel purposefully does not follow the same signal model? The lack of isolation really worries me here. I am wondering if you have thought through the attack vector on this.

Contracts in Jstz and Etherlink run at the same privilege level as Jstz and Etherlink itself. This makes escaping each runtime's sandbox a little easier. The question is, if they manage to exploit the signal handling mechanism, would have have universal access already anyway, or would the exploit enable it.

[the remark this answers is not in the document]
kurtisc: I'm struggling to come up with an answer for what feel like quite open-ended concerns. Do you have any specific examples?

I'd compare it to how a user kernel could similarly run an intentionally malformed function that breaks the assumption that a `jalr x0 0 x1` happens with a well-formed stack frame.

Maybe a malicious user could directly issue `rt_sigreturn` and pop a malformed stack, but the outcomes I see for that are:
 - bad values are written to things users can already write bad values to - `pc`, `sp`, signal masks, signal numbers
 - things operate on unexpected memory - but we already have memory protections, and users can try to write anything to anywhere already
 
So I'm not aware of any additional access that would be provided to a malicious user that can already write to registers and execute `ecall`

Maybe there is some value in sanity checking the values that come off the stack?

src/riscv/lib/src/pvm/linux.rs
emturner: at the least should there then be a `TODO` here?

src/riscv/lib/src/pvm/linux/signals.rs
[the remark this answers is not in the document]
kurtisc: @felixp72 the libc wrapper isn't a security feature, userspace code can still access the stack writing functions by using `rt_sigreturn(2)` directly:
```
#[no_mangle]
pub extern "C" fn call_rt_sigreturn() -> ! {
    unsafe {
        asm!(
            "li a7, 139",
            "ecall",
            options(noreturn)
        );
    }
}
```

but an adversary has no reason to, because userspace code can also just do this:

```
#[no_mangle]
pub extern "C" fn mangle_pc_and_sp() -> ! {
    unsafe {
        asm!(
            "li sp, 0",
            "li t0, 0",
            "jr t0",
            options(noreturn)
        );
    }
}
```

signal masks and signums likewise can be manipulated by userspace with `rt_sigaction(2)`, `signal(2)`, `kill(2)`. The premise is that what is being saved is userspace context.

A threat model that depends on these values being left alone is not safe.

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -522,6 +527,25 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+                    stringify!($system_call),
+                    ($(&$arg),*)
+                }
+            };
+        }
+
+        if system_call_no == RT_SIGRETURN {
+            trace_call!(handle_rt_sigreturn);
+            return self.handle_rt_sigreturn(&mut machine.core).is_ok();
+        }
```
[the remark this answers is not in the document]
kurtisc: Rolled back [link](https://github.com/tezos/riscv-pvm/compare/0627e254fb313f6e187ad3378fdd2cf0d6fe4c69..b6ba63dcd6e0aa0622bf7e9620bc8f75853e428f)

src/riscv/lib/src/pvm/linux.rs
kurtisc: I've updated this code a bit [here](https://github.com/tezos/riscv-pvm/compare/f7f531c40f72febfae326bd2568703d3508444ba..2f4dd7ba5bb3273199be40fc5c9bf675e95ca752), but this was already `false` before. It has been forever [1](https://gitlab.com/tezos/tezos/-/commit/be92521c8f4939e1c673990ced10a0cd6f33ca87#c5ebfd36824a317f7f3eba588591c1ee03cc8a33_457_451) [2](https://gitlab.com/tezos/tezos/-/commit/a55c36b3e4d00335462098b59e89c37e853b72cf#c5ebfd36824a317f7f3eba588591c1ee03cc8a33_126_197). <- Gitlab from 2024!

I suppose we don't *actually* want to terminate ever (hence the desires for being able to handle signals) but writing 0 to the program counter is how that is currently emulated. For that reason I've done that here for when there is no signal or the handling fails.

So:
>returning false means "check the PVM status"
is this de facto, or what we actually want? The code/comments/commit messages I've seen while investigating this imply that `false` is supposed to stop execution, whether that's correct or not.

The `on_tezos` argument is also confusing because it's one thing in one place for runtime and a nominally 'default' handler for all the tests.

I'm happy to improve the data type but don't know what the goal of it is. In fact, should we return *anything*? What do we get from breaking the loop?

src/riscv/lib/src/machine_state.rs
kurtisc: Never mind, that doesn't work out. I need having no handler not to be considered an error for the sigsys path. Reverted & a comment added

vapourismo (approved):

PR #259, opened 2025-08-19
RISC-V: Refactor `jit.rs` tests to use `step_max_inner`

NSant215:
Relates to RV-703.

# What

This MR changes the level at which a sequence is called from. Rather than calling  `run_block` on a specific sequence, we run `step_max_inner` over a JIT machine state and an Interpreted machine state, with the sequence inserted into the block-cache.

# Why

To retain the current JIT tests when moving to intra-sequence jumps. Once changed to intra-sequence jumps, running a sequence as interpreted vs JIT-compiled will not exit at the same point unless the number of steps to complete is passed in. 

# How

Raise the call level to `step_max_inner` rather than directly running a block.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/jit.rs
+++ b/src/riscv/lib/src/jit.rs
@@ -300,10 +300,13 @@ mod tests {
     use crate::instruction_context::LoadStoreWidth;
     use crate::interpreter::float::RoundingMode;
     use crate::machine_state::MachineCoreState;
-    use crate::machine_state::StepManyResult;
+    use crate::machine_state::MachineState;
+    use crate::machine_state::block_cache::TestCacheConfig;
     use crate::machine_state::block_cache::block::Block;
+    use crate::machine_state::block_cache::block::InlineCompiler;
     use crate::machine_state::block_cache::block::Interpreted;
     use crate::machine_state::block_cache::block::InterpretedBlockBuilder;
+    use crate::machine_state::block_cache::block::Jitted;
     use crate::machine_state::memory::M4K;
     use crate::machine_state::memory::Memory;
     use crate::machine_state::memory::MemoryConfig;
@@ -329,6 +332,9 @@ mod tests {
     type SetupHook = dyn Fn(&mut MachineCoreState<M4K, Owned>);
     type AssertHook = dyn Fn(&MachineCoreState<M4K, Owned>);
 
+    /// Machine state for test scenarios with a configurable [`Block`] type.
+    type TestMachineState<B> = MachineState<M4K, TestCacheConfig, B, Owned>;
+
     struct Scenario {
         initial_pc: Option<u64>,
         expected_steps: Option<usize>,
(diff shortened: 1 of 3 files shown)


src/riscv/lib/src/machine_state/block_cache/state.rs
```diff
@@ -397,6 +397,17 @@ impl<const SIZE: usize, B: Block<MC, M>, MC: MemoryConfig, M: ManagerBase>
         let instr = entry.block.instr();
         instr.iter().map(|cell| cell.read_stored()).collect()
     }
+
+    /// *TEST ONLY* - insert a block into an index of the cache.
```
felixp72, line 401: There is a `tests` module in this file. You can either
- move this function there and avoid the cfg and special comment
- this function is currently only used in `jit.rs` - why not move it in the `tests` module there?

at the same time, can you change the comment to saying _why_ you need this function rather/as well what it does?

felixp72 (changes requested):

src/riscv/lib/src/jit.rs
```diff
@@ -348,75 +351,82 @@ mod tests {
...
+                .main_memory
+                .set_all_readable_writeable();
+
+            let mut jitted_state: MachineState<
+                M4K,
+                TestCacheConfig,
+                Jitted<InlineCompiler<M4K>, M4K>,
+                Owned,
+            > = MachineState::new(InlineCompiler::default());
+            jitted_state.core.main_memory.set_all_readable_writeable();
```
felixp72, line 386: Bit of a nit, to clear boilerplate. Works together with adding this type near the other type defs in the module
```
    /// Machine state for test scenarios, can still configure the [`Block`] type.
    type TestMachineState<B> = MachineState<M4K, TestCacheConfig, B, Owned>;
```


```suggestion
            // Create the states for the interpreted and jitted runs.
            let mut interpreted_state: TestMachineState<Interpreted<_, _>> =
                MachineState::new(InterpretedBlockBuilder);
            interpreted_state
                .core
                .main_memory
                .set_all_readable_writeable();

            let mut jitted_state: TestMachineState<Jitted<InlineCompiler<_>, _>> =
                MachineState::new(InlineCompiler::default());
            jitted_state.core.main_memory.set_all_readable_writeable();
```

src/riscv/lib/src/jit.rs
```diff
@@ -348,75 +351,82 @@ mod tests {
...
+
+            let mut interpreted_block = Interpreted::<M4K, _>::new();
```
felixp72, line 388:
```suggestion
            let mut interpreted_block = Interpreted::<_, _>::new();
```

src/riscv/lib/src/jit.rs
```diff
@@ -348,75 +351,82 @@ mod tests {
...
+            interpreted_block.start_block();
+            for instr in self.instructions.iter() {
+                interpreted_block.push_instr(*instr);
+            }
 
-            // Create the block of instructions.
-            let mut block = Interpreted::<M4K, _>::new();
-            block.start_block();
+            let mut jitted_block = Jitted::<InlineCompiler<M4K>, M4K>::new();
```
felixp72, line 394:
```suggestion
            let mut jitted_block = Jitted::<InlineCompiler<_>, _>::new();
```

felixp72 (commented):
nits for getting rid of some types boilerplate

src/riscv/lib/src/machine_state/block_cache/state.rs
NSant215: thanks done [here](https://github.com/tezos/riscv-pvm/compare/5814d30ff5384460ee25d6c4a74180ffe9be0eeb..26d7725d8a14caa5bf0e5b12bc6dac8f8a4b1eec) - I kept in the same file just because I felt that the implementations would be more readable to have together. Let me know what you think

src/riscv/lib/src/jit.rs
NSant215: thanks! done [here](https://github.com/tezos/riscv-pvm/compare/26d7725d8a14caa5bf0e5b12bc6dac8f8a4b1eec..2db4038956e35b131994323ad5bc7281ef24317d)

src/riscv/lib/src/jit.rs
NSant215: thanks! done [here](https://github.com/tezos/riscv-pvm/compare/26d7725d8a14caa5bf0e5b12bc6dac8f8a4b1eec..2db4038956e35b131994323ad5bc7281ef24317d)

vapourismo (approved):

PR #276, opened 2025-08-27
Memory governance request sent to jit

HantangSun:
Closes RV-760

# What

Main thread sends memory request to jit

# Why

This is part of the effort to refactor jit and give jit access to memory. The main thread needs to notify the jit which pages are valid for jit compilation(page with r+x permission)

# How

This MR does the following things

- Introduce a new memory permission request type
- Add memory governance trait for the block builder 
- Pass machine state to handler for mmap/munmap/mprotect and to memory functions
- Main thread send memory permission request to jit when `fn protect` is called

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/machine_state.rs
+++ b/src/riscv/lib/src/machine_state.rs
@@ -596,6 +596,19 @@ impl<MC: memory::MemoryConfig, BCC: BlockCacheConfig, B: Block<MC, M>, M: backen
         Ok(())
     }
 
+    /// Mark all memory as readable and writeable.
+    #[cfg(test)]
+    pub fn set_all_readable_writeable<MB, const PAGES: usize, const TOTAL_BYTES: usize>(&mut self)
+    where
+        M: backend::ManagerReadWrite,
+        MB: memory::buddy::Buddy<M>,
+        MC: MemoryConfig<State<M> = memory::state::MemoryImpl<PAGES, TOTAL_BYTES, MB, M>>,
+    {
+        self.core
+            .main_memory
+            .set_all_readable_writeable(&mut self.block_builder);
+    }
+
     /// Invalidate all instruction caches that are sensitive to
     /// instruction fences.
     fn invalidate_instruction_caches(&mut self)
@@ -1262,7 +1275,7 @@ mod tests {
         );
 
         state.reset();
-        state.core.main_memory.set_all_readable_writeable();
+        state.set_all_readable_writeable();
 
         let stack_top = M4K::TOTAL_BYTES as u64;
         state.core.hart.xregisters.write(sp, stack_top);
@@ -1283,7 +1296,7 @@ mod tests {
         );
 
         state.reset();
-        state.core.main_memory.set_all_readable_writeable();
+        state.set_all_readable_writeable();
 
         let stack_top = M4K::TOTAL_BYTES as u64;
         state.core.hart.xregisters.write(sp, stack_top);
--- a/src/riscv/lib/src/machine_state/block_cache/block.rs
+++ b/src/riscv/lib/src/machine_state/block_cache/block.rs
@@ -25,6 +25,7 @@ use crate::machine_state::StepManyResult;
 use crate::machine_state::instruction::Instruction;
 use crate::machine_state::memory::Address;
 use crate::machine_state::memory::MemoryConfig;
+use crate::machine_state::memory::MemoryGovernanceListener;
 use crate::state::NewState;
 use crate::state_backend::AllocatedOf;
 use crate::state_backend::Atom;
@@ -50,7 +51,7 @@ pub trait Block<MC: MemoryConfig, M: ManagerBase>: NewState<M> {
     ///
     /// `Sized` bound is required to ensure any reference to `BlockBuilder` will be thin -
     /// see [`dispatch::DispatchFn`].
-    type BlockBuilder: Default + Sized;
+    type BlockBuilder: Default + Sized + MemoryGovernanceListener;
 
     /// Bind the block to the given allocated state.
     fn bind(allocated: AllocatedOf<BlockLayout, M>) -> Self
--- a/src/riscv/lib/src/machine_state/block_cache/block/dispatch.rs
+++ b/src/riscv/lib/src/machine_state/block_cache/block/dispatch.rs
@@ -6,6 +6,7 @@
 //! This module exposes wrappers for the style of dispatch and compilation that is done.
 
 use std::marker::PhantomData;
+use std::ops::RangeInclusive;
 use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::atomic::AtomicUsize;
@@ -21,6 +22,8 @@ use crate::machine_state::MachineCoreState;
 use crate::machine_state::instruction::Instruction;
 use crate::machine_state::memory::Address;
 use crate::machine_state::memory::MemoryConfig;
+use crate::machine_state::memory::MemoryGovernanceListener;
+use crate::machine_state::memory::PageState;
 use crate::state_backend::owned_backend::Owned;
 
 /// The function signature for dispatching a block run.
@@ -173,7 +176,7 @@ impl<C: CodeDispatcher<D, MC>, D: DispatchCompiler<MC>, MC: MemoryConfig> Defaul
 
 /// A compiler that can JIT-compile blocks of instructions, and hot-swap the execution of
 /// said block in the given dispatch target.
-pub trait DispatchCompiler<MC: MemoryConfig>: Default + Sized {
+pub trait DispatchCompiler<MC: MemoryConfig>: Default + Sized + MemoryGovernanceListener {
     /// Whether compilation should be attempted for the block.
     fn should_compile<C: CodeDispatcher<Self, MC>>(
         &self,
@@ -198,6 +201,16 @@ pub trait DispatchCompiler<MC: MemoryConfig>: Default + Sized {
 /// JIT compiler for blocks that performs compilation inline, in the same thread as execution.
 pub struct InlineCompiler<MC: MemoryConfig> {
     jit: JIT<MC>,
+}
+
+impl<MC: MemoryConfig> MemoryGovernanceListener for InlineCompiler<MC> {
+    fn handle_memory_permission_change(
+        &mut self,
+        _pages: RangeInclusive<usize>,
+        _state: PageState,
+    ) {
+        // TODO RV-761 - Implement handling of memory permissions in the JIT.
+    }
 }
 
 impl<MC: MemoryConfig> Default for InlineCompiler<MC> {
@@ -280,7 +293,18 @@ pub struct OutlineCompiler<MC: MemoryConfig> {
     // a reference to it - to ensure it is not dropped before we are done with execution,
     // even if the background compilation thread panics.
     _do_not_use_this_is_for_drop_only: Arc<Mutex<internal_corro::SendWrapper<JIT<MC>>>>,
-    sender: Sender<CompilationRequest>,
+    sender: Sender<JitRequest>,
+}
+
+impl<MC: MemoryConfig> MemoryGovernanceListener for OutlineCompiler<MC> {
+    fn handle_memory_permission_change(&mut self, pages: RangeInclusive<usize>, state: PageState) {
+        let _ = self
+            .sender
+            .send(JitRequest::MemoryPermissions(MemoryPermissionsRequest {
+                pages,
+                state,
+            }));
+    }
 }
 
 impl<MC: MemoryConfig + Send> OutlineCompiler<MC> {
@@ -301,25 +325,39 @@ impl<MC: MemoryConfig + Send> OutlineCompiler<MC> {
                 let jit = unsafe { jit_guard.as_mut() };
 
                 while let Ok(msg) = receiver.recv() {
-                    if let Some(jitfn) = jit.compile(&msg.instr, msg.program_counter) {
-                        debug_assert_eq!(
-                            msg.fun.load(Ordering::Acquire),
-                            Jitted::<Self, MC>::run_block_not_compiled as usize,
-                            "Unexpected function pointer in dispatch target"
-                        );
+                    match msg {
+                        JitRequest::Compile(CompilationRequest {
+                            instr,
+                            fun,
+                            program_counter,
+                        }) => {
+                            if let Some(jitfn) = jit.compile(&instr, program_counter) {
+                                debug_assert_eq!(
+                                    fun.load(Ordering::Acquire),
+                                    Jitted::<Self, MC>::run_block_not_compiled as usize,
+                                    "Unexpected function pointer in dispatch target"
+                                );
 
-                        // Safety: this function will be retrieved as a DispatchFn, rather than a
-                        // JitFn. The two function signatures are identical, apart from the first and
-                        // last parameters. These are both thin-pointers, and ignored by the JitFn.
-                        //
-                        // It's therefore safe to cast this function pointer to an identical ABI, where
-                        // this first and last parameter are thin-references to any value. This is the
-                        // case for both `Jitted` and `Jitted::BlockBuilder` which are both Sized.
-                        //
-                        // See <https://doc.rust-lang.org/std/primitive.fn.html#abi-compatibility> for more
-                        // information on ABI compatability.
-                        msg.fun.store(jitfn as usize, Ordering::Release);
-                    };
+                                // Safety: this function will be retrieved as a DispatchFn, rather than a
+                                // JitFn. The two function signatures are identical, apart from the first and
+                                // last parameters. These are both thin-pointers, and ignored by the JitFn.
+                                //
+                                // It's therefore safe to cast this function pointer to an identical ABI, where
+                                // this first and last parameter are thin-references to any value. This is the
+                                // case for both `Jitted` and `Jitted::BlockBuilder` which are both Sized.
+                                //
+                                // See <https://doc.rust-lang.org/std/primitive.fn.html#abi-compatibility> for more
+                                // information on ABI compatability.
+                                fun.store(jitfn as usize, Ordering::Release);
+                            };
+                        }
+                        JitRequest::MemoryPermissions(MemoryPermissionsRequest {
+                            pages: _,
+                            state: _,
+                        }) => {
+                            // TODO RV-761 - Implement handling of memory permissions in the JIT.
+                        }
+                    }
                 }
             }
             // because we used blocking recv with an asynchronous channel, this only fails when the
@@ -385,7 +423,7 @@ impl<MC: MemoryConfig + Send> DispatchCompiler<MC> for OutlineCompiler<MC> {
         // NB - any blocks already JIT compiled are safe to keep calling, as the
         // data behind the mutex (the JIT) is kept alive for as long as we maintain
         // our reference to it, despite the lock itself being poisoned.
-        let _ = self.sender.send(request);
+        let _ = self.sender.send(JitRequest::Compile(request));
 
         fun
     }
@@ -395,4 +433,16 @@ struct CompilationRequest {
     instr: Vec<Instruction>,
     fun: Arc<AtomicUsize>,
     program_counter: Address,
+}
+
+// TODO RV-761 - Implement handling of memory permissions in the JIT.
+#[expect(unused, reason = "Will Be Used Soon")]
+struct MemoryPermissionsRequest {
+    pages: RangeInclusive<usize>,
+    state: PageState,
+}
+
+enum JitRequest {
+    Compile(CompilationRequest),
+    MemoryPermissions(MemoryPermissionsRequest),
 }
--- a/src/riscv/lib/src/machine_state/block_cache/block/interpreted.rs
+++ b/src/riscv/lib/src/machine_state/block_cache/block/interpreted.rs
@@ -5,6 +5,8 @@
 
 //! Interpreted blocks of instructions
 
+use std::ops::RangeInclusive;
+
 use perfect_derive::perfect_derive;
 
 use crate::default::ConstDefault;
@@ -18,6 +20,8 @@ use crate::machine_state::block_cache::block::run_block_inner;
 use crate::machine_state::instruction::Instruction;
 use crate::machine_state::memory::Address;
 use crate::machine_state::memory::MemoryConfig;
+use crate::machine_state::memory::MemoryGovernanceListener;
+use crate::machine_state::memory::PageState;
 use crate::state::NewState;
 use crate::state_backend::AllocatedOf;
 use crate::state_backend::Cell;
@@ -34,6 +38,16 @@ use crate::traps::Exception;
 /// Interpreted blocks are built automatically, and require no additional context.
 #[derive(Debug, Default)]
 pub struct InterpretedBlockBuilder;
+
+impl MemoryGovernanceListener for InterpretedBlockBuilder {
+    fn handle_memory_permission_change(
+        &mut self,
+        _pages: RangeInclusive<usize>,
+        _state: PageState,
+    ) {
+        // do nothing
+    }
+}
 
 /// Blocks that are executed via interpreting the individual instructions.
 ///
--- a/src/riscv/lib/src/machine_state/memory.rs
+++ b/src/riscv/lib/src/machine_state/memory.rs
@@ -2,12 +2,14 @@
 //
 // SPDX-License-Identifier: MIT
 
-mod buddy;
+pub(crate) mod buddy;
 mod config;
 mod protection;
-mod state;
+pub(crate) mod state;
 
 use std::num::NonZeroU64;
+use std::num::NonZeroUsize;
+use std::ops::RangeInclusive;
 
 use tezos_smart_rollup_constants::riscv::SbiError;
 
@@ -54,6 +56,15 @@ pub const PAGE_OFFSET_MASK: u64 = (1 << OFFSET_BITS.get()) - 1;
 
 /// Mask for obtaining a page-aligned-down address.
 pub const PAGE_MASK: u64 = !0 << OFFSET_BITS.get();
+
+/// Calculate the page range for a given address and length.
+pub fn get_page_range(address: Address, length: NonZeroUsize) -> std::ops::RangeInclusive<usize> {
+    let address = address as usize;
+    let start_page = address >> OFFSET_BITS.get();
+
+    let end_page = address.wrapping_add(length.get()).saturating_sub(1) >> OFFSET_BITS.get();
+    start_page..=end_page
+}
 
 /// Memory address
 pub type Address = XValue;
@@ -242,6 +253,7 @@ pub trait Memory<M: ManagerBase>: NewState<M> + Sized {
         address: Address,
         length: usize,
         perms: Permissions,
+        memory_governance_listener: &mut impl MemoryGovernanceListener,
     ) -> Result<(), MemoryGovernanceError>
     where
         M: ManagerWrite;
@@ -272,6 +284,7 @@ pub trait Memory<M: ManagerBase>: NewState<M> + Sized {
         length: usize,
         perms: Permissions,
         allow_replace: bool,
+        memory_governance_listener: &mut impl MemoryGovernanceListener,
     ) -> Result<Address, MemoryGovernanceError>
     where
         M: ManagerReadWrite;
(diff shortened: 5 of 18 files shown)


src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -304,6 +323,7 @@ where
         length: usize,
         perms: Permissions,
         allow_replace: bool,
+        block_builder: &impl MemoryGovernance,
```
vapourismo, line 326: Same here, name seems not appropriate.

src/riscv/lib/src/machine_state/block_cache/block/dispatch.rs
```diff
@@ -131,6 +132,18 @@ pub struct InlineCompiler<MC: MemoryConfig> {
...
 
+impl<MC: MemoryConfig> MemoryGovernance for InlineCompiler<MC> {
+    fn modify_memory_permissions(
+        &self,
+        start_page_index: u64,
+        end_page_index: u64,
+        validate: bool,
+    ) {
+        self.jit
+            .modify_memory_permissions(start_page_index, end_page_index, validate);
```
vapourismo, line 143: Is the `JIT` type an appropriate one to forward this request to? Shouldn't there be a `JITRouter` in the middle?

src/riscv/lib/src/machine_state/block_cache/block/dispatch.rs
```diff
@@ -107,7 +108,7 @@ impl<D: DispatchCompiler<MC>, MC: MemoryConfig> Default for DispatchTarget<D, MC
 
 /// A compiler that can JIT-compile blocks of instructions, and hot-swap the execution of
 /// said block in the given dispatch target.
-pub trait DispatchCompiler<MC: MemoryConfig>: Default + Sized {
+pub trait DispatchCompiler<MC: MemoryConfig>: Default + Sized + MemoryGovernance {
```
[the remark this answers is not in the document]
HantangSun: In jitted.rs, we have 
```
impl<D: DispatchCompiler<MC>, MC: MemoryConfig> Block<MC, Owned> for Jitted<D, MC> {
    type BlockBuilder = D;
```

So I think the dispatch compiler needs to implement the MemoryGovernance trait in blockbuilder

src/riscv/lib/src/machine_state/block_cache/block.rs
```diff
@@ -40,6 +40,11 @@ use crate::traps::Exception;
 /// State Layout for Blocks
 pub type BlockLayout = (Atom<u8>, [Atom<Instruction>; CACHE_INSTR]);
 
+/// Trait for modifying the memory permissions of a memory range.
+pub trait MemoryGovernance {
+    fn modify_memory_permissions(&self, start_page_index: u64, end_page_index: u64, validate: bool);
```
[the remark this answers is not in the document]
HantangSun: The function now takes a mutable reference. [Link](https://github.com/tezos/riscv-pvm/pull/276/commits/e97b4da1abff46b8246badafbd2be44c65768c30#diff-fd7388d4ee4bd01b5523ffe06c6dfefbcd708ef88707fb6f23abb107a0697a81R323-R331)

src/riscv/lib/src/pvm/linux/memory.rs
```diff
@@ -96,9 +102,9 @@ impl<M: ManagerBase> SupervisorState<M> {
         clippy::too_many_arguments,
         reason = "The system call dispatch mechanism needs these arguments to exist, they can't be on a nested structure"
     )]
-    pub(super) fn handle_mmap<MC>(
+    pub(super) fn handle_mmap<MC, BCC: BlockCacheConfig, B: Block<MC, M>>(
```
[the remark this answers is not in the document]
HantangSun: Fixed [here](https://github.com/tezos/riscv-pvm/pull/276/commits/e97b4da1abff46b8246badafbd2be44c65768c30#diff-1dc59f1eedb8a2b5917a02ddf42e7a2117036af5b0a47cfd6cc8d29ad3f42159R107-L118)

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -512,7 +517,7 @@ impl<M: ManagerBase> SupervisorState<M> {
     /// Handle a Linux system call.
     pub fn handle_system_call<MC, BCC, B>(
         &mut self,
-        machine: &mut MachineState<MC, BCC, B, M>,
+        mut machine: &mut MachineState<MC, BCC, B, M>,
```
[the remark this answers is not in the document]
HantangSun: Fixed [here](https://github.com/tezos/riscv-pvm/pull/276/commits/e97b4da1abff46b8246badafbd2be44c65768c30#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7R518-R523)

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -55,8 +56,15 @@ impl<const PAGES: usize, const TOTAL_BYTES: usize, B, M: ManagerBase>
...
     {
-        self.protect_pages(0, TOTAL_BYTES, Permissions::READ_WRITE)
-            .unwrap();
+        use crate::machine_state::block_cache::block::InterpretedBlockBuilder;
+
+        self.protect_pages(
+            0,
+            TOTAL_BYTES,
+            Permissions::READ_WRITE,
+            None::<&InterpretedBlockBuilder>,
```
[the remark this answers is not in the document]
HantangSun: Introduced a dummy handler type and used [here](https://github.com/tezos/riscv-pvm/commit/114af5df9ba7ca731632a22fe7ababc9dda3964f#diff-d6c8df0e6a878ab6d390943ae4ad6f2c8c43b4bc830849eba502c957a6e3ea71R61-R65)

src/riscv/lib/src/machine_state/block_cache/block/dispatch.rs
HantangSun: Yes, this will changed to be a JitRouter in the middle, but I don't think this change belongs to this MR. It makes more sense to do changes on the jit side in a seperate MR

src/riscv/lib/src/machine_state/memory.rs
```diff
@@ -307,6 +317,33 @@ pub trait MemoryConfig: Send + 'static {
         F: FnManager<Ref<'a, M>>;
 }
 
+/// Trait that handles memory permission updates for the jit.
+/// The jit router needs to know when the memory permissions
+/// are updated so it can invalidate the jit instances.
```
vapourismo, line 322: Let's not overload the description of this trait with details of the trait implementations.

```suggestion
/// Implementing types will be informed about changes to memory permissions
```

src/riscv/lib/src/machine_state/memory.rs
```diff
@@ -307,6 +317,33 @@ pub trait MemoryConfig: Send + 'static {
...
+pub trait MemoryGovernanceListener {
+    fn handle_memory_permission_change(
```
vapourismo, line 324: This misses a documentation comment.

```suggestion
    /// Called on changes to memory permissions.
    fn handle_memory_permission_change(
```

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -55,8 +56,15 @@ impl<const PAGES: usize, const TOTAL_BYTES: usize, B, M: ManagerBase>
...
     {
-        self.protect_pages(0, TOTAL_BYTES, Permissions::READ_WRITE)
-            .unwrap();
+        use crate::machine_state::memory::DummyMemoryGovernanceHandler;
+
+        self.protect_pages(
+            0,
+            TOTAL_BYTES,
+            Permissions::READ_WRITE,
+            &mut DummyMemoryGovernanceHandler,
```
vapourismo, line 65: Could this cause any problems with tests in the future? 

Should this method accept the listener parameter?

Or should it be moved higher in the hierarchy? For example, would it work in the `impl MachineState`?

src/riscv/lib/src/machine_state/block_cache/block/interpreted.rs
```diff
@@ -35,6 +36,17 @@ use crate::traps::Exception;
...
 pub struct InterpretedBlockBuilder;
 
+impl MemoryGovernanceListener for InterpretedBlockBuilder {
+    fn handle_memory_permission_change(
+        &mut self,
+        _start_page_index: u64,
+        _end_page_index: u64,
+        _validate: bool,
+    ) {
+        // do nothing
```
vapourismo, line 46: :)

vapourismo (changes requested):

src/riscv/lib/src/machine_state/block_cache/block/dispatch.rs
vapourismo: I understand if this were calling the JIT router, but it isn't right now. The code is forwarding to a JIT compiler instance. It seems you're introducing an empty method and a redundant call only to remove it in a subsequent PR.

HantangSun: Ok, removed [here](https://github.com/tezos/riscv-pvm/compare/99d85bb4438b33dc9deab7210585b9264bf410c8..29f948d38a6465bc01c585a4724a7a309cf73fd0)

vapourismo (changes requested):

src/riscv/lib/src/machine_state/block_cache/block.rs
```diff
@@ -40,6 +40,10 @@ use crate::traps::Exception;
 /// State Layout for Blocks
 pub type BlockLayout = (Atom<u8>, [Atom<Instruction>; CACHE_INSTR]);
 
+pub trait MemoryGovernance {
+    fn modify_memory_permissions(&self, start_page_index: u64, end_page_index: u64, validate: bool);
```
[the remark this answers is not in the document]
HantangSun: This has already been updated. [Link](https://github.com/tezos/riscv-pvm/pull/276/files#diff-fd7388d4ee4bd01b5523ffe06c6dfefbcd708ef88707fb6f23abb107a0697a81R329-R333)

src/riscv/lib/src/pvm/common.rs
```diff
@@ -544,7 +544,7 @@ mod tests {
         pvm.machine_state
             .core
             .main_memory
-            .set_all_readable_writeable();
+            .set_all_readable_writeable(&mut pvm.machine_state.block_builder);
```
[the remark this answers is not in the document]
HantangSun: set_all_readable_writeable is for MemoryImpl and not part of the Memory trait. Therefore, to implement set_all_readable_writeable for machine state, protect_pages need to be called directly. [Done here](https://github.com/tezos/riscv-pvm/compare/c881ae66046dd271299182f05f89ce0367cee5a1..51a8efa7e44a8db9f0d730b873e4efafac55249b)

src/riscv/lib/src/machine_state/memory.rs
```diff
@@ -275,12 +277,20 @@ pub trait Memory<M: ManagerBase>: NewState<M> + Sized {
...
         self.deallocate_pages(address, length)?;
-        self.protect_pages(address, length, Permissions::NONE)
+        self.protect_pages(
+            address,
+            length,
+            Permissions::NONE,
+            memory_governance_listener,
+        )?;
+
+        Ok(())
```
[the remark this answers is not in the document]
emturner: it's not changed on the latest version, for me.

src/riscv/lib/src/machine_state/block_cache/block.rs
```diff
@@ -54,7 +51,7 @@ pub trait Block<MC: MemoryConfig, M: ManagerBase>: NewState<M> {
     ///
     /// `Sized` bound is required to ensure any reference to `BlockBuilder` will be thin -
     /// see [`dispatch::DispatchFn`].
-    type BlockBuilder: Default + Sized + MemoryGovernance;
+    type BlockBuilder: Default + Sized + MemoryGovernanceListener;
```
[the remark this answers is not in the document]
emturner: > Yeah, fixing up stacked commit can be tricky so I usually apply all the fixes to the last commit. I > can squash the two commits into one if that is preferred

I'd either do that, or spend the time to maintain two separate commits - which I acknowledge will take a little longer. Up to you.

src/riscv/lib/src/machine_state/memory.rs
```diff
@@ -307,6 +317,31 @@ pub trait MemoryConfig: Send + 'static {
...
+
+/// Implementing types will be informed about changes to memory permissions
+pub trait MemoryGovernanceListener {
+    /// Called on changes to memory permissions.
+    fn handle_memory_permission_change(&mut self, pages: Range<u64>, state: PageState);
+}
+
+/// Default memory governance handler that does nothing
+#[derive(Default)]
+pub struct DummyMemoryGovernanceHandler;
```
emturner, line 337: should this be test-only?

src/riscv/lib/src/machine_state.rs
```diff
@@ -547,6 +547,23 @@ impl<MC: memory::MemoryConfig, BCC: BlockCacheConfig, B: Block<MC, M>, M: backen
...
 
+    /// Mark all memory as readable and writeable.
+    #[cfg(test)]
+    pub fn set_all_readable_writeable(&mut self)
+    where
+        M: backend::ManagerReadWrite,
+    {
+        self.core
+            .main_memory
+            .protect_pages(
```
emturner, line 558: have you considered raising `set_all_readable_writeable` to the `Memory` trait here, or an alternative?

emturner: alternatively, is it still worth having `set_all_readable_writeable` on the MemoryImpl if (presumably) near everything will be migrated to this method?

HantangSun: I am not sure if moving `set_all_readable_writeable` from `MemoryImpl` to `Memory` is doable. The `Memory` trait does not have context on the total size.

HantangSun: Discussed offline. Method moved to the Memory trait. [Link](https://github.com/tezos/riscv-pvm/compare/111b4de62664e9351c5c5943f7ecda17aca35b25..146b870643ed8737b9991a9423ee7d9db693aff2)

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -240,6 +250,17 @@ where
         self.executable_pages
             .modify_access(address, length, perms.can_exec());
 
+        let start_page_index = address / PAGE_SIZE.get();
+        let end_page_index = (address + length as u64).div_ceil(PAGE_SIZE.get());
```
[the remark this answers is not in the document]
emturner: That is only true for `address % PAGE_SIZE == 0` though?

emturner (commented):
might need a test around the conversion from `address + length` into `page_start..page_end`

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -240,6 +232,19 @@ where
         self.executable_pages
             .modify_access(address, length, perms.can_exec());
 
+        let address = address as usize;
+        let start_page = address >> super::OFFSET_BITS;
+        let end_page = address.wrapping_add(length).wrapping_sub(1) >> super::OFFSET_BITS;
```
vapourismo, line 237: This calculation assumes an inclusive range, but you're constructing an exclusive one below.

vapourismo (changes requested):

HantangSun:
I have re-worked the page range logic as discussed offline. On the using NonZeroUsize type, I believe it's not really in scope of this MR.

src/riscv/lib/src/machine_state/memory/protection.rs
```diff
@@ -115,19 +117,16 @@ impl<const PAGES: usize, M: ManagerBase> PagePermissions<PAGES, M> {
     }
 
     /// Change the access permissions for the given range.
-    pub fn modify_access(&mut self, address: Address, length: usize, accessible: bool)
+    pub fn modify_access(&mut self, page_range: &Range<usize>, accessible: bool)
     where
         M: ManagerWrite,
     {
-        if length < 1 {
+        if page_range.is_empty() {
```
emturner, line 124: this isn't needed, if the range is empty we just won't touch memory

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -17,6 +21,14 @@ use crate::state_backend::ManagerRead;
 use crate::state_backend::ManagerReadWrite;
 use crate::state_backend::ManagerWrite;
 
+fn get_page_range(address: Address, length: usize) -> Range<usize> {
```
emturner, line 24: nit, using `RangeInclusive` would save the `saturating_add(1)`, which isn't needed below

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -227,18 +231,30 @@ where
...
         self.writable_pages
-            .modify_access(address, length, perms.can_write());
+            .modify_access(page_range, perms.can_write());
         self.executable_pages
-            .modify_access(address, length, perms.can_exec());
+            .modify_access(page_range, perms.can_exec());
+
+        memory_governance_listener.handle_memory_permission_change(
+            page_range.clone(),
+            if perms.can_read() && perms.can_exec() && !perms.can_write() {
```
emturner, line 252: feels like there should possibly be a `From<Permissions> for PageState` impl?

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -227,18 +231,30 @@ where
         address: Address,
         length: usize,
         perms: Permissions,
+        memory_governance_listener: &mut impl MemoryGovernanceListener,
     ) -> Result<(), super::MemoryGovernanceError>
     where
         M: ManagerWrite,
     {
         Self::check_bounds(address, length, super::MemoryGovernanceError)?;
```
emturner, line 239: if `length < 1, return` check here? (if not switching to NonZeroU64)

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -17,6 +21,14 @@ use crate::state_backend::ManagerRead;
 use crate::state_backend::ManagerReadWrite;
 use crate::state_backend::ManagerWrite;
 
+fn get_page_range(address: Address, length: usize) -> Range<usize> {
+    let address = address as usize;
+    let start_page = address >> super::OFFSET_BITS;
+    let end_page =
+        (address.wrapping_add(length).saturating_sub(1) >> super::OFFSET_BITS).saturating_add(1);
+    start_page..end_page
+}
```
vapourismo, line 30: Nit: I would expect this function along side the generic memory functions in the `memory` module. Would you mind moving it there?

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -17,6 +21,14 @@ use crate::state_backend::ManagerRead;
 use crate::state_backend::ManagerReadWrite;
 use crate::state_backend::ManagerWrite;
 
+fn get_page_range(address: Address, length: usize) -> Range<usize> {
+    let address = address as usize;
+    let start_page = address >> super::OFFSET_BITS;
+    let end_page =
+        (address.wrapping_add(length).saturating_sub(1) >> super::OFFSET_BITS).saturating_add(1);
```
vapourismo, line 28: There is an off-by-one error when `length == 0`. 

When length is 0:
- start address 4095 results in `0..1` (incorrect)
- start address 4096 results in `1..1` (correct)
- start address 4097 results in `1..2` (incorrect)

vapourismo (changes requested):

src/riscv/lib/src/machine_state/memory/protection.rs
```diff
@@ -115,19 +117,16 @@ impl<const PAGES: usize, M: ManagerBase> PagePermissions<PAGES, M> {
     }
 
     /// Change the access permissions for the given range.
-    pub fn modify_access(&mut self, address: Address, length: usize, accessible: bool)
+    pub fn modify_access(&mut self, page_range: &Range<usize>, accessible: bool)
```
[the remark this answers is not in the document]
HantangSun: Changed [here](https://github.com/tezos/riscv-pvm/compare/5a215bbb51af7c71c1dd7d0e1cd2cad7ad928234..0c975380ee930d7b902fac478d76615be098493a#diff-e46c8f5f20aadbee4463ee3e0baed62d16d060553501a46a0d2b77e6e591eef2R120-R126)

src/riscv/lib/src/machine_state/memory/protection.rs
HantangSun: Removed [here](https://github.com/tezos/riscv-pvm/compare/5a215bbb51af7c71c1dd7d0e1cd2cad7ad928234..0c975380ee930d7b902fac478d76615be098493a#diff-e46c8f5f20aadbee4463ee3e0baed62d16d060553501a46a0d2b77e6e591eef2R120-R126)

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -227,18 +231,30 @@ where
...
         length: usize,
         perms: Permissions,
+        memory_governance_listener: &mut impl MemoryGovernanceListener,
     ) -> Result<(), super::MemoryGovernanceError>
     where
         M: ManagerWrite,
     {
         Self::check_bounds(address, length, super::MemoryGovernanceError)?;
 
+        let page_range = &get_page_range(address, length);
```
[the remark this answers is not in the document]
HantangSun: Fixed [here](https://github.com/tezos/riscv-pvm/compare/5a215bbb51af7c71c1dd7d0e1cd2cad7ad928234..0c975380ee930d7b902fac478d76615be098493a#diff-d6c8df0e6a878ab6d390943ae4ad6f2c8c43b4bc830849eba502c957a6e3ea71R236-R243)

src/riscv/lib/src/machine_state/memory/state.rs
HantangSun: Done [here](https://github.com/tezos/riscv-pvm/compare/5a215bbb51af7c71c1dd7d0e1cd2cad7ad928234..0c975380ee930d7b902fac478d76615be098493a#diff-d6c8df0e6a878ab6d390943ae4ad6f2c8c43b4bc830849eba502c957a6e3ea71R236-R243)

src/riscv/lib/src/machine_state/memory/state.rs
HantangSun: Done [here](https://github.com/tezos/riscv-pvm/compare/5a215bbb51af7c71c1dd7d0e1cd2cad7ad928234..0c975380ee930d7b902fac478d76615be098493a#diff-d6c8df0e6a878ab6d390943ae4ad6f2c8c43b4bc830849eba502c957a6e3ea71R236-R243)

src/riscv/lib/src/machine_state/memory.rs
```diff
@@ -49,6 +50,23 @@ pub const PAGE_SIZE: NonZeroU64 = {
...
+#[expect(
+    clippy::reversed_empty_ranges,
+    reason = "An empty range is expected with zero length"
+)]
+/// Calculate the page range for a given address and length.
+pub fn get_page_range(address: Address, length: usize) -> std::ops::RangeInclusive<usize> {
+    let address = address as usize;
+    let start_page = address >> OFFSET_BITS;
+
+    if length == 0 {
```
vapourismo, line 62: Why don't you accept `length: NonZeroUsize` like we discussed in the weekly meeting?

vapourismo (changes requested):

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -227,18 +226,25 @@ where
...
     {
         Self::check_bounds(address, length, super::MemoryGovernanceError)?;
 
-        self.readable_pages
-            .modify_access(address, length, perms.can_read());
-        self.writable_pages
-            .modify_access(address, length, perms.can_write());
-        self.executable_pages
-            .modify_access(address, length, perms.can_exec());
+        if let Some(length) = NonZeroUsize::new(length) {
```
vapourismo, line 236: If length is zero, you don't need to do anything, right? Might as well exit early if it is `None` via `let Some(length) = ... else { return Ok(()); }` at the start of the function. Or alternatively, this function could accept non-zero lengths too.

vapourismo (changes requested):

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -74,8 +67,14 @@ impl<const PAGES: usize, const TOTAL_BYTES: usize, B, M: ManagerBase>
         Self::check_bounds(address, length, BadMemoryAccess)?;
 
         self.data.write(address as usize, value);
-        self.readable_pages.modify_access(address, length, true);
-        self.executable_pages.modify_access(address, length, true);
+
+        if let Some(length) = NonZeroUsize::new(length) {
```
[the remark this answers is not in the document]
HantangSun: Move the zero check to the start of the function. [Link](https://github.com/tezos/riscv-pvm/compare/460fd69bdbbd63ef7b34128225ad43d586277c31..8fd4d2bae2669fbe3b7da4b98f103a28541ea16f#diff-d6c8df0e6a878ab6d390943ae4ad6f2c8c43b4bc830849eba502c957a6e3ea71R237-R240)

src/riscv/lib/src/pvm/common.rs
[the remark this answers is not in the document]
HantangSun: It should be possible but I don't think it is particularly easy to do. We can't use trait bound for `MachineState` here since MemoryImpl is a struct rather than a trait

src/riscv/lib/src/machine_state/memory/state.rs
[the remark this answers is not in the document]
vapourismo: I'm not sure I see the change you reference.

src/riscv/lib/src/pvm/common.rs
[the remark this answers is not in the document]
vapourismo: Here is an example of how to do it: https://github.com/tezos/riscv-pvm/commit/4700a6367eb145070608bdf06a9de204a66a7306

vapourismo (changes requested):

src/riscv/lib/src/machine_state/memory.rs
```diff
@@ -313,6 +339,47 @@ pub trait MemoryConfig: Send + 'static {
...
+/// Implementing types will be informed about changes to memory permissions
+pub trait MemoryGovernanceListener {
+    /// Called on changes to memory permissions.
+    fn handle_memory_permission_change(&mut self, pages: RangeInclusive<usize>, state: PageState);
+}
+
+/// Default memory governance handler that does nothing
+#[cfg(test)]
+#[derive(Default)]
+pub struct DummyMemoryGovernanceHandler;
```
felixp72, line 370: Why would we want to have tests use a dummy handler that is a no-op for memory governance. Especially since you added the test-only method `set_all_readable_writeable` which should be used in tests to make memory handling easier. You either want diverging behaviour with a dummy handler or introduce helper functions to help use the actual types used in production, I don't see why you would have both.

In tests, probably what you want to do is, depending if you are using the jit or not, to pass the proper handler. Which currently should be no-ops.

src/riscv/lib/src/machine_state/memory/state.rs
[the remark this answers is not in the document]
HantangSun: The code is now 
```
        let Some(non_zero_length) = NonZeroUsize::new(length) else {
            return Ok(());
        };

        Self::check_bounds(address, length, super::MemoryGovernanceError)?;
```

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -337,12 +352,31 @@ where
...
+        B: Buddy<M>,
+        M: ManagerReadWrite,
+    {
+        self.protect_pages(
+            0,
+            TOTAL_BYTES,
+            Permissions::READ_WRITE,
+            memory_governance_listener,
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
HantangSun: Fixed [here](https://github.com/tezos/riscv-pvm/compare/098a14d50beec6bfe9d9bc9099ddef20a02e0b4a..9432b3c5c7d72da3d596fe25467bc71278cb5584#diff-d6c8df0e6a878ab6d390943ae4ad6f2c8c43b4bc830849eba502c957a6e3ea71R99)

src/riscv/lib/src/machine_state/memory.rs
felixp72: From offline: This is needed because some tests use a lower-level API (`MachineState` vs `MachineCoreState`), so it make sense to have this one, at least for now probably it's not worth the effort of getting rid of this dummy.

felixp72 (commented):
Probably was addressed in previous comments, but why isn't the `MemoryGovernanceListener` a generic parameter to `MemoryImpl`/`Memory`? (Or the state in general, similar to the type of compiler / block builder used). Feels like this will leave space for possibly calling the interpreted block handler when the jit is actually used.

PR #279, opened 2025-08-28
interpreter: add OpCode::ForceFetchRun

emturner:
# What

- Part of RV-758
- Related to RV-764

Add exception-raising instruction that forces a fetch_instr/run_instr cycle to occur for the current program counter.

# Why

This will be used by the page cache to ensure that we remain synced with main memory, in the case that an uncompressed instruction would overlap pages.

This allows us to invalidate a range of pages, without having to worry about the page immediately preceding the invalidated range.

# How

Addition-only MR. The handling of the exception will be done one this and #277 are merged.

# Manually Testing

```
make all
```

# Tasks for the Author


kurtisc (changes requested):
I'm surprised to see this done by adding a new instruction, but from what I can find that seems to be allowed with RISC-V.

Please could you expand on why you've chosen this vs a 'typical' function?

PR #287, opened 2025-09-02
Ensure commits follow Conventional Commits spec

vapourismo:
# What

# Why

# How

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author


felixp72 (commented):
Thoughts on adding the conventional commits badge on the main README?

From the official repo:
```
[![Conventional Commits](https://img.shields.io/badge/Conventional%20Commits-1.0.0-%23FE5196?logo=conventionalcommits&logoColor=white)](https://conventionalcommits.org)
```

vapourismo:
> Thoughts on adding the conventional commits badge on the main README?
> 
> From the official repo:
> 
> ```
> [Conventional Commits](https://conventionalcommits.org)
> ```

What is the main README to you?

kurtisc:
I've seen this spec before and I don't really like the concept for this reason:

>Automatically generating CHANGELOGs.
>Automatically determining a semantic version bump (based on the types of commits landed).
>Triggering build and publish processes.

The first line of a commit message is 50 characters by convention. This is a very small space to put information and it's what's presented foremost, to humans, by various views.

Meanwhile there's all the rest of the message (and git tags and git notes) that are available for storing information relating to whatever. This is where git itself puts `Signed-off-by` and where other projects put `Approved-by`, `Acked-by`, `Change-Id`, ticket names etc.

The cost of any automation looking at the footer of a message vs the first line is approximately zero. So I don't think it makes sense to sacrifice the most human-readable part for metadata.

vapourismo:
> I've seen this spec before and I don't really like the concept for this reason:
> 
> > Automatically generating CHANGELOGs.
> > Automatically determining a semantic version bump (based on the types of commits landed).
> > Triggering build and publish processes.
> 
> The first line of a commit message is 50 characters by convention. This is a very small space to put information and it's what's presented foremost, to humans, by various views.
> 
> Meanwhile there's all the rest of the message (and git tags and git notes) that are available for storing information relating to whatever. This is where git itself puts `Signed-off-by` and where other projects put `Approved-by`, `Acked-by`, `Change-Id`, ticket names etc.
> 
> The cost of any automation looking at the footer of a message vs the first line is approximately zero. So I don't think it makes sense to sacrifice the most human-readable part for metadata.

I think the maximum allowed title length is 72 characters. 

I'm not sure if I understand your point about automation. The benefit of the Conventional Commits specification is that is useful for people, while also enabling automation that is beneficial for people as well.

Let's assume there is no automation involved at all: do you think the ability to add a type and a scope in addition to a consistently phrased description is good or bad? For me, this would be valuable in deciding whether is commit is relevant to me or not. Do you think it would overload the title with irrelevant information?

kurtisc:
https://git-scm.com/docs/git-commit
>Though not required, it’s a good idea to begin the commit message with a single short (no more than 50 characters) 

Something like `git log --oneline` will prefix the 'title' with a short hash, bringing it up to the 72 of the body. N.B. we're not actually following this guideline.

Looking at our git history, I see a lot of messages starting with a verb:
>Add
>Bump
>Update
>Fix
>Move

In these cases I don't think the type would give additional information to me because they'd be redundant. In fact the Conventional Commits spec has one of these:
>feat(lang): add Polish language

that could just be `Add Polish language` or `feat(lang): Polish` and you could probably get a list of commits that have modified languages by using a command like `git log src/languages.rs`.

If we adopt this then I hope we can avoid replicating the types with the verbs, because I think it is unnecessary clutter. I would still also welcome some discussion on normalising how we do commit messages and am interested in what cases you'd use this information to filter commits.

PR #303, opened 2025-09-04
Add support for using handler instead of signal disposition

kurtisc:
Relates to RV-756

# What

Adds support for using signal handlers instead of the default disposition.
Adds support for ignoring signals.

I recommend viewing by-commit.

# Why
Signals each have a default disposition which declares how they operate by default when a signal is raised.
# How
Some mistakes were made in the implementation of signal handlers. They are alternatives to the dispositions, rather than temporary diversions from them. Signals can do one of three things when they are raised:
 - The default action (`SIG_DFL`) (0)
 - Be ignored (`SIG_IGN`)
 - Call a handler
   - The handler can have additional parameters if it is registered using `rt_sigaction(2)` instead of `signal(2)`
   
This fixes those mistakes and adds support for the default behaviour.

`SIG_DFL` is used instead of `NO_HANDLER`

# Manually Testing

```
make all
```
Changing the flag to `SIG_DFL` causes the test to fail.

# Regressions

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/machine_state.rs
+++ b/src/riscv/lib/src/machine_state.rs
@@ -34,7 +34,6 @@ use crate::parser::is_compressed;
 use crate::parser::parse_compressed_instruction;
 use crate::parser::parse_uncompressed_instruction;
 use crate::program::Program;
-use crate::pvm::linux::signals::NO_HANDLER;
 use crate::pvm::linux::signals::Signal;
 use crate::pvm::linux::signals::SignalActions;
 use crate::pvm::linux::signals::SignalActionsLayout;
@@ -612,11 +611,7 @@ impl<MC: memory::MemoryConfig, BCC: BlockCacheConfig, B: Block<MC, M>, M: backen
     where
         M: ManagerReadWrite,
     {
-        // We only currently support the `action` field.
-        // TODO RV-732 Parse `sa_flags` subset.
-        let handler = self.core.signal_actions.read_action(signal);
-
-        if handler == NO_HANDLER || self.core.dispatch_signal(signal).is_err() {
+        if self.core.dispatch_signal(signal).is_err() {
             self.core.hart.pc.write(0);
         }
     }
--- a/src/riscv/lib/src/pvm/linux/signals.rs
+++ b/src/riscv/lib/src/pvm/linux/signals.rs
@@ -47,6 +47,8 @@ pub enum SignalError {
     Memory(#[from] BadMemoryAccess),
     #[error("Misaligned stack pointer")]
     MisalignedStackPointer,
+    #[error("Execution shall terminate")]
+    Terminate,
 }
 
 /// Linux sigaction struct, see <https://man7.org/linux/man-pages/man2/sigaction.2.html>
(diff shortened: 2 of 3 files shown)


src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -248,12 +259,19 @@ impl<MC: MemoryConfig, M: ManagerReadWrite> MachineCoreState<MC, M> {
...
         }
 
+        if handler == SIG_DFL {
+            match Disposition::default(signal) {
+                Disposition::Term | Disposition::Core => {
+                    return Err(SignalError::Terminate);
+                }
+                _ => return Ok(()),
+            };
+        }
```
vapourismo, line 273: Nit: I think a `match` block could be used instad of two ifs. Up to you though.

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -282,19 +300,9 @@ impl<MC: MemoryConfig, M: ManagerReadWrite> MachineCoreState<MC, M> {
...
-            | Signal::Sigbus
-            | Signal::Sigsegv
-            | Signal::Sigpipe
-            | Signal::Sigterm
-            | Signal::Sigsys => 0,
+        // For signals with a TERM or CORE disposition, we want to change the program counter to 0.
+        let pc = match Disposition::default(signal) {
+            Disposition::Term | Disposition::Core => 0,
             _ => self.hart.pc.read(),
         };
```
vapourismo, line 307: Does this mean you return to PC 0 regardless of the current disposition of `signal`?

vapourismo (changes requested):

src/riscv/lib/src/pvm/linux/signals.rs
kurtisc: Done on [L272](https://github.com/tezos/riscv-pvm/compare/6e38c344da094c0560aff6d602d0385ea5cd5340..c9466be2d0c82ce2a6e84eaf68a2856d915c35d8#diff-4293ba17c6d4b17be9dd006e0ca50f51760cc8ee0bc78d11aaaaef4f81bf74baR272), apologies for diff munged with a rebase.

src/riscv/lib/src/machine_state.rs
```diff
@@ -899,12 +895,12 @@ mod tests {
...
+                    let syscall_number = machine.core.hart.xregisters.read(a7);
 
-                    // The `block-cache-tester` kernel will issue a system call with number -1
-                    // to indicate that it has reached the signal point. This is our cue that we
-                    // have produced the right "start" state.
-                    if syscall_number == -1 {
+                    // The `block-cache-tester` kernel will issue an IOCTL system call to indicate
+                    // that it has reached the signal point. This is our cue that we have produced
+                    // the right "start" state.
+                    if syscall_number == IOCTL {
```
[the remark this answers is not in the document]
kurtisc: -1 is missing so it raises a `SIGSYS` signal which is `CORE`, so the test loops forever at zero.

I wrote this before adding `SIG_IGN` support. I thought `IOCTL` was a fitting system call. Setting `SIGSYS` to `SIG_IGN` may work but it will be more code: constructing a `libc::sigaction` is a pain. Up to you.

vapourismo:
The "Linux Tests" job seems to be timing out.

PR #310, opened 2025-09-05
Use kernel RISC-V sigaction struct fields

kurtisc:
Relates to RV-756

# What

Corrects the `LinuxSigAction` type to use analogues of the same fields as `sigaction struct` on RISC-V *in the Linux kernel*. The old code replicated the userspace libc type.

Uses a restorer function loaded at runtime, a la the vDSO, as the RISC-V `sigaction struct` *in the Linux kernel* lacks a field for it to be passed in.

Updates the regression tests to reflect the changes in data stored in the machine core state.

# Why
The existing code replicates the `sigaction struct` of the C library. In the kernel, it is different: there is no restorer field, and the mask field is placed last.

Instead of a C-library restorer function being passed in as a function pointer, Linux on RISC-V has a number of small functions called the [vDSO](https://man7.org/linux/man-pages/man7/vdso.7.html) ([some examples](https://github.com/torvalds/linux/tree/d69eb204c255c35abd9e8cb621484e8074c75eaa/arch/riscv/include/asm/vdso)), of which the restorer is one.

This change adds that (very small) function to the initialisation of a Linux process and emulates it in a test. 

# How

 - Reorder and redefine fields of `LinuxSigAction`
 - Change from an array of per-signal restorers storage to storage for a single restorer 
 - Write to that storage using `load_restorer` when setting up a Linux process
   - While this only uses the fields of `self.machine_state.core`, it depends on the private `linux::VirtAddr` and is only used for Linux code, so is situated in the impl for `Pvm`
 - Change the test to write a restorer function
# Manually Testing

Tested by [kurtis/fix-fields-test](https://github.com/tezos/riscv-pvm/commit/bfcaca54221043b38f9802bbc0d7f64ef92010ed) #324

I've decided to split the test into its own PR as:
 - It's complex
 - It frees up 3 PRs to be worked on in parallel.
 - It depends on 2 of those PRs (this and #312)

# Regressions
This changes what is stored in the `MachineCoreState` and what is written to memory when a Linux program is loaded, so the goldenfiles have been updated.

# Tasks for the Author


src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -50,15 +50,14 @@ pub enum SignalError {
...
-    sa_sigaction: VirtAddr,
-    sa_mask: u32,
-    sa_flags: u32,
-    sa_restorer: VirtAddr,
+    pub sa_sigaction: VirtAddr,
+    pub sa_flags: u32,
+    pub sa_mask: u64,
 }
 
 /// Currently we only support the `sa_sigaction` field, so in tests it's useful to be able to
```
felixp72, line 63: Is this still true? Or rather, what does "support" really means here?

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -80,68 +79,50 @@ pub const NO_HANDLER: VirtAddr = VirtAddr::new(0u64);
...
-            dest.add(offset).cast::<u32>().write(self.sa_mask.to_le());
-            let offset = offset + size_of_val(&self.sa_mask);
-
             dest.add(offset).cast::<u32>().write(self.sa_flags.to_le());
             let offset = offset + size_of_val(&self.sa_flags);
 
-            dest.add(offset)
-                .cast::<u64>()
-                .write(self.sa_restorer.to_machine_address().to_le());
+            let offset = offset + SIGRETURN_PADDING;
```
felixp72, line 124: Is this used to align to `size_of_val(&self.sa_mask)`? Having a function for alignment can be more self-documenting and avoids magic numbers. Maybe adding an `unchecked_align_up` correspondent to the `VirtAddr::align_up` working directly on `u64` ?


```suggestion
            let offset = align_by_increasing(offset, &self.sa_mask);
```

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -311,6 +311,56 @@ where
         Ok(())
     }
 
+    fn load_restorer(&mut self, address: VirtAddr) -> Result<VirtAddr, MachineError>
```
felixp72, line 314: Although not required since its not `pub`, but doc-comment saying what is the purpose of the restorer function might help curious minds in the future.

felixp72 (changes requested):

src/riscv/lib/src/pvm/linux/signals.rs
kurtisc: Fixed in the PR just merged [link](https://github.com/tezos/riscv-pvm/commit/8b3bbcfef70c0fbebe1e1ec53a126874609ed1f0#diff-4293ba17c6d4b17be9dd006e0ca50f51760cc8ee0bc78d11aaaaef4f81bf74baR64)

src/riscv/lib/src/pvm/linux/signals.rs
kurtisc: >Is this used to align to size_of_val(&self.sa_mask)? 

No, it's because the kernel struct has unused padding between the flags and the mask that would be used by the restorer if the ABI had it as part of the struct. So the padding relates to the restorer, not to the mask.

Would that change things? The value is constant, so it would just be a const function with a similar name returning a constant.

src/riscv/lib/src/pvm/linux.rs
kurtisc: Agreed, went into some detail [here](https://github.com/tezos/riscv-pvm/compare/210e56fd0017ddc8919d635259365879abcc285f..e1b8021610d44c1770bbc992a6e9ce5b5856d01c#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7R314) as it isn't something intuitive or widely known.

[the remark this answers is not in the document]
kurtisc: I agree on the associations, changed [here](https://github.com/tezos/riscv-pvm/compare/210e56fd0017ddc8919d635259365879abcc285f..e1b8021610d44c1770bbc992a6e9ce5b5856d01c#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7R319) on L319 of linux.rs. Another munged diff :(

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -311,6 +311,56 @@ where
...
+            .core
+            .main_memory
+            .write_all(address.to_machine_address(), &RESTORER_FUNCTION)?;
+
+        // Remove access to the restorer function that has just been placed into memory
+        self.machine_state.core.main_memory.protect_pages(
+            address.to_machine_address(),
+            RESTORER_LENGTH,
+            Permissions::NONE,
+        )?;
```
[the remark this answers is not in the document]
kurtisc: I don't think I know what you're asking here, but I only set this page to `NONE` because the same is done in `load_program` (which you've noticed I was inspired by).

I've currently [deleted it ](https://github.com/tezos/riscv-pvm/compare/210e56fd0017ddc8919d635259365879abcc285f..e1b8021610d44c1770bbc992a6e9ce5b5856d01c#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7L347) on L347 of linux.rs. If the problem is that it is unnecessary, I think that resolves it.

If the problem is that you are worried that this is *insufficient*, would you be able to expland on how corruption could be solved?

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -166,50 +148,38 @@ impl Elem for LinuxSigAction {
...
     flags: [Cell<u32, M>; SignalIndex::COUNT],
-    /// An array of [VirtAddr]s, restorers for each signal, see
+    /// An array of bitmasks, one mask for each supported signal
+    masks: [Cell<u64, M>; SignalIndex::COUNT],
+    /// A pointer to a restorer address to jump to after returning from a signal handler. A
+    /// function to call `rt_sigreturn` should be loaded into here.
+    ///
     /// <https://www.man7.org/linux/man-pages/man2/sigreturn.2.html>
-    restorers: [Cell<VirtAddr, M>; SignalIndex::COUNT],
+    pub restorer: Cell<VirtAddr, M>,
```
vapourismo, line 161: If you move the `write_restorer` function into this module, then you close the interface leakage. The field doesn't need to be public.

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -94,40 +97,27 @@ pub const SA_SIGINFO: u32 = 0x4000000;
...
 
             Self {
-                sa_handler: VirtAddr::new(u64::from_le(sa_handler_bits)),
-                sa_sigaction: VirtAddr::new(u64::from_le(sa_sigaction_bits)),
-                sa_mask: u32::from_le(sa_mask_bits),
                 sa_flags: u32::from_le(sa_flags_bits),
-                sa_restorer: VirtAddr::new(u64::from_le(sa_restorer_bits)),
+                sa_sigaction: VirtAddr::new(u64::from_le(sa_sigaction_bits)),
+                sa_mask: u64::from_le(sa_mask_bits),
             }
```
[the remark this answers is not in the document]
kurtisc: It can be, [here](https://github.com/tezos/riscv-pvm/compare/026730a3521b2355bb2be29a61462199875823c4..5fb5f86b063589449c483628bf20ef171046e0de#diff-4293ba17c6d4b17be9dd006e0ca50f51760cc8ee0bc78d11aaaaef4f81bf74baR120) and I like it. The biggest problem I see is that the `Elem` methods are infallible while the zerocopy methods are falllible

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -52,15 +67,17 @@ pub enum SignalError {
...
-    sa_sigaction: VirtAddr,
-    sa_mask: u32,
-    sa_flags: u32,
-    sa_restorer: VirtAddr,
+    pub sa_sigaction: VirtAddr,
+    pub sa_flags: u32,
+    /// The kernel struct has padding between the flags and mask that would be used by the restorer if
+    /// RISC-V's ABI didn't use the vDSO for a restorer.
+    __sigreturn_padding: [u8; 12],
+    pub sa_mask: u64,
```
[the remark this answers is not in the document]
kurtisc: [Swapped](https://github.com/tezos/riscv-pvm/compare/5fb5f86b063589449c483628bf20ef171046e0de..6c6974b4a8f316236d38802d2c19dffe166fa0b9#diff-4293ba17c6d4b17be9dd006e0ca50f51760cc8ee0bc78d11aaaaef4f81bf74baR80) to `U32<LittleEndian>` and `U64<LittleEndian>`

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -311,6 +311,69 @@ where
...
+        const A7_ENC: u32 = 0b10001;
+        const OP_ADDI: u32 = 0b001_0011;
+        const LOAD_SIGRETURN: u32 =
+            (IMM << 20) | (X0_ENC << 15) | (F3_0 << 12) | (A7_ENC << 7) | OP_ADDI;
+
+        // Encoding of ECALL instruction
+        const ECALL: u32 = 0b1110011;
+
+        const RESTORER_FUNCTION: [u32; 2] = [LOAD_SIGRETURN, ECALL];
+        const RESTORER_LENGTH: usize = size_of_val(&RESTORER_FUNCTION);
```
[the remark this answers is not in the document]
kurtisc: I see, this was easier than I thought, I've updated the test to use the real thing [here](https://github.com/tezos/riscv-pvm/compare/5fb5f86b063589449c483628bf20ef171046e0de..6c6974b4a8f316236d38802d2c19dffe166fa0b9#diff-4293ba17c6d4b17be9dd006e0ca50f51760cc8ee0bc78d11aaaaef4f81bf74baR831)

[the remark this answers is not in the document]
vapourismo: Is it possible to test the restorer in isolation? Would be good to test the restorer as a unit, not in integration with the remainder of the signal functionality. No need to amend the test, just add a separate one that tests "jumping to the restorer does the intended thing".

vapourismo (approved):

PR #314, opened 2025-09-05
Initial sampling profiler for the PVM

victor-dumitrescu:
Closes TZX-22.

# What

This PR introduces:
- A sampling profiler for the PVM which can produce a [folded stacks](https://profilerpedia.markhansen.co.nz/formats/folded-stacks/) profile. Initially, this does not contain the full stack of every sample, only the running function (follow-up RV-763).
- A command for the Etherlink kernel which runs the profiler.

# Why

To understand the performance characteristics of kernels running on the RISC-V PVM in order to inform its development.

# How

The profiler is implemented as a subcommand of `run` in the sandbox. It takes a kernel compiled in debug mode, reads its debug symbols, and builds a map associating symbols to their address. It then runs a PVM stepper with the kernel in a separate thread, samples its `pc` at regular intervals, and resolves it to a symbol. 

The output profile simply contains a list of symbols and the number of samples which resolved to it. I chose the folded stacks profile because it is easy to output, can easily be turned into a flamegraph, and will also be suitable when outputting full call stack traces.

# Manually Testing

```
make -C kernels/etherlink profile-kernel
```

The profile can be read using tools such as flamegraph.com (example [here](https://flamegraph.com/share/f42ca2ed-a389-11f0-94ec-760777e76ccd)) or locally using https://github.com/brendangregg/FlameGraph.

The Firefox profiler cannot show folded stacks profiles and I could not find a convenient way to convert it to one of the formats it supports.

# Tasks for the Author

### DIFF
--- a/kernels/etherlink/Makefile
+++ b/kernels/etherlink/Makefile
@@ -12,6 +12,13 @@ ifneq ($(FEATURES),)
 FEATURES_OPT := --features "$(FEATURES)"
 endif
 
+# Build the kernel in release mode, unless a profile is passed
+ifneq ($(PROFILE),)
+BUILD_OPT := --profile $(PROFILE)
+else
+BUILD_OPT := --release
+endif
+
 # TODO RV-769: Remove when xtask alias works in this directory
 XTASK := cargo run $(NATIVE_OPT) -p xtask
 
@@ -30,7 +37,7 @@ build-kernel: cargo-build-kernel
 
 build-kernel-native: cargo-build-kernel-native
 
-test: bench-kernel bench-kernel-native trace-kernel
+test: bench-kernel bench-kernel-native trace-kernel profile-kernel
 
 check: cargo-format cargo-check cargo-clippy cargo-check-docs
 
@@ -55,10 +62,10 @@ cargo-check-docs:
 	@INBOX_FILE=$(INBOX_FILE) cargo doc --all-features --document-private-items --no-deps
 
 cargo-build-kernel:
-	@INBOX_FILE=$(INBOX_FILE) cargo build --package etherlink --release $(FEATURES_OPT)
+	@INBOX_FILE=$(INBOX_FILE) cargo build --package etherlink $(BUILD_OPT) $(FEATURES_OPT)
 
 cargo-build-kernel-native:
-	@INBOX_FILE=$(INBOX_FILE) cargo build --package etherlink --release $(FEATURES_OPT) $(NATIVE_OPT)
+	@INBOX_FILE=$(INBOX_FILE) cargo build --package etherlink $(BUILD_OPT) $(FEATURES_OPT) $(NATIVE_OPT)
 
 cargo-build-inbox-bench:
 	@cargo build $(NATIVE_OPT) --bin inbox-bench --release
@@ -72,6 +79,9 @@ bench-kernel-native: build-kernel-native
 
 trace-kernel:
 	@$(XTASK) bench -t
+
+profile-kernel:
+	@$(XTASK) profile -s 500
 
 # Mark all targets as phony to make sure they're always executed
 .PHONY: all $(MAKECMDGOALS)
--- a/kernels/etherlink/xtask/src/main.rs
+++ b/kernels/etherlink/xtask/src/main.rs
@@ -23,10 +23,11 @@ const DEFAULT_ROLLUP_ADDRESS: &str = "sr163Lv22CdE8QagCwf48PWDTquk6isQwv57";
 const DEFAULT_INBOX: &str = "assets/etherlink-erc20-inbox.json";
 
 #[derive(Debug, Default)]
-struct BenchConfig {
+struct BuildConfig {
     static_inbox: bool,
     native: bool,
     tracing: bool,
+    profiling: bool,
     data_dir: Option<PathBuf>,
 }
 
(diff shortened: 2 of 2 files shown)


src/riscv/sandbox/src/commands/profile.rs
```diff
@@ -0,0 +1,103 @@
...
+            Some(max) => {
+                let remaining = max.saturating_sub(total_steps);
+                if remaining == 0 {
+                    break;
+                }
+                std::cmp::min(sample_interval, remaining)
+            }
+        };
+
+        let result = stepper.step_max(Bound::Included(steps));
```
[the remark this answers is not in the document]
vapourismo: I think we need time-based sampling for accurate measures. In smaller scopes, the step-based sample could be interesting. It is an intriguing insight for now.

src/riscv/sandbox/src/commands/profile.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use crate::cli::ProfileOptions;
+use crate::commands::run::BlockImpl;
+use crate::commands::run::make_pvm_stepper;
+use crate::memory_config::MemoryConfigValue;
+
+/// Run profiling with the specified memory configuration
+pub fn profile_with_memory_config<MC: memory::MemoryConfig>(
+    opts: ProfileOptions,
+) -> Result<(), Box<dyn Error>> {
```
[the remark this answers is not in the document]
victor-dumitrescu: It's just a way of saying it could be any value that implements `Error`, but the exact error type isn't known at runtime so neither is its size, but `Result` expects it to be. `Box` is a pointer and has a known size, which fulfils this requirement. Have a look at https://doc.rust-lang.org/1.8.0/book/trait-objects.html#dynamic-dispatch.

victor-dumitrescu:
> Why did you go for a new command over integrating this into `run`? It feels like sampling is an extension of the `run` command.

In the latest version I've made it an extension of `run`.

tools/sandbox/src/commands/run/sample.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Record a sample from the stepper in its current state
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that:
+    /// - `pc_ptr` points to a valid u64 value
+    /// - The pointed-to value remains valid for the lifetime of this call
+    pub unsafe fn sample_from_ptr(&mut self, pc_ptr: *const u64) {
+        // SAFETY: Caller guarantees pc_ptr is valid and concurrent reads are safe
+        let pc = unsafe { pc_ptr.read() };
```
emturner, line 43: possibly you may want `read_volatile` here?

tools/sandbox/src/commands/run/profile.rs
```diff
@@ -0,0 +1,167 @@
...
+fn run_stepper_concurrent<S>(
+    mut stepper: S,
+    sampler: &mut Sampler,
+    sample_interval: Duration,
+    max_steps: Bound<usize>,
+) -> Result<usize, String>
+where
+    S: Stepper<Manager = Owned> + Send + 'static,
+{
+    let keep_running = Arc::new(Mutex::new(true));
```
emturner, line 105: why use a mutex and not just an `AtomicBool`? you don't actually need locking here IIUC

tools/sandbox/src/commands/run/profile.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// Run the profiling stepper loop with time-based concurrent sampling
+fn run_stepper_concurrent<S>(
+    mut stepper: S,
+    sampler: &mut Sampler,
+    sample_interval: Duration,
+    max_steps: Bound<usize>,
+) -> Result<usize, String>
+where
+    S: Stepper<Manager = Owned> + Send + 'static,
```
emturner, line 103: is the `Send` needed? IIUC, you could equally spin the sampler into the background thread.

victor-dumitrescu: Right! This also simplifies the running logic, since the same stepper can be run either normally or for profiling. Done [here](https://github.com/tezos/riscv-pvm/compare/86572e492d725416e7f1235e79593c3904f1897e..2101a9f127e2c4495f18487f32608908eb1b36b2).

tools/sandbox/src/commands/run/profile.rs
```diff
@@ -0,0 +1,115 @@
...
+    max_steps: Option<usize>,
+    output_path: &Path,
+) -> Result<(), Box<dyn Error>> {
+    let symbols = Symbols::new(program)?;
+
+    let is_running = Arc::new(AtomicBool::new(true));
+    let is_running_sampler = Arc::clone(&is_running);
+
+    // Channel to send pc pointer to sampler thread
+    let (pc_tx, pc_rx) = mpsc::channel::<usize>();
```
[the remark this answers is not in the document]
victor-dumitrescu: Indeed, now that the background thread does the sampling this is not longer needed. Done [here](https://github.com/tezos/riscv-pvm/compare/432c7d2965e573453c92d8e8c21f36de0e869aea..982c5ef15ffad551210f2e50dbd913d666f563f8.

victor-dumitrescu:
> I'm wondering if you made a `profile` struct that wrapped the background thread and the `samples` struct - you could move all this up into the `run` function directly.
> 
> (e.g. you'd end up with an `option<sampling>`) and then map it at the end to finish?

I've played around with this idea and I'm not convinced I see the benefit. The upside would be that a single `run` function could handle both regular running and profiling instead of 2 different functions, but the downsides are that the `run` function's logic becomes more complex because it has to check whether it's profiling to know when to launch the sampler thread, run the stepper, then check again to know whether to signal the sampler thread to stop. We also need to create this new type which wraps the closure for the thread and a pointer to the `is_running` bool only to pass it from the function which creates them to the `run` function.

PR #316, opened 2025-09-08
feat(page_cache): add PageCache top-level trait

emturner:
- Part of RV-764

# What

Add a trait that any `PageCache` will implement.

# Why

This allows us to potentially:
- more easily tie in the `PAGES` constant from main memory using its config
- allow in future, a `NoOp` implementor for all non-Owned backends. This would be a proof optimisation, essentially (as all proofs will be generated/verified with an empty page cache). The no-op would remove the page-population step from the proof.

# Manually Testing

```
make all
```

# Tasks for the Author


src/riscv/lib/src/machine_state/page_cache.rs
```diff
@@ -0,0 +1,404 @@
...
+
+/// Mask for indexing into a page's memory.
+const PAGE_OFFSET_MASK: u64 = (1 << memory::OFFSET_BITS) - 1;
+
+/// Instance of the page cache.
+#[expect(
+    unused,
+    reason = "PageCache is not integrated into the rest of the codebase"
+)]
+pub trait PageCache<MC: MemoryConfig, M: ManagerBase> {
```
vapourismo, line 52: It's a bit hard to evaluate this trait when there is no usage/implementation of it in the PR.

src/riscv/lib/src/machine_state/page_cache.rs
```diff
@@ -0,0 +1,404 @@
...
+
+/// Per page, we store exactly the number of instruction halfwords we could fetch from that page's
+/// memory.
+///
+/// This is exactly half the page size - we know the instruction program counter is always
+/// halfword-aligned.
+const INSTRUCTION_ENTRIES: usize = 1
+    << memory::OFFSET_BITS
+        .checked_sub(1)
+        .expect("OFFSET_BITS is non-zero") as usize;
```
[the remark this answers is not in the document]
vapourismo: Perhaps the `OFFSET_BITS` could be made `NonZero`.

src/riscv/lib/src/machine_state/page_cache.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// or no longer executable.
+    ///
+    /// While `pages` are expected to be `PAGE_SIZE` aligned, this is not a hard requirement.
+    /// Rather, all pages partially or fully overlapping with the range will be invalidated.
+    fn invalidate_range(&mut self, pages: std::ops::Range<u64>);
+}
+
+/// A page containing entrypoints that may then be dispatched against the [`MachineCoreState`].
+pub(crate) struct PageDispatch<'a, MC: MemoryConfig, M: ManagerBase> {
+    page: &'a [Instruction; INSTRUCTION_ENTRIES],
```
vapourismo, line 76: Nit: I find the mention of "dispatch" a little misleading given the JIT and block cache have the "dispatch target" concept. It's not the most meaningful change but might improve cognitive load if this was a "CodePage" or "InstructionPage".

src/riscv/lib/src/machine_state/page_cache.rs
```diff
@@ -0,0 +1,404 @@
...
+    _pd: PhantomData<(MC, M)>,
```
vapourismo, line 77: The `Instruction` itself isn't specific to `MemoryConfig` or `Manager`. How come you need capture those type variables in this type?

vapourismo (changes requested):

src/riscv/lib/src/machine_state/page_cache.rs
emturner: that is fair. 

In terms of usage - that's a bit hard right now, as e.g. blocked by #276. In terms of implementation, I had split out interpreted, as it lead to a significantly larger amount of code trying to land (especially in terms of required tests), but I'll shift that back into here with a second commit. Will take a little while, however

src/riscv/lib/src/machine_state/page_cache.rs
emturner: I like `CodePage`

emturner: renamed [here](https://github.com/tezos/riscv-pvm/compare/ccd68f93a498d66ea32a271cd29966b0a664c5bb..89f19cf77865754501854d4d3e0cdf94ac4a083c)

emturner:
@vapourismo @NSant215 this PR now includes the initial `PageCacheImpl`

vapourismo (changes requested):
I added a few comments, but LGTM overall.

PR #323, opened 2025-09-09
fix(etherlink/bench): find repo-root independent of parent folder name

emturner:
# What

Fix etherlink/test when repository has been cloned elsewhere that `riscv-pvm`.

# Why

On my machine, I use `riscv/pvm` as the path, as other riscv-related work is contained in the `riscv` folder.

# How 

Look for `.git` or `.jj` instead.

# Manually Testing

```
cd ..
mv riscv-pvm pvm-riscv
cd pvm-riscv
make etherlink/test
```

# Tasks for the Author

### DIFF
--- a/kernels/etherlink/xtask/src/main.rs
+++ b/kernels/etherlink/xtask/src/main.rs
@@ -124,7 +124,7 @@ fn find_repo_root() -> Result<PathBuf> {
 
     let mut dir = current_dir.as_path();
     loop {
-        if dir.ends_with("riscv-pvm") && dir.join(".github").exists() {
+        if dir.join(".github").exists() {
             return Ok(dir.to_path_buf());
         }
 


kernels/etherlink/xtask/src/main.rs
```diff
@@ -124,7 +124,7 @@ fn find_repo_root() -> Result<PathBuf> {
 
     let mut dir = current_dir.as_path();
     loop {
-        if dir.ends_with("riscv-pvm") && dir.join(".github").exists() {
+        if dir.join(".git").exists() || dir.join(".jj").exists() {
```
victor-dumitrescu, line 127:
```suggestion
        if dir.join(".github").exists() {
```
If the issue is with the path name, would dropping that not be enough? I think `.github` would be present on everyone's machine since it's checked in and there would be no reason for it to be present anywhere else in the repo.

emturner: good point, don [here](https://github.com/tezos/riscv-pvm/compare/25e796ad004e6371268c69dcb1947e36eb4f605c..c5fbdfc38ad8008248ac8216959f1de8531cdbce)

victor-dumitrescu (approved):

PR #327, opened 2025-09-10
Self PC update by "tezos" system calls

edwin1729:
Closes RV-777
# What
Remove the illusion that PC moves forward by an instruction on every system call. Currently the pc is incremented by `instr_width=4` of an ECALL at the start of the top-level handler, and then some of the system calls choose to further modify the PC (eg, `handle_srt_sigreturn` sets it to a fixed address).

This PR has every system call handler effectively returning a `ProgramCounterUpdate` which has been added as a field to `SystemCallResultExecution`. In practice most syscall handlers actually return a `Result<u64,..>` only which is converted using `Into` to the default `SystemCallResultExecution` which represent a `Next` update. It is helpful to know which syscalls don't return `Result<u6,..>`, but directly return `SystemCallResultExecution`, while reviewing. These are:
`exit`, `tkill` `rt_sigreturn`

# Why
PC is possibly not incremented when running an ECall corresponding to a parallel syscall.

# How
Special care needs to be taken
Current
# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/machine_state.rs
+++ b/src/riscv/lib/src/machine_state.rs
@@ -456,7 +456,7 @@ impl<MC: memory::MemoryConfig, BCC: BlockCacheConfig, B: Block<MC, M>, M: backen
     pub fn step_max_handle<E>(
         &mut self,
         mut step_bounds: Bound<usize>,
-        mut handle: impl FnMut(&mut Self, Bound<usize>) -> ControlFlow<E>,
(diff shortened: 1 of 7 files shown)


vapourismo (changes requested):

edwin1729:
The diff of the updated version of this pr and this pr are mutually exclusive. I have checked that none of the reviews apply to the updated pr as well. So I close this PR. Please see #346 for the updated version

PR #329, opened 2025-09-10
Move riscv-sandbox folder to repository root

vapourismo:
# What

# Why

# How

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author


emturner (commented):
overall LGTM 

I actually quite like having analysise-jit-functions in tools?

In my mind, a good split would be tooling we expect to be used externally by other teams (ie sandbox) to be at the top level.

Otherwise tooling we use internally could exist under `tools`?

Though I apologise for not actually saying this when you raised the discussion, other than with cryptic emojis.

vapourismo:
It's perfectly fine to raise it on the PR instead of Slack 👍

vapourismo:
> Any thoughts on moving `src/riscv` -> `riscv` ?

I think we could even move it to the root. We just need to be a bit careful given the library is worked on a lot, so we want to avoid conflicts.

vapourismo:
> > I am not sure on the distinction between the analysis tool and the sandbox. Say Jstz, Etherlink, or Tezlink starting using the analysis tool for profiling, do we then move that tool also to the root of the repository?
> 
> Perhaps a better way of phrasing that is that our main executable is the sandbox, and we have tooling (in this case the analyser) that may optionally be used to consume/act on 'some output' of the sandbox? ie that of a notion of 'primary' and 'secondary' tooling?
> 
> Though I acknowledge this is not a completely firm distinction.
> 
> > I believe we must consider only criteria that we control when deciding on a directory structure. Whether something is externally used, is out of our control and is therefore a flaky criterion. I acknowledge this is not exactly the same as your point, which is more about the intention to be used externally. So if we were to combine the two, we end up with categories: a) guaranteed externally used (library), b) likely externally used (sandbox), c) unlikely externally used.
> 
> This does indeed make sense.
> 
> > Because the applicability of these categories can change over time, I would rather go for something more static: a) essential (`/`), b) utility executable (`/tools`), c) utility kernel (`/kernels`). Alternatively, utility executable could also be at the root as well given they share the "ecosystem" like Rust version, libraries and so on.
> 
> Perhaps we could use the notion of stability here instead? ie the default behaviour of the sandbox is something we do keep (roughly) stable, whereas there are no stability guarantees for e.g. analyser (in terms of output, or cli arguments etc). likewise for the inbox-bench executables, which are not at the top-level, we allow to modify at length the behaviour/output etc
> 
> But otherwise, if anything - I think I'd rather move the sandbox to `tools` in the end after all 😅 - but admittedly this is personal preference to avoid a super-cluttered top level of the repisitory

I think stability is a good one too. However, I wouldn't consider the sandbox executable stable. We have been liberally adding and modifying things without concern for breakage beyond our uses. I think the stability might have been accidental and not intentional.

If we look to the future, when we need to be very deliberate with changes to the core library, I would want to avoid including a utility executable like the sandbox into our changelog considerations. Obviously, the folder locations do not directly impact a changelog. I am just linking back to stability.

Should we just move all tools to `/tools` and declare that these are not interface stable?

felixp72:
> I think we could even move it to the root. We just need to be a bit careful given the library is worked on a lot, so we want to avoid conflicts.

I feel like the root of the repository has more responsabilities than being a cargo workspace for the core pvm library. Looking at scripts, the separate kernels workspace, nix, docs/ folder, the tools workspace.

emturner (approved):

vapourismo:
> The PR title, description, and commit message are out of date since the change is now moving the sandbox to `tools/` but otherwise LGTM.

Those should be up-to-date now.

PR #335, opened 2025-09-12
Self PC update by linux system call handlers

edwin1729:
Part of RV-777
Project Data Parallelism in System Calls: [Design Doc](https://linear.app/tezos/document/design-doc-0211bbf47d5d), [excalidraw](https://app.excalidraw.com/s/8ILYpjPL16O/8jXiLZz9ST2?next=%2Fs%2F8ILYpjPL16O%2F8jXiLZz9ST2)
# What
Remove the illusion that PC moves forward by an instruction on every system call. Currently the pc is incremented by `instr_width=4` on an ECALL at the start of the top-level handler, and then some of the system calls choose to further modify the PC (eg, `handle_srt_sigreturn` sets it to a fixed address). Use `ProgramCounterUpdate` to make the PC update logic more clear and explicit. 

# Why
For code clarity, especially since more system calls with custom PC update logic are going to be implemented soon.

# How
This PR has every system call handler effectively returning a `ProgramCounterUpdate` which has been added as a field to `SystemCallResultExecution`. In practice most syscall handlers actually return a `Result<u64,..>` only which is converted using `Into` to the default `SystemCallResultExecution` which represent a `Next` update. It is helpful to know which syscalls don't return `Result<u6,..>`, but directly return `SystemCallResultExecution`, while reviewing. These are:
`exit`, `tkill` `rt_sigreturn`

The PC increments for host functions have been moved to `handle_tezos` and they happen without use of `ProgramCounterUpdate` at the moment. The next PR will involve a similar refactor to this PR for host functions.

There is a minor semantic change I've made [here](https://github.com/tezos/riscv-pvm/pull/335/files#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7R533). It only concerns logging. Previously the pc was reported after being incremented by 4. I made it the pc before incrementing, as pc+4 is not always actually the next instruction executed. This is because of `rt_sigreturn`, the only syscall which sets the pc to some value.

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/pvm/linux.rs
+++ b/src/riscv/lib/src/pvm/linux.rs
@@ -16,6 +16,7 @@ use std::ffi::CStr;
 use std::ops::ControlFlow;
 use std::ops::Range;
 
+use parameters::SystemCallResultExecution;
 use perfect_derive::perfect_derive;
 use tezos_smart_rollup_constants::riscv::SBI_FIRMWARE_TEZOS;
 
@@ -26,6 +27,7 @@ use super::Pvm;
 use crate::machine_state::MachineCoreState;
 use crate::machine_state::MachineError;
 use crate::machine_state::MachineState;
+use crate::machine_state::ProgramCounterUpdate;
 use crate::machine_state::block_cache::BlockCacheConfig;
 use crate::machine_state::block_cache::block::Block;
 use crate::machine_state::memory::Address;
@@ -36,6 +38,7 @@ use crate::machine_state::memory::Permissions;
 use crate::machine_state::registers;
 use crate::program::Program;
 use crate::pvm::hooks::PvmHooks;
+use crate::pvm::linux::parameters::ECALL_WIDTH;
 use crate::pvm::linux::signals::Signal;
 use crate::state::NewState;
 use crate::state_backend::AllocatedOf;
@@ -527,10 +530,7 @@ impl<M: ManagerBase> SupervisorState<M> {
         B: Block<MC, M>,
         M: ManagerReadWrite,
     {
-        // We need to jump to the next instruction. The ECall instruction which triggered this
-        // function is 4 byte wide.
-        let pc = machine.core.hart.pc.read().saturating_add(4);
-        machine.core.hart.pc.write(pc);
+        let pc = machine.core.hart.pc.read();
 
         /// Read an argument from a register and interpret it as a system call argument.
         /// If that fails, log the failure.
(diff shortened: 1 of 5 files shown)


src/riscv/lib/src/pvm/linux.rs
```diff
@@ -869,10 +880,9 @@ impl<M: ManagerBase> SupervisorState<M> {
         self.exited = true;
         self.exit_code = signal.exit_code();
 
-        // Return 0 as an indicator of success, even if this might not actually be used
         Ok(parameters::SystemCallResultExecution {
```
vapourismo, line 883: Lets keep the `result` field, it is quite useful to see here.

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -359,6 +359,10 @@ pub(super) fn handle_tezos<MC, M>(
     MC: MemoryConfig,
     M: ManagerReadWrite,
 {
+    // TODO: RV-777: remove below and instead have each system call return a `ProgramCounterUpdate`
+    let pc = machine.hart.pc.read().saturating_add(4);
```
vapourismo, line 363: There is a bug in the original code. It should be wrapping addition. Would you mind changing that?

src/riscv/lib/src/pvm/linux/signals.rs
```diff
@@ -779,13 +781,14 @@ impl<M: ManagerBase> SupervisorState<M> {
...
+    ) -> Result<SystemCallResultExecution, Error>
     where
         M: ManagerReadWrite,
     {
         let pc = core.pop_signal_context().map_err(|_| Error::Fault)?;
-        core.hart.pc.write(pc);
-        // Return 0 as an indicator of success
-        Ok(0)
+        Ok(SystemCallResultExecution {
+            pc_update: ProgramCounterUpdate::Set(pc),
```
vapourismo, line 790: Do you mind keeping the `result: 0` in here to make it clear what this system call returns in terms of the return value?

vapourismo (approved):

src/riscv/lib/src/pvm/linux/signals.rs
edwin1729: Addressed [here](https://github.com/tezos/riscv-pvm/compare/9e539e2ee53f955aa07c6f53d5de01c2c34c28f2..9b2e1c07d0d82f3b9c2601af0ac8836288383cf9)

PR #346, opened 2025-09-17
Self PC update by "tezos" system calls

edwin1729:
Closes RV-777
# What
Remove the restriction that PC moves forward by an instruction on every tezos host function. 

# Why
Currently all host functions increment the PC to got the next one. However, soon, parallel host functions will run the same `ECALL` instr multiple times so they don't increment the PC. They only increment the PC when they hit the base case.

# How
Introduce a `TezosCallResult` struct which represents all the information returned on handling a system call. Implement the `From` `Result<XValue, SbiError>` to preserve usage of `?` in the sequential system call handlers.

Move `ECALL_WIDTH` to the parent module to avoid making `linux::parameters` module public

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author


src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -359,19 +405,22 @@ pub(super) fn handle_tezos<MC, M>(
     MC: MemoryConfig,
     M: ManagerReadWrite,
 {
-    // TODO: RV-777: remove below and instead have each system call return a `ProgramCounterUpdate`
-    let pc = machine.hart.pc.read().wrapping_add(4);
-    machine.hart.pc.write(pc);
-
     let sbi_function = machine.hart.xregisters.read(a6);
-    match sbi_function {
+    let TezosCallResult { result, pc_update } = match sbi_function {
```
victor-dumitrescu, line 409: Introducing `TezosCallResult` and restructuring the code around it seems quite convoluted to me. I think the core the of the change here is that you now want to know how to update the PC, while not changing anything else. What do you think of something like

```rust
let pc_update = match sbi_function {
```

which you then update at the end like you currently do, but without bringing the implementation of `sbi_wrap` along. You'd need to change `sbi_wrap` and the functions it wraps and the functions dispatched on the match to return the PC update instead.

edwin1729: I can let handling of `result: u64` stay within `sbi_wrap`. But I will need to introduce `TezosCallResult` regardless, as the [next PR](https://github.com/tezos/riscv-pvm/pull/347/files#diff-585d1ff984e81a6862d79c9b7e36b2465446fab4ec06db74d4716ed0f3be63deR55-R70) requires the handlers to return both `pc_update` and 'steps_completed`

As explained in the comments in that PR, including `result: u64` makes it follow the same pattern as `linux::parameters::SystemCallResultExecution`

edwin1729: Okay I tried to do what you've suggested but that introduces two kinds of result types 1) looks like my original type including `result: u64`. This one is the return of each handler, so it would be received by `sbi_wrap` when it calls its `inner` function. 2) The return of `sbi_wrap`: this would be same as (1) but without the `result:u64` field.

So I think it is cleaner to do all the result handling in one place, and use a single result type as currently in this PR.

src/riscv/lib/src/pvm/linux/parameters.rs
```diff
@@ -29,9 +29,6 @@ pub struct SystemCallResultExecution {
     pub pc_update: ProgramCounterUpdate<Address>,
 }
 
-/// An ECALL is 4 bytes wide, ie, an uncompressed instr
```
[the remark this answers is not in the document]
edwin1729: Done [here](https://github.com/tezos/riscv-pvm/pull/346/files#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7L41-R41)

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -38,7 +38,7 @@ use crate::machine_state::memory::Permissions;
 use crate::machine_state::registers;
 use crate::program::Program;
 use crate::pvm::hooks::PvmHooks;
-use crate::pvm::linux::parameters::ECALL_WIDTH;
+pub(crate) use crate::pvm::linux::parameters::ECALL_WIDTH;
```
vapourismo, line 41: Let's make the `parameter` module `pub(crate)` instead, so we don't have to rely on re-exports too much.

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -43,10 +45,53 @@ use crate::machine_state::registers::a1;
...
 use crate::state_backend::ManagerReadWrite;
 use crate::state_backend::ManagerWrite;
 
+/// The result of handling a host function.
+/// This is similar to `linux::parameters SystemCallResultExecution` but there
+/// are slight differences between what we allow tezos and linux calls to do.
+struct TezosCallResult {
+    /// The value written to output register `a0`.
+    /// Functions such as `tezos_inbox_next` don't write register hence the `Option`.
+    result: Option<Result<XValue, SbiError>>,
```
vapourismo, line 59: This field type screams to be turned into a proper enum. It's documentation alone only makes sense if you understand the connection to the SBI calling conventions. With a dedicated enum you can express its function way clearer:

```rust
/// Status of a Tezos host function call
enum CallStatus {
    /// The call is suspended and will be resumed later.
    Suspended,

    /// The host function handler has finished successfully.
    Success(XValue),

    /// The host function handler errored.
    Error(SbiError),
}
```

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -43,10 +45,53 @@ use crate::machine_state::registers::a1;
...
+        // In cases where the execution should halt, this should be specified.
+        Self {
+            result: Some(Err(value)),
+            pc_update: ProgramCounterUpdate::Next(ECALL_WIDTH),
+        }
+    }
+}
+
+impl TezosCallResult {
+    const NO_RESULT: Self = TezosCallResult {
```
vapourismo, line 89: Technically Tezos host functions always have results. It's just that those are populated later. This seems important information to convey.

```suggestion
    const SUSPENDED: Self = TezosCallResult {
```

vapourismo (changes requested):

src/riscv/lib/src/pvm/tezos.rs
victor-dumitrescu: I don't think you need the `TezosCallResult` type at all if you handle the result using `sbi_wrap` like before. You only need to return the `pc_update` and in your next PR you'll need to return `(pc_update, steps_complete)`.

In your current version, handling the result is split between `sbi_wrap` for certain calls and `handle_tezos` for others. You can keep result handling in `sbi_wrap` only, like in the previous version, and have all of these functions dispatched in `handle_tezos` return `ProgramCounterUpdate<Address>` instead of `TezosCallResult`. This would also eliminate the need to shoehorn a pc update as a result in functions which don't really need to return results (`handle_tezos_inbox_next`, `handle_not_supported`).

src/riscv/lib/src/pvm/linux.rs
edwin1729: Done [here](https://github.com/tezos/riscv-pvm/compare/7f472b71784d25a9d740d1795602a9146fff81ce..24f27f2951e1136e68d65cf7f9f67543f2143baf)

src/riscv/lib/src/pvm/tezos.rs
victor-dumitrescu: Or another idea which might be even simpler and wouldn't require changing any of the existing handlers, is to redefine `sbi_wrap` as a function which wraps calls which only advance the pc by 4 (so every call to `sbi_wrap` will always return `ProgramCounterUpdate::Next(ECALL_WIDTH)`). Handlers which can result in different pc updates would not use `sbi_wrap` but handle the result logic themselves.

src/riscv/lib/src/pvm/tezos.rs
edwin1729: Done [here](https://github.com/tezos/riscv-pvm/compare/bbd664db30806cc1bf6bf9bfc34ce9013257d30b..37368f2b5f08780e9da54d69ae157e07661d6ec5)

src/riscv/lib/src/pvm/tezos.rs
edwin1729: This has been addressed

src/riscv/lib/src/pvm/tezos.rs
edwin1729: I think the reason for dissatisfaction with my PR is that `sbi_wrap` is still used and hasn't been deleted, and the function that sbi_wrap performs is "duplicated" in a match block in handle_tezos [here](https://github.com/tezos/riscv-pvm/pull/346/files#diff-585d1ff984e81a6862d79c9b7e36b2465446fab4ec06db74d4716ed0f3be63deR431-R435). But I argue I am not really duplicating any code here, since I'm using the helpers `sbi_return1` and `sbi_return_error`.

`sbi_wrap` is still used as it was in the cases where the `status: CallStatus::Suspended` is being "resumed", like `provide_reveal_response`. These functions are called directly and not through `handle_tezos`.  Here we directly write the `a0` register using `sbi_wrap`. It may be that these "resumptions" may also want to perhaps return a `pc_update` or `steps_completed` explicitly. But I think this is outside the scope of this PR.

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -43,10 +45,61 @@ use crate::machine_state::registers::a1;
...
+    /// The error is handled by the caller (kernel)
+    /// We move to the next instr (increment the pc)
+    fn error(err: SbiError) -> Self {
+        Self {
+            status: CallStatus::Error(err),
+            pc_update: ProgramCounterUpdate::Next(ECALL_WIDTH),
+        }
+    }
+
+    /// Usually the PC is incremented after handling an ECALL
```
victor-dumitrescu, line 91: This comment should say what the function does rather than when it would be used. What do you think about "Advance the pc to the next instruction"

src/riscv/lib/src/pvm/tezos.rs
```diff
@@ -359,19 +413,26 @@ pub(super) fn handle_tezos<MC, M>(
...
-        _ => handle_not_supported(&mut machine.hart.xregisters),
+        _ => handle_not_supported(),
+    };
+
+    machine.update_pc(machine.hart.pc.read(), pc_update);
+    match status {
+        CallStatus::Success(reg_val) => sbi_return1(&mut machine.hart.xregisters, reg_val),
+        CallStatus::Error(error) => sbi_return_error(&mut machine.hart.xregisters, error),
+        CallStatus::Suspended => {}
     }
```
felixp72, line 437: This part is more or less the same in what it does to the general `handle_system_call` function in `linux.rs`. Do you think you can unify these behaviours? What I see as the source for this maybe unnecessary duplication is the fact that for tezos functions we return early in the match statement [here](https://github.com/tezos/riscv-pvm/blob/59fdecc5982111e7051e8400ab18ea43db2499e6/src/riscv/lib/src/pvm/linux.rs#L782)

PR #367, opened 2025-09-26
fix: targets for updating assets

victor-dumitrescu:
# What

Fix the mechanism for updating checked-in kernels after the `assets` directory was moved in #356.

# How

A new Makefile in `assets` where the targets from `src/riscv/Makefile` are moved to. 

# Manually Testing

```
make -C assets
```

# Tasks for the Author


emturner (commented):
might be worth updating the assets at the same time? or at least the checksums.

Currently the checksums fail the check due to the bad path

emturner:
(on a separate note - would it be worth having a CI sanity check that ensures various checksums are up to date? (cc @vapourismo))

assets/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+.PHONY: update-assets
+update-assets: riscv-dummy.elf jstz
+
+riscv-dummy.elf riscv-dummy.elf.checksum::
+	@make -C .. dummy/build
+	@cp ../kernels/dummy/target/riscv64gc-unknown-linux-musl/release/riscv-dummy $@
+	@sha256sum $@ > $@.checksum
+
+jstz jstz.checksum::
```
vapourismo, line 14: Neither of these rules require double colon.

```suggestion
riscv-dummy.elf riscv-dummy.elf.checksum:
	@make -C .. dummy/build
	@cp ../kernels/dummy/target/riscv64gc-unknown-linux-musl/release/riscv-dummy $@
	@sha256sum $@ > $@.checksum

jstz jstz.checksum:
```

vapourismo (approved):

emturner (approved):

victor-dumitrescu:
> (on a separate note - would it be worth having a CI sanity check that ensures various checksums are up to date? (cc @vapourismo))

Good point, I'll open a separate PR for that.

assets/Makefile
victor-dumitrescu: Updated [here](https://github.com/tezos/riscv-pvm/compare/0a7eb8c252f229af3241d0c1d949d138c939c198..51104f9e6d85a3ac3c55475e3de4979fa4321de2). Not sure why, but the double colon made them also behave like `.PHONY` rules, I had to mark them explicitly when I removed it.

vapourismo (approved):

PR #384, opened 2025-09-30
refactor(memory): operate over lengths of `NonZeroU64`

emturner:
- closes RV-561
- related to RV-760

# What

adjust `lengths` in memory to be `NonZeroU64`.

# Why

reads/writes/permission changes of `length == 0` have no effect. This is now made explicit.

Additionally, this makes clearer, for future, that only permission changes that could actually have an effect need be listened to by the JIT Router/Page Cache. Essentially, we'll adjust `protections.rs` to work over page index ranges (with the exception of `can_access_narrow`). This is far easier/more clear if we can avoid any empty ranges.

We adjust length from usize -> u64 to be consistent w.r.t. lengths in the Pvm. e.g. we wish to avoid potential issues when running on 32-bit hardware, should someone choose to.

# Manually Testing

```
make all
```

# Regressions

None

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/machine_state/memory/state.rs
+++ b/src/riscv/lib/src/machine_state/memory/state.rs
@@ -75,9 +75,7 @@ impl<const PAGES: usize, const TOTAL_BYTES: usize, B, M: ManagerBase>
         E: Elem,
         M: ManagerWrite,
     {
-        let length = E::STORED_SIZE
-            .try_into()
-            .expect("Elem size `E::STORED_SIZE` must fit into u64");
+        let length = E::STORED_SIZE;
 
         Self::check_bounds(address, length, BadMemoryAccess)?;
 
@@ -100,9 +98,7 @@ where
         E: Elem,
         M: ManagerRead,
     {
-        let length = E::STORED_SIZE
-            .try_into()
-            .expect("Elem size `E::STORED_SIZE` must fit into u64");
+        let length = E::STORED_SIZE;
 
         Self::check_bounds(address, length, BadMemoryAccess)?;
 
@@ -122,9 +118,7 @@ where
         E: Elem,
         M: ManagerRead,
     {
-        let length = E::STORED_SIZE
-            .try_into()
-            .expect("Elem size `E::STORED_SIZE` must fit into u64");
+        let length = E::STORED_SIZE;
 
         Self::check_bounds(address, length, BadMemoryAccess)?;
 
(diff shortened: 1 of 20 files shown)


src/riscv/lib/src/pvm/linux/memory.rs
```diff
@@ -82,8 +82,13 @@ impl<M: ManagerBase> SupervisorState<M> {
         MC: MemoryConfig,
         M: ManagerReadWrite,
     {
+        let Some(length) = NonZeroU64::new(length) else {
+            // nothing to do
```
vapourismo, line 86:
```suggestion
            // Length 0 means no protections need to be changed.
```

src/riscv/lib/src/pvm/linux.rs
```diff
@@ -271,7 +272,12 @@ where
             .map(|(addr, data)| addr.saturating_add(data.len() as u64))
             .max()
             .unwrap_or(0);
-        let program_length = program_end.saturating_sub(program_start) as usize;
+
+        let Some(program_length) = NonZeroU64::new(program_end.saturating_sub(program_start))
+        else {
+            // there is no program to load
+            return Ok(());
```
vapourismo, line 279: There is a bit of state that doesn't get set if you exit here. You could either move that up or use conditionals on the blocks that require `program_length`.

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -70,7 +75,10 @@ impl<const PAGES: usize, const TOTAL_BYTES: usize, B, M: ManagerBase>
         E: Elem,
         M: ManagerWrite,
     {
-        let length = E::STORED_SIZE.get();
+        let length = E::STORED_SIZE
+            .try_into()
+            .expect("Elem size `E::STORED_SIZE` must fit into u64");
```
vapourismo, line 80: Conversions like these happen quite a lot. Perhaps it is best to make `STORED_SIZE` a `u64` as well?

vapourismo (changes requested):

src/riscv/lib/src/pvm/linux.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/7c0394853287802ad5c045c8c35c6d52b1867f1e..cb79f02099b4bf62a908003bf3535c78d0e2d8e4)

emturner:
further on the benchmarking - we now have `STORED_SIZE` and `STORED_SIZE_USIZE` which eliminates the performance regression

emturner:
(for reviewers - treat this as draft again for now)

PR #385, opened 2025-09-30
`Get`, `Set`, `Delete` operations for the Persistence Layer

felixp72:
Closes RV-796

# What

Add `Get`, `Set`, `Delete` operations for the Persistence layer.

Update: With the revised design in mind, the changes in this PR correspond to the interface which will be used by the Merkle layer.

# Why

The persistence needs to be able to handle basic CRUD-like operations for saving data on disk.

# How

The persistence layer works with content-addressable blobs of data. Meaning the type `HashedData` associates a `blake3` hash to a byte-array and will use that hash as the key to identify it in rocksdb.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/persistence_layer.rs
+++ b/durable-storage/src/persistence_layer.rs
@@ -34,9 +34,32 @@ use crate::repo::DirectoryManagerError;
 /// Type alias for a 32-byte hash used for identifying key-value blobs & commits.
 type Hash = [u8; 32];
 
+/// Represents a key-value blob to be stored in the persistence layer.
+///
+/// Invariant: [`HashedData`] is content-addressable by the stored hash.
+#[derive(Debug, Clone)]
+pub struct HashedData<'d> {
+    /// The BLAKE3 hash of the value.
+    hash: Hash,
+
+    /// The addressable content of the blob.
+    value: &'d [u8],
+}
+
+impl<'d> HashedData<'d> {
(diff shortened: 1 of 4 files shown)


NSant215 (approved):
If it is possible, it might also be good to check the number of entries in the database at the point of deletion in the tests, otherwise looks good.

durable-storage/src/persistence_layer.rs
```diff
@@ -176,4 +229,67 @@ mod tests {
                 .exists()
         );
     }
+
+    #[test]
+    fn test_basic_ops() {
+        let test = |value_a: String, value_b: String| {
+            let repo = DirectoryManager::new(std::path::Path::new("/tmp/test_basic_ops"))
```
victor-dumitrescu: Nit: Why not use a `tempdir`-generated path here?

felixp72: This should make it easier to debug if the test actually fails

emturner: it's in general not _super_ pleasant to ignore e.g. `TMPDIR` environment variable though? 

Why would it be harder to debug if using `tempdir`? (because the path would be cleaned automatically?)

If so, you _could_ have a wrapper over `tempdir` that only drops the path if we're [not panicking](https://doc.rust-lang.org/std/thread/fn.panicking.html) and otherwise if we are, print the path?

PR #393, opened 2025-10-02
feat(durable_storage): checkout & commit operations for persistence layer

felixp72:
Closes RV-795

# What

The `PersistenceLayer` is now capable of saving itself to a specified `CommitId` and instantiate itself from disk given a `CommitId` and a reference to a `Repo`

# Why

This mechanism is used by the rollup node to save & restore the PVM & the associated durable storage.

# How

The file structure of the `DirectoryManager` now becomes:

```
<repo_path>:
    temporary/
        db_<random>/checkpoint/
            rocksdb internals
    commits/
        <commit_id>/
            rocksdb internals
```

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/persistence_layer.rs
+++ b/durable-storage/src/persistence_layer.rs
@@ -17,14 +17,18 @@
 //! - checkout a specific commit
 //!
 //! The folder structure of the [`DirectoryManager`] is:
-//! ```
+//! ```text
 //! <repo_path>:
 //!    temporary/
 //!        db_<random>/checkpoint/
 //!            <rocksdb internals>
+//!    commits/
+//!       <commit_id>/
+//!          <rocksdb internals>
 //! ```
 
 use std::mem::ManuallyDrop;
+use std::path::Path;
 
 use rocksdb::checkpoint::Checkpoint;
 use tempfile::TempDir;
@@ -34,6 +38,24 @@ use crate::repo::DirectoryManagerError;
 
 /// Type alias for a 32-byte hash used for identifying key-value blobs & commits.
 type Hash = [u8; 32];
+
+/// [`CommitId`]'s are used to generate [`PersistenceLayer`] commits & to checkout specific commits
+/// from a [`DirectoryManager`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CommitId(Hash);
+
+impl From<blake3::Hash> for CommitId {
+    fn from(hash: blake3::Hash) -> Self {
+        Self(hash.into())
+    }
+}
+
+impl CommitId {
+    /// Returns the hex encoded commit id.
+    pub fn hex_encode(&self) -> String {
+        hex::encode(self.0)
+    }
+}
 
 /// Represents a key-value blob to be stored in the persistence layer.
 ///
@@ -58,6 +80,9 @@ impl<'d> HashedData<'d> {
 /// Errors encountered when interacting with the persistence layer.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    #[error("Commit not found")]
+    CommitNotFound,
+
     #[error("Key not found")]
     KeyNotFound,
 
@@ -69,7 +94,7 @@ pub enum Error {
 }
 
 /// Mode in which the [`PersistenceLayer`] was instantiated.
-enum Mode {
+enum Mode<'a> {
     /// Either a new database, or a clone of an existing database.
     Temporary {
         /// We keep it here to keep it alive for the lifetime of the rockdb instance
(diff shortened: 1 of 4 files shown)


durable-storage/src/repo.rs
```diff
@@ -57,4 +76,9 @@ impl DirectoryManager {
 
         Ok(tempdir)
     }
+
+    /// Obtain the path to the commit directory for the given `id`.
+    pub fn commit_dir(&self, id: &CommitId) -> PathBuf {
+        self.commits_dir.join(id.to_bs58())
```
vapourismo: What is the reason for picking base 58? Is it better than hex encoding, for example?

durable-storage/src/persistence_layer.rs
```diff
@@ -110,10 +136,17 @@ pub struct PersistenceLayer {
...
 
     /// What mode was the [`PersistenceLayer`] opened in.
-    mode: Mode,
+    mode: Mode<'a>,
 }
 
-impl PersistenceLayer {
+impl<'a> PersistenceLayer<'a> {
+    /// Creates a checkpoint of the current database at the given `path`.
+    fn checkpoint_at(&self, path: &PathBuf) -> Result<(), Error> {
```
vapourismo: `&PathBuf` requires a `PathBuf`, whereas `&Path` works for all sorts of paths. Even better would be `impl AsRef<Path>`, like `create_checkpoint` requires.

```suggestion
    fn checkpoint_at(&self, path: &Path) -> Result<(), Error> {
```

durable-storage/src/persistence_layer.rs
```diff
@@ -149,23 +182,68 @@ impl PersistenceLayer {
...
+    /// Checks out a specific commit in the repository from the given `repo`
+    pub fn checkout(repo: &'a DirectoryManager, id: &CommitId) -> Result<Self, Error> {
+        let db_path = repo.commit_dir(id);
+
+        // We assume the commit is not found if the folder does not exist.
+        if !Path::exists(&db_path) {
+            return Err(Error::CommitNotFound);
+        };
+
+        let db = rocksdb::DB::open(&rocksdb::Options::default(), &db_path)?;
```
vapourismo: Should you use the `rocksdb_options()` here instead of the default ones?


```suggestion
        let db = rocksdb::DB::open(&rocksdb_options(), &db_path)?;
```

durable-storage/src/persistence_layer.rs
```diff
@@ -149,23 +182,68 @@ impl PersistenceLayer {
...
 
+    /// Whenever a mutable operation is performed, we need to ensure that the persistence layer is
+    /// in [`Mode::Temporary`] mode.
+    fn make_temporary(&mut self) -> Result<(), Error> {
+        match &self.mode {
+            Mode::Temporary { .. } => (),
+            Mode::FromCommit { repo } => {
+                *self = self.try_clone(repo)?;
+            }
+        };
```
vapourismo: I would have thought Clippy yells at you for not making this an `if-let` statement. 

Nit but a tiny bit more readable:

```suggestion
        if let Mode::FromCommit { repo } = &self.mode {
            *self = self.try_clone(repo)?;
        };
```

vapourismo (changes requested):

durable-storage/src/persistence_layer.rs
```diff
@@ -35,6 +36,24 @@ use crate::repo::DirectoryManagerError;
 /// Type alias for a 32-byte hash used for identifying key-value blobs & commits.
 type Hash = [u8; 32];
 
+/// [`CommitId`]'s are used to generate [`PersistenceLayer`] commits & to checkout specific commits
+/// from a [`DirectoryManager`].
+#[derive(Debug, Clone, PartialEq)]
+pub struct CommitId(Hash);
+
+impl From<&Hash> for CommitId {
```
emturner: nit/feel free to resolve 

`Hash` implements `Copy` - so there's no need to take a reference? Especially if the hash is only created once, you can just pass it

durable-storage/src/persistence_layer.rs
```diff
@@ -35,6 +36,24 @@ use crate::repo::DirectoryManagerError;
...
+    fn from(hash: &Hash) -> Self {
+        Self(*hash)
+    }
+}
+
+impl CommitId {
+    /// Returns the hex encoded commit id.
+    pub fn hex_encode(&self) -> String {
```
emturner: could also consider `b58check` encoding? but perhaps for later :)

Cargo.toml
```diff
@@ -31,6 +31,7 @@ bincode = "2.0"
 blake3 = "1.8.2"
 capstone = "0.13"
 cfg-if = "1.0.4"
+bs58 = "0.5.1"
```
emturner: it looks like you're not using this anywhere?

PR #395, opened 2025-10-02
feat(proofs): use whole byte to represent tags for nodes and leafs

vapourismo:
Part of RV-798

# What

Currently, we treated the first bytes of a proof serialisation as a bit stream, where each 2 bits represent a tag (node, leaf, or blinded) that can be used to validate and navigate through the shape of the proof's Merkle tree.

This PR changes this so that each tag takes 1 byte. With that, the prefix becomes a byte stream.

# Why

This is a requisite for #398.

As part of RV-798, we are moving to a proof encoding where shape information (tags) and data are interleaved. The interleaving is necessary to enable monadic parsing. See RV-798 for more details.

Interleaving tags and data makes compression via bit stream less efficient and increases the complexity of the parser implementation. We accept the slightly decreased efficiency for a much simpler implementation.

# Manually Testing

```
make all
```

# Regressions

The change in proof serialisation format impacts `jstz/proof_initial` as it captures a serialised proof.

# Tasks for the Author


vapourismo:
> I think this needs to be rebased on to main again now that rc-refcell has gone in

Done.

src/riscv/lib/src/state_backend/proof_backend/proof/deserialise_stream.rs
```diff
@@ -352,20 +327,17 @@ impl<'t, R> DeserialiserNode<R> for StreamBranchComb<'t, R> {
 impl<R> StreamParserComb<'_, R> {
     /// Deserialise the input and return the result if all bytes have been consumed.
     pub fn into_result(self) -> Result<R> {
-        let mut input = self.deser.tags.remaining_to_stream_input();
-        let result = (self.f)(&mut input);
+        let mut input = self.deser.input;
+        let result = (self.f)(&mut input)?;
 
         // A correct deserialisation makes sure all bytes are consumed.
         // Otherwise, if tags dictate more a bigger part of the proof has been blinded then
```
NSant215, line 334: I don't understand what this comment is saying

src/riscv/lib/src/state_backend/proof_backend/proof/deserialiser.rs
NSant215, line 345: Is this comment still valid? IIUC they are each their own byte now?

src/riscv/lib/src/state_backend/proof_backend/proof.rs
```diff
@@ -448,13 +501,15 @@ mod tests {
...
+        check_serialisation(root, &[TAG_NODE, TAG_READ, 12, 15, 30, 40]);
 
         let root = MerkleProof::Node(vec![n1.clone(), n2.clone()]);
         check_serialisation(
             root,
             &[
-                [(TAG_NODE << 6) | (TAG_READ << 4) | (TAG_BLIND << 2)].as_ref(),
+                [TAG_NODE].as_ref(),
+                [TAG_READ].as_ref(),
+                [TAG_BLIND].as_ref(),
```
emturner, line 512: nit, all of these can be 
```suggestion
                &[TAG_BLIND],
```

src/riscv/lib/src/state_backend/proof_backend/proof/deserialise_stream.rs
emturner: I think it's saying that if we have data left over at the end, then the tags are not correct as it implies more values/data were blinded (according to the tags) than were included in the proof

emturner (approved):

src/riscv/lib/src/state_backend/proof_backend/proof/deserialise_stream.rs
NSant215: That makes sense - I think the comment could be made clearer but otherwise happy with it

NSant215 (approved):

PR #396, opened 2025-10-02
feat(jit): Refactor step-counting mechanism in JIT graph

NSant215:
Relates to RV-703
Closes RV-791.

## Note to Reviewers
I think the produced diff in the review tool is quite confusing. I would recommend starting with the new version of `control_flow_graph.rs` and just reading through the new version of the function `find_step_counter_updates`.

# What

Refactor the step-counting mechanism in the JIT sequence to a transition-based method. 

# Why

As detailed in RV-791, the current method is not optimal and results in a higher number of step-variable flushes than is necessary. 

# How

We change the analysis to a transition-based approach, explained in detail in RV-791.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/src/riscv/lib/src/jit/builder/control_flow_graph.rs
+++ b/src/riscv/lib/src/jit/builder/control_flow_graph.rs
@@ -14,8 +14,8 @@
 // once later.
 #![cfg(test)]
 
-use crate::jit::builder::graph_walker::GraphWalker;
-use crate::jit::builder::instr_map::InstrId;
+use std::collections::VecDeque;
+
 use crate::jit::builder::instr_map::InstrMap;
 use crate::jit::builder::instr_map::InstrMapBuilder;
 use crate::jit::builder::outcome_map::Graph;
@@ -85,6 +85,11 @@ pub struct StepCounterUpdate<'info, T> {
     /// Unsuccessful outcomes increase by just `base_diff`.
     base_diff: usize,
 
+    /// If the program exits after this outcome, the step counter must be further increased by this
+    /// amount (in addition to `base_diff`-related increments),
+    /// as subsequent step-counter updates will not fire.
+    exit_delta: usize,
+
     /// Edge where the step counter needs updating
     edge: &'info DirectedEdgeInfo<T>,
 }
@@ -92,7 +97,8 @@ pub struct StepCounterUpdate<'info, T> {
 impl<'info, T> StepCounterUpdate<'info, T> {
     /// Number of steps to increment the step counter on a successful outcome.
     pub fn success_delta(&self) -> usize {
-        self.base_diff + 1
+        // wrapping_add is required as base_diff can be negative.
+        self.base_diff.wrapping_add(1)
     }
 
     /// Number of steps to increment the step counter on an exception outcome.
@@ -107,6 +113,7 @@ impl<'info, T> StepCounterUpdate<'info, T> {
 }
 
 /// Control flow graph of instructions
+#[derive(Debug)]
 pub struct ControlFlowGraph<'info, T> {
     /// Nodes in the control flow graph, indexed by instruction ID
     nodes: InstrMap<&'info NodeInfo<T>>,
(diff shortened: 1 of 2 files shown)


src/riscv/lib/src/jit/builder/control_flow_graph/tests.rs
```diff
@@ -55,6 +55,10 @@ struct TestInstrOutcome {
 
     /// Amount to increment the step counter by if this outcome is taken
     step_delta: Cell<usize>,
+
+    /// Extra amount to increment the step counter if the program
+    /// exits after executing this outcome.
+    exit_delta: Cell<usize>,
```
emturner, line 61: this PR turns a situation that was previously not-optimal into a more optimal solution IIUC? it would be great to have a test that demonstrates this

src/riscv/lib/src/jit/builder/control_flow_graph/tests.rs
```diff
@@ -55,6 +55,10 @@ struct TestInstrOutcome {
```
emturner, line 55: nit: this is more of a `feat` type PR no? as it would actually make a difference to the performance of step counting

src/riscv/lib/src/jit/builder/control_flow_graph/tests.rs
NSant215: good point - I have added a test to the mid-sequence recursion that ensures there isn't a step update from the first to second instruction. Let me know what you think

[diff](https://github.com/tezos/riscv-pvm/compare/5b67a32738a7209f38b453853852946432212183..7902ec511a3c795f6ce05cf87fb7ea237edb4a49)

src/riscv/lib/src/jit/builder/control_flow_graph.rs
```diff
@@ -85,14 +86,17 @@ pub struct StepCounterUpdate<'info, T> {
     /// Unsuccessful outcomes increase by just `base_diff`.
     base_diff: usize,
 
+    /// If the program exits after this outcome, the step counter must be increased by this amount.
```
vapourismo, line 89: Add why this needs to be done. I.e. because the subsequent step counter updates won't fire.

src/riscv/lib/src/jit/builder/control_flow_graph.rs
```diff
@@ -179,122 +184,91 @@ where
...
+                .graph
+                .outgoing_outcomes(SourceInstrLoc::Internal(start_idx))
+            {
+                queue.push_back(*outcome);
             }
         }
 
-        step_deltas.map(|_idx, delta| delta.expect("All nodes should have a step delta by now"))
+        queue
     }
```
vapourismo, line 210: The existence of this function seems odd. The algorithm we discussed doesn't call for an initial traversal. The rules applied for each transition should cover this case without the need for duplicate traversal.

src/riscv/lib/src/jit/builder/control_flow_graph.rs
```diff
@@ -179,122 +184,91 @@ where
...
             }
-        })
-    }
-}
-
-/// Classification of edges from a node
-enum EdgeClass {
-    /// Has exactly one explicit edge and no implicit edges
-    Unambiguous(InstrId),
+        }
```
vapourismo, line 268: This algorithm seems more involved than what we discussed and what's written in the proposed solution. What is the reasoning behind that?

src/riscv/lib/src/jit/builder/control_flow_graph.rs
```diff
@@ -179,122 +184,91 @@ where
...
+                    }
+                    None => {
+                        step_deltas[instr_id] = Some(source_delta + 1);
+
+                        for outcome in self
+                            .graph
+                            .outgoing_outcomes(SourceInstrLoc::Internal(instr_id))
+                        {
+                            outcome_queue.push_back(*outcome);
+                        }
```
vapourismo, line 256: Use `outcome_queue.extend(...)`.

NSant215: but for a breadth-first search we want to add to the back of the queue right?

src/riscv/lib/src/jit/builder/control_flow_graph.rs
NSant215: I think the problem was that I was trying to handle impossible edge cases a lot. I have done a large refactor that I think has simplified it, let me know what you think [diff](https://github.com/tezos/riscv-pvm/compare/fc19b73..6daa25c)

src/riscv/lib/src/jit/builder/control_flow_graph.rs
```diff
@@ -179,122 +184,62 @@ where
...
-
-        step_deltas.map(|_idx, delta| delta.expect("All nodes should have a step delta by now"))
-    }
-
     /// Find all the edges where the step counter needs updating.
     pub fn find_step_counter_updates(&self) -> OutcomeMap<Option<StepCounterUpdate<'_, T>>> {
-        let step_deltas = self.find_step_deltas();
+        // An instruction delta is the number of steps that have been performed since the last step counter
+        // update. 0 means the step counter was updated through an incoming edge of a node.
+        let mut instr_deltas: InstrMap<Option<usize>> = self.nodes.map(|_, _| None);
```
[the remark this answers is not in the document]
NSant215: thanks [diff](https://github.com/tezos/riscv-pvm/compare/a4652ff1e2496536a9dfd01c57efb8bbd9688dda..fe30dbad2a55eb5453e095229a3a2d3bd265052f)

src/riscv/lib/src/jit/builder/control_flow_graph.rs
```diff
@@ -179,122 +184,65 @@ where
...
                 TargetInstrLoc::Exit => {
-                    // Exit transitions need to ensure the step counter is updated to reflect the
-                    // number of executed instructions in the current context.
-                    Some(StepCounterUpdate {
-                        base_diff: step_deltas[source],
-                        edge,
-                    })
+                    let sc_update = StepCounterUpdate {
+                        base_diff: source_delta.expect("Exit must have a source delta."),
+                        exit_delta: 0,
```
emturner: is this correct? what if there's an exception that occurs while executing this instruction/transition?

src/riscv/lib/src/jit/builder/control_flow_graph.rs
```diff
@@ -85,14 +85,18 @@ pub struct StepCounterUpdate<'info, T> {
     /// Unsuccessful outcomes increase by just `base_diff`.
     base_diff: usize,
 
+    /// If the program exits after this outcome, the step counter must be increased by this
+    /// amount, as subsequent step-counter updates will not fire.
+    exit_delta: usize,
```
emturner: is this 'instead of `base_diff`, or `in addition to`?

src/riscv/lib/src/jit/builder/control_flow_graph.rs
```diff
@@ -179,122 +184,65 @@ where
...
+                        edge: edge.expect("Exit edge must have edge info."),
+                    };
+                    *step_updates[outcome_id].data_mut() = Some(sc_update);
                 }
+                TargetInstrLoc::Internal(dest_id) => match instr_deltas[dest_id] {
+                    Some(existing_delta) => {
+                        // an unset source delta means the source is an entry point, implying
+                        // a delta of 0.
+                        let source_delta = source_delta.unwrap_or_default();
+                        let base_diff = source_delta.wrapping_sub(existing_delta);
```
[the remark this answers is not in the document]
NSant215: done [diff](https://github.com/tezos/riscv-pvm/compare/6e89b518264c0bafb97043c391960a735ef7f132..434b032)

src/riscv/lib/src/jit/builder/control_flow_graph.rs
NSant215: oh sorry didn't see the second comment for some reason. Exit delta is indeed only for `normal` exits. The timing of `exit_delta` would only be relevant after `success_delta` has already been applied, so it is in a different flow to when `exception_delta` would be used.

emturner: (feel free to resolve unless I've misunderstood!)

PR #401, opened 2025-10-06
feat(memory): allow listener to hook into memory permission updates

emturner:
- closes RV-760

# What

# Why

# How

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author


src/riscv/lib/src/machine_state/page_cache/state.rs
```diff
@@ -155,17 +158,28 @@ impl<const PAGES: usize, CPE: CodePageEntry<MC, M>, MC: MemoryConfig, M: Manager
...
+    CPE: CodePageEntry<MC, M>,
+    MC: MemoryConfig,
+    M: ManagerBase,
+{
+    fn handle_permissions_update(
+        &mut self,
+        pages: std::ops::RangeInclusive<u64>,
+        permissions: memory::Permissions,
+    ) {
+        let needs_invalidation = permissions.can_write() || !permissions.can_exec();
```
felixp72, line 179: The doc-comment for `handle_permissions_update` states the purpose of the function in a generic sense, but in the impl block like here it is not clear why I want the handle to perform invalidation. 

Don't know what's better, a comment for why invalidation is computed this way or a doc-string on top of this function, but one of these should be present here imo.

src/riscv/lib/src/machine_state.rs
```diff
@@ -328,10 +330,24 @@ impl<MC: memory::MemoryConfig, BCC: BlockCacheConfig, B: Block<MC, M>, M: backen
...
+    /// Get access to the main memory, with a listener to hook into any permission updates.
+    ///
+    /// This is required to keep parts of [`MachineState`], that do not form part of the PVM state,
+    /// synchronised with main memory so as to ensure determinism regardless of whether these
+    /// additional parts of the state are populated or not.
+    pub(crate) fn memory_with_listener(
+        &mut self,
+    ) -> (
+        &mut MC::State<M>,
+        impl MemoryGovernanceListener + use<MC, BCC, B, M>,
```
felixp72, line 346: The code compiled just fine by eliminating the `+ use<MC, BCC, B, M>`. I knew that this is usually used for lifetimes. What would be the purpose here? (I'm fairly new to the `use` clause)

src/riscv/lib/src/machine_state.rs
```diff
@@ -328,10 +330,24 @@ impl<MC: memory::MemoryConfig, BCC: BlockCacheConfig, B: Block<MC, M>, M: backen
...
+    ) {
+        (&mut self.core.main_memory, NoopMemoryGovernanceListener)
```
felixp72, line 348: If the returned listener is a no-op, why do you need a dedicated function for it? You can just instantiate a `NoopMemoryGovernanceListener` wherever needed.

Otherwise, at call-site I expect the listener to actually do something at least in _some_ cases. What do you say about having a constant exposed `DEFAULT_LISTENER/NOOP_LISTENER`? Avoids some noise destructuring the tuple at the same time & makes the behaviour more explicit at call site.

emturner: because when we subsequently introduce the page cache, this will be replaced by the page cache here.

doing it in this order allows to update everywhere that needs to handle this, without making the page cache MR even more noisy

src/riscv/lib/src/machine_state/page_cache/state.rs
emturner: added a doc-comment [here](https://github.com/tezos/riscv-pvm/compare/fdc27d492e92be2dd50d8bbf43761ae76ab443fc..7ea21465fedbb4c74e1b35f38380812a097345ee)

src/riscv/lib/src/machine_state.rs
emturner: removed the `use` constraint [here](https://github.com/tezos/riscv-pvm/compare/7ea21465fedbb4c74e1b35f38380812a097345ee..4f50efeffe7519f314c4cf9f68d70678ca8bd09a)

src/riscv/lib/src/machine_state.rs
felixp72: A, ok, for the future change it makes sense 👍

src/riscv/lib/src/machine_state/memory/listener.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use super::Permissions;
+
+/// A `MemoryGovernanceListener` is allowed to 'hook-into' any permission updates that may occur in
+/// memory.
+///
+/// This allows other parts of the state (e.g. the *PageCache* or future *JIT Router* to ensure
+/// they remain synchronised with memory).
+pub trait MemoryGovernanceListener {
+    /// Handle notification that a region of memory (set of pages) have had their permissions
```
victor-dumitrescu, line 18:
```suggestion
    /// Handle notification that a region of memory (set of pages) has had its permissions
```

src/riscv/lib/src/machine_state/memory/listener.rs
```diff
@@ -0,0 +1,42 @@
...
+    ) {
+    }
+}
+
+impl<MGL> MemoryGovernanceListener for &mut MGL
+where
+    MGL: MemoryGovernanceListener,
+{
+    fn handle_permissions_update(&mut self, pages: RangeInclusive<u64>, permissions: Permissions) {
+        MGL::handle_permissions_update(self, pages, permissions);
```
victor-dumitrescu, line 40:
```suggestion
        MGL::handle_permissions_update(*self, pages, permissions);
```

I think? Since `self` is `&mut &mut MGL` at this point?

src/riscv/lib/src/machine_state/memory/state.rs
```diff
@@ -359,12 +383,24 @@ where
     }
 }
 
+fn page_range(address: Address, length: NonZeroUsize) -> RangeInclusive<u64> {
+    let start_page = address_to_page_index(address) as u64;
+
+    let end_address = address
+        .wrapping_add(length.get() as Address)
+        .wrapping_sub(1);
```
victor-dumitrescu, line 391: Should these not be saturating? Particularly relevant for the subtraction.

emturner: no, it should be wrapping.

Particularly: 
- we know the length in non-zero. So the `wrapping_sub` will never underflow - unless the addition _already_ overflowed.
- take the case where we are writing 1 byte to `u64::MAX` (so the last byte of a page). The combination of wrapping add/wrapping sub ensures we remain 'within' the current page (we overflow to `0` and back to `u64::MAX`).
- if we write 2 bytes to `u64::MAX` however, we want to write to the next page too (the page starting at `0`). So we have to use the wrapping add. If we use saturating sub instead then the 1-byte case fails.

src/riscv/lib/src/machine_state/memory/listener.rs
emturner: done, see [diff](https://github.com/tezos/riscv-pvm/compare/4f50efeffe7519f314c4cf9f68d70678ca8bd09a..d78605143299120c709a769fc022c318ca665fdd)

src/riscv/lib/src/machine_state/memory/listener.rs
emturner: done, see [diff](https://github.com/tezos/riscv-pvm/compare/4f50efeffe7519f314c4cf9f68d70678ca8bd09a..d78605143299120c709a769fc022c318ca665fdd)

PR #407, opened 2025-10-07
Felix/kernels makefile

felixp72:
# What

# Why

# How

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author


Makefile
```diff
@@ -33,20 +31,11 @@ riscv/%:
...
-dummy/%: 
-	@make -C kernels/dummy ${@:dummy/%=%}
-
-block-cache-tester/%: 
-	@make -C kernels/block-cache-tester ${@:block-cache-tester/%=%}
-
-etherlink/%: 
-	@make -C kernels/etherlink ${@:etherlink/%=%}
+kernels/%: 
+	@make -C kernels ${@:jstz/%=%}
```
danielpapp-trilitech, line 35:
```suggestion
	@make -C kernels ${@:kernels/%=%}
```

Makefile
```diff
@@ -33,20 +31,11 @@ riscv/%:
...
 
 assets/%:
 	@make -C assets ${@:assets/%=%}
 
 # Mark all non-pattern targets as phony to make sure they're always executed
-.PHONY: all build-deps build-deps-slim check audit build test test-miri clean
```
danielpapp-trilitech, line 52: `test-miri` was removed here. Is this intentional?

danielpapp-trilitech: Ah, I see even the folder is removed.

victor-dumitrescu (dismissed):
I'm not sure I see this as an improvement. Particularly, I think it's valuable to have individual targets in the root Makefile in order to selectively build/test/check only the relevant target for a specific scenario.

Makefile
```diff
@@ -4,21 +4,19 @@
...
-build-deps: riscv/build-deps jstz/build-deps etherlink/build-deps
+build-deps: riscv/build-deps kernels/build-deps
 
 build-deps-slim: riscv/build-deps-slim
 
-check: riscv/check sandbox/check jstz/check dummy/check block-cache-tester/all etherlink/check assets/check
+check: riscv/check kernels/check assets/check
 
-build: sandbox/build jstz/build dummy/build block-cache-tester/build etherlink/build
+build: riscv/build kernels/build
```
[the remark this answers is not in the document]
felixp72: building the sandbox is part of building riscv. The riscv makefile under `src/riscv` is actually a cargo workspace level build (and tests, checks). So it will also perform the targets for the sanbox and the tools that are part of the root workspace `Cargo.toml`.

Ultimately this is what I would've liked to do, to have all the cargo workspaces correspond to a makefile that are in the same folder. Currently this is not the case, but at least we can cut the noise from the kernel targets.

Makefile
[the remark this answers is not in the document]
felixp72: Aha! So that's where it's gone. Reintroduced it [here](https://github.com/tezos/riscv-pvm/compare/ce18b91c8f8018d4434335e40d76b15930853807..80f88ec68ca18f044bae17cb9b21393210578c30). Had to bump the nightly version though.

felixp72: Turns out the miri-test target was lost a while ago. I reintroduced it now. [diff](https://github.com/tezos/riscv-pvm/compare/ce18b91c8f8018d4434335e40d76b15930853807..80f88ec68ca18f044bae17cb9b21393210578c30) is here but looking by commits might still be better

emturner:
> Remove the test-miri target - is not used / defined
This is no longer the case :)

could you link the PR description to [RV-49](https://linear.app/tezos/issue/RV-49/use-miri-to-verify-our-crates)?

.rustfmt.toml
```diff
@@ -2,3 +2,4 @@ style_edition = "2024"
 required_version = "1.8.0"
 group_imports = "StdExternalCrate"
 imports_granularity = "Item"
+overflow_delimited_expr = true
```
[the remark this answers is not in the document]
emturner: (@vapourismo am happy for this to be removed if you prefer though, either on this MR or subsequently! Don't feel too strongly either way as to style).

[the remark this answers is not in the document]
vapourismo: > false causes a number of formatting changes in the code base.

But isn't `false` the default?

felixp72:
I've removed the last commit which adds the dedicated makefile for the kernels. We'll discuss how we want our folder structure to look like. @victor-dumitrescu @vapourismo

NSant215 (approved):

Makefile
```diff
@@ -4,21 +4,21 @@
...
+all: riscv/all jstz/all dummy/all page-cache-tester/all etherlink/all
 
 build-deps: riscv/build-deps jstz/build-deps etherlink/build-deps
 
 build-deps-slim: riscv/build-deps-slim
 
-check: riscv/check sandbox/check jstz/check dummy/check page-cache-tester/all etherlink/check assets/check
+check: riscv/check jstz/check dummy/check page-cache-tester/check etherlink/check assets/check
 
-build: sandbox/build jstz/build dummy/build page-cache-tester/build etherlink/build
```
vapourismo: Why build the library but not the sandbox?

Makefile
```diff
@@ -4,21 +4,21 @@
...
+build: riscv/build jstz/build dummy/build page-cache-tester/build etherlink/build
```
vapourismo: There is nothing to build of value using `riscv/build`, why spend time running it?

PR #731, opened 2026-01-20
feat(data): add partial vector data structure

vapourismo:
Part of RV-861

# What

Introduces a partial vector data structure that lets you define arbitrary parts of a large space (`0..=usize::MAX`).

# Why

I found this data structure to be very useful in implementing the byte array state component (#732). It was also handy in making `DataSpace` more space efficient during proof generation (#736).

# How

I added lots of commentary within the data structure module. Please let me know what you think is missing.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/data/src/partial_vec.rs
+++ b/data/src/partial_vec.rs
@@ -1,0 +1,439 @@
+// SPDX-FileCopyrightText: 2026 TriliTech <contact@trili.tech>
+//
+// SPDX-License-Identifier: MIT
+
+//! Like [`Vec`] but not continuously defined
+//!
+//! # Structure
+//!
+//! A [`PartialVec`] is a binary tree where leaves are either defined data or undefined gaps.
+//! This allows efficient representation of sparse data without allocating memory for gaps.
+//!
+//! ```text
+//! Logical view:    [a][b][c][ ? ][ ? ][ ? ][ ? ][d][e][f]
+//!                   0  1  2   3    4    5    6   7  8  9
+//!                  |_______|___________________|_________|
+//!                   defined      undefined       defined
+//!
+//! Tree structure:
+//!
+//!                        Concatenated
+//!                         (len: 10)
+//!                        /          \
+//!                Concatenated       Defined
+//!                 (len: 7)          [d,e,f]
+//!                /        \
+//!           Defined     Undefined
+//!           [a,b,c]      (len: 4)
+//! ```
+//!
+//! Unlike a `BTreeMap<usize, T>`, this structure is optimized for range queries making it
+//! memory-efficient for sparse but clustered data.
+
+use std::ops::Range;
+
+/// Partial vector that may not be defined for certain ranges
+///
+/// This data type can be seen as an alternative to `BTreeMap<usize, T>` which provides a more
+/// memory-efficient representation for when there are many adjacent keys and the primary query
+/// mechanism is using ranges.
+#[derive(Debug, Clone)]
+pub enum PartialVec<T> {
+    /// Undefined vector of a certain width
+    Undefined {
+        /// How many elements are undefined
+        length: usize,
+    },
+
+    /// Defined vector of a certain width
+    Defined {
+        /// Elements that define the range
+        data: Vec<T>,
+    },
+
+    /// Concatenation of two partial vectors
+    Concatenated {
+        /// Combined width of both vectors
+        ///
+        /// This is not the number of defined items.
+        node_width: usize,
+
+        /// First partial vector
+        left: Box<Self>,
+
+        /// Second partial vector
+        right: Box<Self>,
+    },
+}
+
+impl<T> PartialVec<T> {
+    /// Create a vector of defined data.
+    pub fn defined(data: Vec<T>) -> Self {
+        Self::Defined { data }
+    }
+
+    /// Create a vector of undefined data with the provided length.
+    pub fn undefined(length: usize) -> Self {
+        Self::Undefined { length }
+    }
+
+    /// Retrieve the width of the partial vector.
+    ///
+    /// This is not the number of defined elements, but the total number of elements represented by
+    /// the partial vector - defined or undefined.
+    /// In other words, `PartialVec::undefined(10).width() == 10`.
+    fn width(&self) -> usize {
+        match self {
+            Self::Undefined { length } => *length,
+            Self::Defined { data } => data.len(),
+            Self::Concatenated { node_width, .. } => *node_width,
+        }
+    }
+
+    /// Define a range within the partial vector.
+    ///
+    /// Existing data is overwritten if it overlaps the newly defined range.
+    pub fn define(&mut self, offset: usize, new_data: Vec<T>) {
+        let mut work = vec![(self, offset, new_data)];
+
+        while let Some((target, offset, mut new_data)) = work.pop() {
+            // If there is nothing left to insert, we might as well stop.
+            if new_data.is_empty() {
+                break;
+            }
+
+            match target {
+                Self::Concatenated {
+                    left,
+                    right,
+                    node_width,
+                } => {
+                    let new_data_end = offset + new_data.len();
+
+                    // Inserting may extend the range represented by the node.
+                    *node_width = std::cmp::max(*node_width, new_data_end);
+
+                    // When the offset is larger than what the left node stores, then we can safely
+                    // delegate all the work to the right node.
+                    // This case also handles when we want to define something beyond the current
+                    // node's width. In that scenario, we descend into the right node.
+                    if offset >= left.width() {
+                        let new_offset = offset - left.width();
+                        work.push((right, new_offset, new_data));
+                        continue;
+                    }
+
+                    // When the inserted range fits entirely within the left node, we can focus on
+                    // the left node only.
+                    if new_data_end <= left.width() {
+                        work.push((left, offset, new_data));
+                        continue;
+                    }
+
+                    // At this point we know that the inserted range overlaps with both the left and
+                    // right nodes. So we divide the insertion vector accordingly.
+                    let left_overlap_tail = left.width() - offset;
+                    let new_right_data = new_data.split_off(left_overlap_tail);
+
+                    work.push((left, offset, new_data));
+                    work.push((right, 0, new_right_data));
+                }
+
+                Self::Undefined { length } => {
+                    let new_data_end = offset + new_data.len();
+
+                    let undefined_prefix_len = offset;
+                    let undefined_suffix_len = length.saturating_sub(new_data_end);
+
+                    // This is the "center" node. We prepend and append undefined nodes as needed.
+                    // We don't need to worry about merging adjacent nodes, as this node is either
+                    // the root or the right-most node in a concatenation. The `Concatenated` case
+                    // deals with this for us.
+                    let mut new_data_node = Self::defined(new_data);
+
+                    // If the insertion does not happen at the beginning, then we keep a portion of
+                    // the undefined range. The length of the existing undefined range is not
+                    // considered, as we are only interested in filling the gap.
+                    if undefined_prefix_len > 0 {
+                        let prefix = Self::undefined(undefined_prefix_len);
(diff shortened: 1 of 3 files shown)


vapourismo:
* **#731** <a href="https://app.graphite.com/github/pr/tezos/riscv-pvm/731?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#732](https://github.com/tezos/riscv-pvm/pull/732) <a href="https://app.graphite.com/github/pr/tezos/riscv-pvm/732?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#736](https://github.com/tezos/riscv-pvm/pull/736) <a href="https://app.graphite.com/github/pr/tezos/riscv-pvm/736?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/tezos/riscv-pvm/731?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

data/src/partial_vec.rs
```diff
@@ -0,0 +1,441 @@
...
+                    // needs to be kept.
+                    if undefined_suffix_len > 0 {
+                        let suffix = Self::undefined(undefined_suffix_len);
+                        new_data_node = new_data_node.concat(suffix);
+                    }
+
+                    *target = new_data_node;
+                }
+
+                PartialVec::Defined { data } => {
```
NSant215:
```suggestion
                Self::Defined { data } => {
```

data/src/partial_vec.rs
```diff
@@ -0,0 +1,441 @@
+// SPDX-FileCopyrightText: 2025 TriliTech <contact@trili.tech>
```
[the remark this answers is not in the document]
vapourismo: Fixed in this [diff](https://github.com/tezos/riscv-pvm/compare/98b80a564a87cf3a9ad58a911d254225ee4684ea..2e5e7ed2a550fb9a789c4c9c0b59b438daa58ee5).

data/src/partial_vec.rs
```diff
@@ -0,0 +1,441 @@
...
+                        *target = Self::defined(data).concat(new_node);
+
+                        continue;
+                    }
+
+                    let new_data_end = offset + new_data.len();
+
+                    // When we start inserting within the defined range, but the new data extends
+                    // beyond the existing range, we can truncate and append in place.
+                    if new_data_end >= data.len() {
```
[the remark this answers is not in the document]
vapourismo: I added a bit more commentary to reference that the `Concatenated` case gives us these guarantees.

vapourismo (commented):
(GitHub is kidnapping my comments, so I had to send them via the "review" function).

data/src/partial_vec/tests.rs
```diff
@@ -0,0 +1,507 @@
+// SPDX-FileCopyrightText: 2026 TriliTech <contact@trili.tech>
+//
+// SPDX-License-Identifier: MIT
+
+//! Tests for [`PartialVec]
```
thomasathorne: Super nit, I don't think this bit of doc will be generated anyway?

```suggestion
//! Tests for [`PartialVec`]
```

data/src/partial_vec/tests.rs
```diff
@@ -0,0 +1,507 @@
...
+    // After truncating to 0, the vector should have no defined data
+    assert!(!vec.is_partially_defined(0..5));
+}
+
+/// Check that truncating an undefined vector works correctly.
+#[test]
+fn truncate_undefined_vector() {
+    // BUG: When truncating an Undefined node, the code does `length -= *undefined_length`
+    // which causes underflow when *undefined_length > length.
+    // It should instead set *undefined_length = length.
```
kurtisc: Are these tests that these bugs have been fixed? The comments make it seem like this is what they do, rather than what they did.

data/src/partial_vec/tests.rs
```diff
@@ -0,0 +1,507 @@
...
+        .copied()
+        .collect::<Vec<_>>();
+    assert_eq!(values, vec![1, 2, 3]);
+}
+
+/// Check that truncating to zero on a defined vector works correctly.
+#[test]
+fn truncate_to_zero_defined() {
+    // BUG: Truncating to 0 should clear the vector, but the loop condition
+    // `while length > 0` prevents this from happening.
```
kurtisc: See comment on `truncate_undefined_vector`

data/src/partial_vec.rs
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    /// Create a vector of undefined data with the provided length.
+    pub fn undefined(length: usize) -> Self {
+        Self::Undefined { length }
+    }
+
+    /// Concatenation of two partial vectors into a new one.
+    fn concat(self, other: Self) -> Self {
+        Self::Concatenated {
```
victor-dumitrescu: When concatenating 2 partial vectors of the same type, why not create a new partial vector of that type instead of a concatenation? I assume it's for proof safety, to avoid having large copies in one step?

data/src/partial_vec.rs
```diff
@@ -0,0 +1,448 @@
...
+            while !range.is_empty() {
+                let Some(node) = work.pop_front() else {
+                    // If we ran out of things to do, but the range is not empty yet, then the partial
+                    // vector is undefined for the rest of the queried range.
+                    let entry = RangeEntry::Undefined {
+                        length: range.len(),
+                    };
+
+                    // We must drain the range to prevent further iterations.
+                    range = 0..0;
```
victor-dumitrescu: Nit, could use `Range::default()`

NSant215 (commented):
Generally LGTM, just wanna sanity check against test coverage as I can't see these cases covered but could easily be missing them:
- defining beyond current width of a vec but overlapping 
- `defined_range` querying a range with no defined ranges.

Also is there a reason you went for the name `PartialVec` rather than `SparseVec`?

data/src/partial_vec.rs
```diff
@@ -0,0 +1,448 @@
...
+        loop {
+            // SAFETY: Dereferencing `target` is safe because we have not aliased the pointer.
+            match unsafe { &mut *target } {
+                Self::Undefined { length } => {
+                    *length = keep_length.min(*length);
+                    break;
+                }
+
+                Self::Defined { data } => {
+                    data.truncate(keep_length);
```
danielpapp-trilitech: Is being "undefined" implicit for the piece that we cut off here?

data/src/partial_vec.rs
```diff
@@ -0,0 +1,448 @@
...
+    pub fn truncate(&mut self, mut keep_length: usize) {
+        // The borrow checker is unfortunately a little in the way in this method. Please see the
+        // `PartialVec::Concatenated` case below for more information.
+        let mut target: *mut Self = self;
+
+        loop {
+            // SAFETY: Dereferencing `target` is safe because we have not aliased the pointer.
+            match unsafe { &mut *target } {
+                Self::Undefined { length } => {
+                    *length = keep_length.min(*length);
```
danielpapp-trilitech: Why are we cutting here? I mean this range is already undefined, so matching on this should be a no-op, no?

data/src/partial_vec.rs
```diff
@@ -0,0 +1,448 @@
...
+/// Partial vector that maybe not be defined for certain ranges
+///
+/// This data type can be seen as a `BTreeMap<usize, T>` which provides a more memory-efficient
+/// representation for when there are many adjacent keys and the primary query mechanism is using
+/// ranges.
+#[derive(Debug, Clone)]
+pub enum PartialVec<T> {
+    /// Undefined vector of a certain width
+    Undefined {
+        /// How many bytes are undefined
```
thomasathorne: This was unintuitive (to me at least) so I'd add something emphasising it is bytes and _not_ elements (like you've done below for `node_width`).

thomasathorne (commented):
Looks nice so far, a few comments. Will keep reviewing.

data/src/partial_vec.rs
vapourismo: Yes.

data/src/partial_vec.rs
```diff
@@ -0,0 +1,448 @@
...
+        self.range(range).filter_map(move |chunk| match chunk {
+            RangeEntry::Undefined { length } => {
+                start += length;
+                None
+            }
+
+            RangeEntry::Defined { data } => {
+                let new_start = start + data.len();
+                let item = (start, data);
+                start = new_start;
```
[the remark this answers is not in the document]
vapourismo: Argh no! That's an artefact from a previous version. I applied your suggestion.

data/src/partial_vec.rs
vapourismo: It's possible to do that. I hadn't added it until now because that case never happens.

I added it now, given it might be useful in the future. I'm hoping the inlining will take care of preventing the additional cycles spend on checking the variants.

data/src/partial_vec.rs
vapourismo: I tried it without updating the length. There are enough indirect guardrails in place that this will currently not break things.

My initial assessment was this:

image

If we don't update the length, this will happen:

image

The data structure itself is in an incoherent state. However! The current guardrails seem to make the data structure converge towards a coherent state - at least for the operations I've tried.

I don't fully trust this yet, though. So I'd prefer to update the length of undefined nodes as a matter of precaution.

NSant215 (approved):

thomasathorne (changes requested):
This generally looks great, the PR was fun to review!

My one concern is that there is nothing keeping the tree balanced. So for example
```
#[test]
fn unbalanced_tree() {
    let mut vec: PartialVec<u8> = PartialVec::undefined(0);

    for i in 0..40000 {
        vec.define(i * 2 + 1, vec![0]);
    }
}
```
is enough to fail with a stack overflow.

That is obviously a long way from the intended use case, but 40,000 entries isn't _that_ many. I'd argue we should either consider fixing this or be very careful not to use this in any situations where an adversary can choose the shape of the data.

data/src/partial_vec/tests.rs
```diff
@@ -0,0 +1,510 @@
...
+    // Now create a structure with defined regions on both sides
+    vec.define(0, vec![1, 2, 3]); // Left side: indices 0-2
+    vec.define(7, vec![7, 8, 9]); // Right side: indices 7-9
+
+    // Now define data that spans across the middle, overlapping both regions
+    // This should trigger the cross-boundary insertion code path
+    vec.define(2, vec![20, 30, 40, 50, 60]);
+
+    // Verify the result: indices 0-1 should be [1, 2], indices 2-6 should be [20, 30, 40, 50, 60]
+    assert!(vec.is_any_defined(0..7));
```
thomasathorne: This line isn't actually checking what the comment above claims.

emturner (commented):
overall makes sense, main comment is that I'd argue a significant number of edge cases seem to exist w.r.t. integer overflow that should be guarded against

emturner (approved):

thomasathorne (approved):
Happy to approve this now and look forward to the follow-up.

data/src/partial_vec.rs
```diff
@@ -0,0 +1,441 @@
...
+                        continue;
+                    }
+
+                    // At this point we know that the inserted range overlaps with both the left and
+                    // right nodes. So we divide the insertion vector accordingly.
+                    let left_overlap_tail = left.width() - offset;
+                    let new_right_data = new_data.split_off(left_overlap_tail);
+
+                    work.push((left, offset, new_data));
+                    work.push((right, 0, new_right_data));
```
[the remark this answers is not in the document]
vifar: Fixed in commit fb772c530. price2 is now at the top level of the orderInfo type in both helpers and accessed via orderInfo.price2.

PR #838, opened 2026-02-23
feat(pvm-storage): enable option to use `PersistenceLayer` in pvm storage

thomasathorne:
Closes TZX-72
Closes TZX-73

# What

This PR adds a feature flag `rocksdb` that allows the PVM storage to use the persistence layer (without changing the chunking or folding logic at all).

# Why

For now we want the option of using the persistence layer for benchmarking, in future we may switch to using it for all PVM commits.

# How

We add a module `storage::rocksdb_store` that contains a wrapper around the persistence layer. This wrapper type implements `BlobStore` and `PersistentBlobStore` meaning it can be used by the PVM storage as a drop-in replacement for the existing `Store`.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/pvm/src/storage.rs
+++ b/pvm/src/storage.rs
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: MIT
 
 mod chunked_io;
+pub mod rocksdb_store;
 
 use std::io;
 use std::io::Write;
@@ -25,6 +26,7 @@ use octez_riscv_data::store::BlobStore;
 use octez_riscv_data::store::BlobStoreError;
 use octez_riscv_data::store::fold::BlobStoreFold;
 use octez_riscv_data::store::unfold::BlobStoreUnfold;
+use octez_riscv_durable_storage::errors::OperationalError;
 use thiserror::Error;
 
 const CHUNK_SIZE: usize = 4096;
@@ -43,6 +45,9 @@ pub enum StorageError {
     #[error("Invalid repo")]
     InvalidRepo,
 
+    #[error("RocksDBStore error: {0:?}")]
+    RocksDBError(#[from] OperationalError),
+
     #[error("Committed chunk {0} not found")]
     ChunkNotFound(String),
 
@@ -53,8 +58,8 @@ pub enum StorageError {
     UnfoldError(#[from] UnfoldError),
 }
 
-/// A subtrait for `BlobStore` to provide extra functionality required by the PVM storage to export
-/// PVM snapshots.
+/// A subtrait for `BlobStore` to provide extra functionality required by the PVM storage to
+/// persist and export PVM snapshots.
 pub trait PersistentBlobStore: BlobStore {
     /// Initialise a store. Either create a new directory if `path` does not exist or initialise in
     /// an existing directory.
@@ -63,6 +68,13 @@ pub trait PersistentBlobStore: BlobStore {
     fn init_from_path(path: impl AsRef<Path>) -> Result<Self, StorageError>
     where
         Self: Sized;
+
+    /// Some implementations do not automatically persist all the blobs; when the blob store is
+    /// dropped, they will be lost. Calling this method will take a 'snapshot' of the state of the
+    /// blob store which is persistent at the `path` the store was created from.
+    ///
+    /// This will be a no-op in implementations where it is uneccessary.
+    fn persist(&self) -> Result<(), StorageError>;
 
     /// Copy a specific blob across to a different store. While this should be functionally
     /// equivalent to using `blob_get` followed by `blob_set` to copy the blob across, it could in
@@ -125,6 +137,11 @@ impl PersistentBlobStore for Store {
         Ok(Store {
             path: path.into_boxed_path(),
         })
+    }
+
+    fn persist(&self) -> Result<(), StorageError> {
+        // With `Store` the blobs are automatically persisted, so this is a no-op.
+        Ok(())
     }
 
     fn export_blob(&self, other: &mut Self, hash: Hash) -> Result<(), StorageError> {
@@ -208,6 +225,9 @@ impl<BS: PersistentBlobStore> Repo<BS> {
             self.backend.export_blob(&mut other, chunk)?;
         }
         self.backend.export_blob(&mut other, id)?;
+
+        other.persist()?;
+
         Ok(())
     }
 
@@ -227,15 +247,15 @@ impl<BS: PersistentBlobStore> Repo<BS> {
         let source = BlobStoreUnfold::new(Arc::clone(&self.backend), id);
         let state = State::unfold(source)?;
 
-        let other = Self::init_empty(path)?;
-        let builder = BlobStoreFold::from(Arc::new(other));
+        let other = Arc::new(Self::init_empty(path)?);
+        let builder = BlobStoreFold::from(Arc::clone(&other));
         state.fold(builder)?;
+
+        other.persist()?;
 
         Ok(())
     }
-}
 
-impl<BS: BlobStore> Repo<BS> {
     pub fn new(store: BS) -> Self {
         Repo {
             backend: Arc::new(store),
@@ -259,6 +279,8 @@ impl<BS: BlobStore> Repo<BS> {
         let hashed_commit_bytes = HashedData::from_data(commit_bytes);
         self.backend.blob_set(&hashed_commit_bytes)?;
 
+        self.backend.persist()?;
+
         Ok(hashed_commit_bytes.hash())
     }
 
@@ -275,6 +297,8 @@ impl<BS: BlobStore> Repo<BS> {
         let hashed_commit_bytes = HashedData::from_data(commit_bytes);
         self.backend.blob_set(&hashed_commit_bytes)?;
 
+        self.backend.persist()?;
+
         Ok(hashed_commit_bytes.hash())
     }
 
@@ -284,8 +308,11 @@ impl<BS: BlobStore> Repo<BS> {
         subject: &impl Foldable<BlobStoreFold<BS>>,
     ) -> Result<Hash, StorageError> {
         let builder = BlobStoreFold::from(Arc::clone(&self.backend));
+        let hash = subject.fold(builder)?;
 
-        Ok(subject.fold(builder)?)
+        self.backend.persist()?;
+
+        Ok(hash)
     }
 
     /// Checkout the bytes committed under `id`, if the commit exists.
--- a/pvm/src/storage/rocksdb_store.rs
+++ b/pvm/src/storage/rocksdb_store.rs
@@ -1,0 +1,103 @@
+// SPDX-FileCopyrightText: 2026 Trilitech <contact@trili.tech>
+//
+// SPDX-License-Identifier: MIT
+
+#![cfg(feature = "rocksdb")]
+
+//! Implementation of `BlobStore` that can persist to disk using RocksDB.
+
+use std::path::Path;
+use std::path::PathBuf;
+
+use octez_riscv_data::hash;
+use octez_riscv_data::store::BlobStore;
+use octez_riscv_data::store::BlobStoreError;
+use octez_riscv_durable_storage::errors::Error;
+use octez_riscv_durable_storage::errors::InvalidArgumentError;
+use octez_riscv_durable_storage::persistence_layer::PersistenceLayer;
+use octez_riscv_durable_storage::repo::DirectoryManager;
+use octez_riscv_durable_storage::storage::KeyValueStore;
+use octez_riscv_durable_storage::storage::PersistentKeyValueStore;
+
+use super::PersistentBlobStore;
+use super::StorageError;
+
+/// An on-disk blob store implementation that uses RocksDB. Basically a fairly simple wrapper
+/// around the persistence layer implemented in the durable storage crate.
+///
+/// Does not behave exactly like `Store` in terms of persistence; when it is dropped, the
+/// persistence layer deletes its main 'temporary' directory, so in order to make blobs persistent
+/// beyond dropping this struct you must call the `persist` method.
+pub struct RocksDBStore {
+    inner: PersistenceLayer,
+    persist_path: PathBuf,
+    tmp_commit_path: PathBuf,
+}
+
+fn from_durable_storage_error(key: hash::Hash, e: Error) -> BlobStoreError {
+    match e {
+        Error::InvalidArgument(InvalidArgumentError::KeyNotFound) => BlobStoreError::NotFound(key),
+        _ => BlobStoreError::Custom(Box::new(e)),
+    }
+}
+
+impl BlobStore for RocksDBStore {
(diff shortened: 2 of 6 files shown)


pvm/src/storage.rs
```diff
@@ -64,6 +65,13 @@ pub trait PersistentBlobStore: BlobStore {
     where
         Self: Sized;
 
+    /// Some implementations do not automatically persist all the blobs; when the blob store is
+    /// dropped, they will be lost. Calling this method will take a 'snapshot' of the state of the
+    /// blob store which is persistent at the `path` the store was created from.
+    ///
+    /// This will be a no-op in implementations where it is uneccessary.
+    fn persist(&self) -> Result<(), StorageError>;
```
kurtisc, line 73: Could we just make persisting part of the Drop?

pvm/src/storage/rocksdb_store.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    fn persist(&self) -> Result<(), StorageError> {
+        if self.persist_path.as_path().exists() {
+            std::fs::remove_dir_all(self.persist_path.as_path())?;
+        }
+
+        self.inner
+            .commit_to_path(self.persist_path.as_path())
+            .map_err(|e| BlobStoreError::Custom(Box::new(e)))?;
```
victor-dumitrescu, line 107: If a previously persisted store exists, it is deleted before we commit the current state. If the commit were to fail, the previously persisted store would already be removed. This might be an issue if we need to recover from a known good state? If that's the case, maybe it would make sense to first commit, then remove the old store and rename the path?

pvm/src/storage/rocksdb_store.rs
```diff
@@ -0,0 +1,94 @@
...
+            .map_err(|e| BlobStoreError::Custom(Box::new(e)))
+    }
+}
+
+impl PersistentBlobStore for RocksDBStore {
+    fn init_from_path(path: impl AsRef<Path>) -> Result<Self, StorageError>
+    where
+        Self: Sized,
+    {
+        let repo = DirectoryManager::new(path.as_ref()).map_err(|_| StorageError::InvalidRepo)?;
```
victor-dumitrescu: Squashing every `OperationalError` that `DirectoryManager::new` can raise into `InvalidRepo` potentially erases useful information, such as which IO error was encountered.

pvm/tests/test_pvm_storage.rs
```diff
@@ -21,6 +20,16 @@ use proptest::prelude::*;
 use proptest::strategy::ValueTree;
 use proptest::test_runner::TestRunner;
 
+cfg_if::cfg_if! {
```
emturner: could you possibly add a follow up? 

we really should test that both `TestStore` and `RocksDb` work in CI. Doing this means (for now) that rocksdb won't be tested in CI.

A good follow up would be to try and allow the tests to run over both stores, but only supply `rocksdb` one if the feature flag is enabled.

pvm/src/storage/rocksdb_store.rs
```diff
@@ -0,0 +1,94 @@
...
+use octez_riscv_durable_storage::errors::InvalidArgumentError;
+use octez_riscv_durable_storage::persistence_layer::PersistenceLayer;
+use octez_riscv_durable_storage::repo::DirectoryManager;
+use octez_riscv_durable_storage::storage::KeyValueStore;
+use octez_riscv_durable_storage::storage::PersistentKeyValueStore;
+
+use super::PersistentBlobStore;
+use super::StorageError;
+
+pub struct RocksDBStore {
```
emturner, line 32: could you please add docs? (optional - also a very brief one at the module level too would be nice)

pvm/src/storage/rocksdb_store.rs
```diff
@@ -0,0 +1,94 @@
...
+    inner: PersistenceLayer,
+    persist_path: PathBuf,
+}
+
+fn from_durable_storage_error(key: hash::Hash, e: Error) -> BlobStoreError {
```
emturner, line 37: optional - perhaps could implement `From<Error> for BlobStoreError` here? Would allow you to remove most of the `.map_err` calls below. (or simplify them to `.map_err(BlobStoreError::From)`

This method is also only used once - but looks like it could be used everywhere?

pvm/src/storage/rocksdb_store.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn init_from_path(path: impl AsRef<Path>) -> Result<Self, StorageError>
+    where
+        Self: Sized,
+    {
+        let repo = DirectoryManager::new(path.as_ref()).map_err(|_| StorageError::InvalidRepo)?;
+        let tempdir = repo
+            .temp_database_dir()
+            .map_err(|e| BlobStoreError::Custom(Box::new(e)))?;
+
+        let persist_path = PathBuf::new().join(path).join("persisted");
```
emturner:
```suggestion
        let persist_path = PathBuf::from(path.as_ref()).join("persisted");
```

pvm/src/storage/rocksdb_store.rs
emturner: I'd even argue persisting to a path that already exists, should even be a hard error of some kind?

pvm/src/storage/rocksdb_store.rs
emturner: nit: could you possibly move this into a constant? We might want to group these `DirManager` constants together sometime in the future

pvm/src/storage/rocksdb_store.rs
emturner: we'd likely want a mechanism for `checkout`, then `commit` into a new place automatically I think
(perhaps by `root_hash`?)

(would it be possible to add a follow up TODO for this?)

pvm/src/storage.rs
emturner: > /// Some implementations do not automatically persist all the blobs

Why wouldn't they?

emturner: > Could we just make persisting part of the Drop?

I suspect that itself might not be desirable - e.g. you might not want to persist if the fold `panics` for example

pvm/src/storage/rocksdb_store.rs
thomasathorne: The reason this is only used once is because all the other errors are `OperationalError` not `Error`.

I tried implementing `From<Error>` (and `From<OperationalError>`) but I couldn't as both `Error` and `BlobStoreError` are defined in different crates. 🤦‍♂️ I could put those impls in the durable storage `errors` module but they'd look a bit out of place, there's nothing else there that is blob-store related. 🤔

pvm/src/storage/rocksdb_store.rs
thomasathorne: Done [here](https://github.com/tezos/riscv-pvm/compare/f184afca48b12103e7de10adf42f217864c272a7..612fd1d5962515dcfd7cc3d9a75e74c8d4152860)

pvm/src/storage/rocksdb_store.rs
thomasathorne: Sorry, added those [here](https://github.com/tezos/riscv-pvm/compare/612fd1d5962515dcfd7cc3d9a75e74c8d4152860..82855606367c477642374d2f1c5238920801d923). Also switched the name from `persist` to `commit` as you suggested off-line.

pvm/src/storage/rocksdb_store.rs
thomasathorne: Good point, added an optional field to contain the IO error or whatever [here](https://github.com/tezos/riscv-pvm/compare/82855606367c477642374d2f1c5238920801d923..feca84dde93185cbe13ed8dc8ce351bbf98db153).

pvm/src/storage/rocksdb_store.rs
thomasathorne: On second thoughts, the term 'commit' is kind of overloaded already here and it's now confusing as there's a `commit` method on the `Store` and on the `Repo` type as well... 🤦‍♂️ I may leave it as `persist`.

thomasathorne: I've switched back to `persist` [here](https://github.com/tezos/riscv-pvm/compare/a2aa4659cd661cc72d755c8090894db615fa879a..ac5a94d5c91cded5d442adde36299c40b9bc5a4e)

pvm/src/storage/rocksdb_store.rs
thomasathorne: Have followed the plan of committing, then deleting, then renaming the path [here](https://github.com/tezos/riscv-pvm/compare/ac5a94d5c91cded5d442adde36299c40b9bc5a4e..f503330a95c7dcf926b338175ae3aac540b777f6).

Will also create a follow-up issue.

thomasathorne: Created TZX-130 and added a TODO [here](https://github.com/tezos/riscv-pvm/compare/f503330a95c7dcf926b338175ae3aac540b777f6..cf378ba2499811f9bc3020099d84d7a8c216f576)

pvm/src/storage/rocksdb_store.rs
thomasathorne: I'm a bit confused which part of this expression could be a constant. 🤔 The `persist_path` is stored in the struct because it depends on the struct's own `DirectoryManager` path.

thomasathorne: Ah you just mean the string `"persisted"`? This is actually the only place in the code where it appears, and it's going to be gone after TZX-130 anyway. 🤔

pvm/tests/test_pvm_storage.rs
thomasathorne: I was a bit unsatisfied with this myself, actually wasn't too much work to fix it [here](https://github.com/tezos/riscv-pvm/compare/cf378ba2499811f9bc3020099d84d7a8c216f576..6af88a597ecbbe0a6ffe902fe6597b5cc2c44183). So now with the feature disabled it just runs over `Store`, with the feature enabled there's a second set of tests that run over `RocksDBStore` as well as the first ones.

pvm/src/storage.rs
thomasathorne: After discussing this I'm leaving it as it is. After TZX-130 this will be a lot nicer anyway.

pvm/src/storage/rocksdb_store.rs
victor-dumitrescu: I think this might overcomplicate `InvalidRepo`. What do you think about a new `StorageError` variant, e.g., `StoreInitError(#[from] OperationalError)`?

pvm/src/storage/rocksdb_store.rs
victor-dumitrescu: I'm not sure it's a good idea to rely on always being able to create a temporary directory, I think the temporary path should be in the same tree as the persistent one.

thomasathorne: Hmm, I see your point, ok I've switched it [here](https://github.com/tezos/riscv-pvm/compare/5617031c57e7099545c2b4bca5b5160d43197ed2..bf2d7e35d81363f4f0210905aa2cee59d0cb33a6) so it now uses another path in the same (originally provided) dir.

pvm/src/storage/rocksdb_store.rs
thomasathorne: How does [this](https://github.com/tezos/riscv-pvm/compare/bf2d7e35d81363f4f0210905aa2cee59d0cb33a6..563d1461b45a2e9fff60604ab79a82e18c686420) look?

victor-dumitrescu: Sorry, thinking about this a bit more, I think you're right that you do in fact want these to fall under the same error. I still think the optional boxed error is a bit clunky, but perhaps something like:

```rust
#[derive(Error, Debug)]
pub enum StorageError {
...

    #[error("Repo error")]
    RepoError(#[from] RepoError),
}

#[derive(Error, Debug)]
pub enum RepoError {
    #[error("Invalid repo")]
    InvalidRepo,

    #[error("RocksDBStore error: {0:?}")]
    RocksDBError(#[from] OperationalError),
}
```

It's all a bit messy because IO errors can be raised at the storage level but some persistence layer operational errors are also ultimately IO errors  😅

But this test code makes it clear to me that my initial suggestion wasn't great:

```rust
    assert!(matches!(
        load_result,
        Err(StorageError::InvalidRepo | StorageError::RocksDBError(_))
    ));
```

thomasathorne: The reason for that test code is more the fact that the test 'helper' function has to be generic over the store and work for both `RocksDBStore` and `Store`, it's not non-deterministic or anything.

I like the new version as it gets rid of a few `map_err` calls, and `RocksDBError` is at least very explicit where the error comes from. At this point I've given up trying to make error types perfectly match the semantics of different failures. 🤷‍♂️

pvm/tests/test_pvm_storage.rs
```diff
@@ -79,21 +93,32 @@ fn test_repo() {
...
+    test_repo_serialised_helper::<Store>();
+}
+
+#[test]
+#[cfg(feature = "rocksdb")]
+fn test_repo_serialised_rocks_db() {
+    test_repo_serialised_helper::<RocksDBStore>();
+}
+
+#[cfg(test)]
```
victor-dumitrescu: I think you can declare this whole module `#![cfg(test)]` and remove all these attributes.

victor-dumitrescu (approved):

pvm/tests/test_pvm_storage.rs
thomasathorne: Done [here](https://github.com/tezos/riscv-pvm/compare/563d1461b45a2e9fff60604ab79a82e18c686420..2c1a0cc2eff3d50c49467f195f709d5e29f53e34)

pvm/tests/test_pvm_storage.rs
emturner: ah awesome, thanks!

Would it be possible to also follow up to that, and add a macro helper for this? (would be similar to `mode_test!` or the new `kv_test!` that [victor is adding](https://github.com/tezos/riscv-pvm/pull/1009)?

pvm/src/storage/rocksdb_store.rs
emturner: ah fair enough 👍🏻

pvm/tests/test_pvm_storage.rs
thomasathorne: Ah ok, I was avoiding the macro but you're right, it's not too tricky: [here](https://github.com/tezos/riscv-pvm/compare/2c1a0cc2eff3d50c49467f195f709d5e29f53e34..6979fd8c7e40bfb469e07a24ee587bbc76f3c3e4)

emturner (approved):

PR #989, opened 2026-04-09
feat(dursto): Round-trip of Normal-Proof-Verify for MAVL Tree for non-structural-change operations

NSant215:
Closes RV-962.
Part of RV-961.

# What

Introduce `MerkleProofFold` and `PartialHashFold` implementations over MAVL objects to allow for proof production and verification. 

# Why

Working towards completing proof support in DS.

# How

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/data/src/merkle_proof/proof_tree.rs
+++ b/data/src/merkle_proof/proof_tree.rs
@@ -305,6 +305,19 @@ impl MerkleProofFold {
     pub fn new_leaf(constraint: MinimumPresence, data: Vec<u8>) -> impl Foldable<Self> {
         CompressibleMerkleProof::new_leaf(constraint, data)
     }
+
+    /// Create a blinded proof leaf from a known hash.
+    ///
+    /// Use this when a subtree was not accessed during proof generation and should be represented
+    /// only by its hash in the proof.
+    /// TODO: RV-968 - remove workaround for creating CompressibleMerkleProof, as users can also be
+    /// in "may-blind" state.
+    pub fn into_blind(self, hash: Hash) -> CompressibleMerkleProof {
+        CompressibleMerkleProof {
+            constraint: MinimumPresence::MayOmit,
+            tree: MerkleProof::leaf_blind(hash),
+        }
+    }
 }
 
 impl Fold for MerkleProofFold {
--- a/durable-storage/src/avl/resolver.rs
+++ b/durable-storage/src/avl/resolver.rs
@@ -453,6 +453,18 @@ pub struct ProveNodeId {
     node: LazyNodeId,
 }
 
+// TODO RV-895: This method is implemented to satisfy trait-bounds in `tree::upsert`.
+// The current implementation around proofs for created nodes during proof-generation
+// needs to be fixed.
+impl From<Node<ProveTreeId, Prove<'static>>> for ProveNodeId {
+    fn from(node: Node<ProveTreeId, Prove<'static>>) -> Self {
+        ProveNodeId {
+            inner: OnceLock::from(Rc::new(node)),
+            node: LazyNodeId::from(Hash::hash_bytes(&[])),
+        }
+    }
+}
+
 impl Foldable<HashFold> for ProveNodeId {
     fn fold(&self, builder: HashFold) -> <HashFold as Fold>::Folded {
         match self.inner.get() {
@@ -480,6 +492,17 @@ impl Foldable<MerkleProofFold> for ProveNodeId {
 pub struct ProveTreeId {
     inner: OnceLock<Tree<ProveNodeId>>,
     tree: LazyTreeId,
+}
+
+// `Node::new` requires `TreeId: Default`. TODO RV-895: fix implementation around proofs
+// for created nodes during proof-generation.
+impl Default for ProveTreeId {
+    fn default() -> Self {
+        ProveTreeId {
+            inner: OnceLock::from(Tree::default()),
+            tree: LazyTreeId::default(),
+        }
+    }
 }
 
 impl Foldable<HashFold> for ProveTreeId {
@@ -729,8 +752,14 @@ mod tests {
     use std::sync::atomic::AtomicUsize;
     use std::sync::atomic::Ordering;
 
+    use octez_riscv_data::components::atom::AtomMode;
     use octez_riscv_data::components::bytes::Bytes;
+    use octez_riscv_data::components::bytes::BytesMode;
     use octez_riscv_data::hash::Hash;
+    use octez_riscv_data::hash::PartialHash;
+    use octez_riscv_data::merkle_proof::FromProof;
+    use octez_riscv_data::merkle_proof::proof_tree::MerkleProof;
+    use octez_riscv_data::merkle_proof::proof_tree::ProofTree;
     use octez_riscv_data::mode::Normal;
     use octez_riscv_data::mode::Verify;
     use octez_riscv_data::mode::utils::NotFound;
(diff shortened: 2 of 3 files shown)


emturner:
> Closes [RV-961](https://linear.app/tezos/issue/RV-961).

Possibly, could you split that issue up into two sub issues? (we should follow up with delete and insert)

durable-storage/src/avl/resolver.rs
```diff
@@ -1386,4 +1459,93 @@ mod tests {
...
+        let merkle_proof = MerkleProof::from_foldable(&prove_tree);
+
+        // ---- Verify: deserialize proof, replay operations ----
+        let proof_deser = ProofTree::Present(&merkle_proof);
+        let verify_tree_id = VerifyTreeId::from_proof(proof_deser)
+            .expect("proof deserialization should succeed")
+            .into_result();
+
+        let VerifyTreeId::Present(mut verify_tree) = verify_tree_id else {
+            panic!("expected Present tree from proof, got blinded or absent");
```
victor-dumitrescu:
```suggestion
            panic!("expected Present tree from proof, got {:verify_tree_id?}");
```

Would be good to know what it is in case of a failure. It only requires deriving `Debug` for a few more types.

durable-storage/src/avl/resolver.rs
```diff
@@ -1386,4 +1459,93 @@ mod tests {
 
         assert!(resolved.root().is_some(), "tree should have a root");
     }
+
+    /// Build a three-node tree, persist it, then run a round-trip through Prove and Verify
+    /// modes using only read (`get`) and in-place write operations (no structural changes).
+    #[test]
+    fn prove_verify_round_trip_get_and_write() {
```
victor-dumitrescu, line 1504: Any reason why this couldn't be generalised to a PBT which tests larger trees and a wider range of keys and values? As long as overwritten values are never larger than existing ones for the moment.

victor-dumitrescu (commented):
I mostly reviewed the test, IIUC most trait implementations will have to change in order to support structural changes?

emturner:
> I mostly reviewed the test, IIUC most trait implementations will have to change in order to support structural changes?

They will - mainly just trying to get something working so we have more context as to how things break as we introduce the structural changes

durable-storage/src/avl/resolver.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/6b2ae987b271f0be644e02836b4d7e36224a6311..4139682)

durable-storage/src/avl/resolver.rs
kurtisc: It's hard to create a PBT which you know will produce trees that don't rotate. e.g. something as simple as:

`[0], [1], [2]`

Will rotate, but:

`[1], [0], [2]`

won't

victor-dumitrescu: Ah that's a good point, thanks!

durable-storage/src/avl/resolver.rs
victor-dumitrescu: I think you didn't mean to remove the "got" and add a question mark at the end? Resolve at will.

victor-dumitrescu (approved):

data/src/merkle_proof/proof_tree.rs
```diff
@@ -305,6 +305,17 @@ impl MerkleProofFold {
...
+    /// Create a blinded proof leaf from a known hash.
+    ///
+    /// Use this when a subtree was not accessed during proof generation and should be represented
+    /// only by its hash in the proof.
+    pub fn into_blind(self, hash: Hash) -> CompressibleMerkleProof {
+        CompressibleMerkleProof {
+            constraint: MinimumPresence::MayOmit,
+            tree: MerkleProof::leaf_blind(hash),
+        }
+    }
```
emturner, line 320: I wonder if this would be possible to not add, if we switch from `into_leaf` being `foldable` to some degree? 

ie - if something is `MayOmit` or `MayBlind`, then we only require it to be `HashFoldable`

E.g. I imagine at a minimum this might be better to take `minimum_prescence` too and panic if it's present? 
Could also (possibly) leave a TODO on this to come back to it. 

I'm worried atm with extending this that we might easily `Omit` what would actually need to be `MayBlind`

durable-storage/src/avl/resolver.rs
```diff
@@ -459,6 +462,15 @@ impl Foldable<HashFold> for ProveNodeId {
     }
 }
 
+impl Foldable<MerkleProofFold> for ProveNodeId {
+    fn fold(&self, builder: MerkleProofFold) -> <MerkleProofFold as Fold>::Folded {
+        match self.inner.get() {
+            Some(inner) => inner.as_ref().fold(builder),
+            None => builder.into_blind(Hash::from_foldable(&self.node)),
```
emturner, line 481: ie this could easily require `MayBlind` here I think (definitely a TODO to follow up)

durable-storage/src/avl/resolver.rs
```diff
@@ -1432,4 +1459,93 @@ mod tests {
...
+        tree.set(&key1, b"alpha", &mut resolver)
+            .expect("set should succeed");
+        tree.set(&key2, b"beta", &mut resolver)
+            .expect("set should succeed");
+        tree.set(&key3, b"gamma", &mut resolver)
+            .expect("set should succeed");
+
+        let root_hash = Hash::from_foldable(tree.root().expect("tree should have a root node"));
+
+        let persistence_layer = Arc::new(InMemoryKeyValueStore::default());
```
emturner: nit - could you use the `setup_repo` possibly? Then this test will run over both in memory and rocksdb 🙏🏻

durable-storage/src/avl/resolver.rs
```diff
@@ -1432,4 +1459,93 @@ mod tests {
...
+        let verify_tree_id = VerifyTreeId::from_proof(proof_deser)
+            .expect("proof deserialization should succeed")
+            .into_result();
+
+        let VerifyTreeId::Present(mut verify_tree) = verify_tree_id else {
+            panic!("expected Present tree from proof, got {verify_tree_id:#?}");
+        };
+        let mut verify_resolver = VerifyResolver;
+
+        let final_hash = catch_not_found(move || {
```
emturner, line 1565: could you possibly add a TODO to rewrite some of these tests using an 'OperationStream' in future? 

Alternatively, would it be possible to have a generic function that runs the same operations over both Proof and Verify trees?

emturner (approved):

data/src/merkle_proof/proof_tree.rs
NSant215: you're right - added TODO with linear issue [here](https://github.com/tezos/riscv-pvm/compare/9e0aac7..1c20ba12538ea607705018de8cfe85b91ee34fde)

durable-storage/src/avl/resolver.rs
NSant215: addressed above

durable-storage/src/avl/resolver.rs
NSant215: ah yes - done [here](https://github.com/tezos/riscv-pvm/compare/1c20ba12538ea607705018de8cfe85b91ee34fde..7f56dfe803fc8648d1662e671e00c4033936893a)

durable-storage/src/avl/resolver.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/1c20ba1..3da6a5fb53dc6f4d117e4aac1f9caea4b2e7c9b7)

PR #993, opened 2026-04-10
feat(pvm-storage): chunked io using `BlobStore`

thomasathorne:
Part of TZX-86.

# What

Modify the 'chunked IO' methods `commit_serialised` and `checkout_serialised` that are currently used for PVM storage so that they can run against any implementation of the `BlobStore` trait.

We also add an extension to that trait, `ExportStore`, that allows copying blobs from one store into another at a given filesystem path. This is needed to implement the `export_snapshot` method. Having this extra trait will make it easier to switch to using the durable-storage persistence-layer, as that can easily be made to implement `ExportStore` on top of `BlobStore`.

# Why

We want to benchmark the existing (chunked) commit and checkout as well as the new (folded) methods. If both approaches can be run against an arbitrary `BlobStore` it makes the benchmarking much easier.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/pvm/src/storage.rs
+++ b/pvm/src/storage.rs
@@ -15,7 +15,6 @@ use bincode::Encode;
 use bincode::error::DecodeError;
 use bincode::error::EncodeError;
 use octez_riscv_data::hash::Hash;
-use octez_riscv_data::hash::HashError;
 use octez_riscv_data::hash::HashedData;
 use octez_riscv_data::serialisation;
 use octez_riscv_data::store::BlobStore;
@@ -38,14 +37,34 @@ pub enum StorageError {
     #[error("Invalid repo")]
     InvalidRepo,
 
-    #[error("Hashing error")]
-    HashError(#[from] HashError),
-
-    #[error("Data for hash {0} not found")]
-    NotFound(String),
-
     #[error("Committed chunk {0} not found")]
     ChunkNotFound(String),
+
+    #[error("Blob store error")]
+    BlobStore(#[from] BlobStoreError),
+}
+
+/// A subtrait for `BlobStore` to provide extra functionality required by the PVM storage to export
+/// PVM snapshots.
+pub trait PersistentBlobStore: BlobStore {
+    /// Initialise a store. Either create a new directory if `path` does not exist or initialise in
+    /// an existing directory.
+    ///
+    /// Should throw [`StorageError::InvalidRepo`] if `path` is a file.
+    fn init_from_path(path: impl AsRef<Path>) -> Result<Self, StorageError>
+    where
+        Self: Sized;
+
+    /// Copy a specific blob across to a different store. While this should be functionally
+    /// equivalent to using `blob_get` followed by `blob_set` to copy the blob across, it could in
+    /// many impls be more efficient (especially for large blobs) by using `std::fs::copy` or
+    /// equivalent instead.
+    fn export_blob(&self, other: &mut Self, hash: &Hash) -> Result<(), StorageError> {
+        let blob = self.blob_get(*hash)?;
+        other.blob_set(&HashedData::from_data(blob.as_ref()))?;
+
+        Ok(())
+    }
 }
 
 #[derive(Debug, PartialEq)]
@@ -54,21 +73,6 @@ pub struct Store {
 }
 
 impl Store {
-    /// Initialise a store. Either create a new directory if `path` does not
-    /// exist or initialise in an existing directory.
-    /// Throws `StorageError::InvalidRepo` if `path` is a file.
-    pub fn init(path: impl AsRef<Path>) -> Result<Self, StorageError> {
-        let path = path.as_ref().to_path_buf();
-        if !path.exists() {
-            std::fs::create_dir(&path)?;
-        } else if path.metadata()?.is_file() {
-            return Err(StorageError::InvalidRepo);
-        }
-        Ok(Store {
-            path: path.into_boxed_path(),
-        })
-    }
-
     fn file_name_of_hash(hash: &Hash) -> String {
         hex::encode(hash)
     }
@@ -98,23 +102,25 @@ impl Store {
         self.write_data_if_new(file_name, data)?;
         Ok(hash)
     }
+}
 
-    /// Load data corresponding to `hash`, if found.
-    pub fn load(&self, hash: &Hash) -> Result<Vec<u8>, StorageError> {
-        let file_name = self.path_of_hash(hash);
-        std::fs::read(file_name).map_err(|e| {
-            if e.kind() == io::ErrorKind::NotFound {
-                StorageError::NotFound(hex::encode(hash))
-            } else {
-                StorageError::IoError(e)
-            }
+impl PersistentBlobStore for Store {
+    fn init_from_path(path: impl AsRef<Path>) -> Result<Self, StorageError> {
+        let path = path.as_ref().to_path_buf();
+        if !path.exists() {
+            std::fs::create_dir(&path)?;
+        } else if path.metadata()?.is_file() {
+            return Err(StorageError::InvalidRepo);
+        }
+
+        Ok(Store {
+            path: path.into_boxed_path(),
         })
     }
 
-    /// Copy the data corresponding to `hash` to `path`.
-    pub fn copy(&self, hash: &Hash, path: impl AsRef<Path>) -> Result<(), StorageError> {
+    fn export_blob(&self, other: &mut Self, hash: &Hash) -> Result<(), StorageError> {
         let source_path = self.path_of_hash(hash);
-        let target_path = path.as_ref().join(Self::file_name_of_hash(hash));
+        let target_path = other.path_of_hash(hash);
         std::fs::copy(source_path, target_path)?;
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


pvm/src/storage.rs
```diff
@@ -38,14 +37,29 @@ pub enum StorageError {
...
     #[error("Committed chunk {0} not found")]
     ChunkNotFound(String),
+
+    #[error("Blob store error")]
+    BlobStore(#[from] BlobStoreError),
+}
+
+/// A subtrait for `BlobStore` to provide extra functionality required by the PVM storage to
+/// 'export' PVM snapshots.
+pub trait ExportStore: BlobStore {
```
emturner: `ExportStore` as a name seems possibly a little odd? 

Would something like `PersistentBlobStore` make more sense - and also to mirror the naming from `PersistentKeyValueStore`?

pvm/src/storage.rs
```diff
@@ -38,14 +37,29 @@ pub enum StorageError {
...
+    /// Initialise a store. Either create a new directory if `path` does not exist or initialise in
+    /// an existing directory.
+    ///
+    /// Should throw `StorageError::InvalidRepo` if `path` is a file.
+    fn init(path: impl AsRef<Path>) -> Result<Self, StorageError>
```
emturner: re naming, could go for `checkout_from_path` perhaps?

pvm/src/storage.rs
```diff
@@ -38,14 +37,29 @@ pub enum StorageError {
...
+    where
+        Self: Sized;
+
+    /// Copy a specific blob across to a different path. While this should be functionally
+    /// equivalent to creating a new store at `path` and using `blob_get` followed by `blob_set` to
+    /// copy the blob across, it should be more efficient (especially for large blobs) as it uses
+    /// `std::fs::copy` or equivalent.
+    fn copy(&self, hash: &Hash, path: impl AsRef<Path>) -> Result<(), StorageError>;
```
emturner:
```suggestion
    fn export_blob(&self, &mut other, hash: &Hash) -> Result<(), StorageError>;
```
(not certain on name, but struggling to get a better one).

would it make sense to have this opening both blob stores, then do the copy? E.g. for the persistence layer/rocks db - it would be more effecient to only connect to the database once per export, rather than once per blob per export.

Additionally, you could easily provide a default implementation of this method, and then let the chunked_io variant override it to be more efficient.

pvm/src/storage.rs
```diff
@@ -150,61 +150,120 @@ impl BlobStore for Store {
...
-pub struct Repo {
-    backend: Store,
+pub struct Repo<BS> {
+    backend: BS,
+}
+
+impl<BS> Repo<BS> {
+    pub fn backend(self) -> BS {
+        self.backend
+    }
```
emturner: why is this method needed?

pvm/src/storage.rs
thomasathorne: Good point, done [here](https://github.com/tezos/riscv-pvm/compare/c4666a95bf467bd5cbb4e34ba217a6ecb6d4c6f7..9aadbeabea6d03b8c76ab44deac1f877a70b004f). Apologies for the merge noise, the last two files of that diff are the two that are relevant. 😅

pvm/src/storage.rs
thomasathorne: Ah it's not, for this PR anyway. Think it should have been in the next PR with the benchmark itself.

Removed it [here](https://github.com/tezos/riscv-pvm/compare/9aadbeabea6d03b8c76ab44deac1f877a70b004f..be8724b927d4503612a89d07fcf8d99601cfabd0).

pvm/src/storage.rs
thomasathorne: Hmm, yeah I think it would make sense: done [here](https://github.com/tezos/riscv-pvm/compare/be8724b927d4503612a89d07fcf8d99601cfabd0..70e997179b12e46a7e2f5240c84d3c9e1121ca0c).

pvm/src/storage.rs
thomasathorne: I've called it `init_from_path` (done on the diff for the comment below this one).

pvm/src/storage/chunked_io.rs
```diff
@@ -102,13 +105,17 @@ impl<'a> ChunkedReader<'a> {
             return Ok(());
         };
 
-        self.buffer = Cursor::new(self.store.load(&hash)?);
+        let bytes = self.store.blob_get(hash)?;
+
+        self.buffer.get_mut().clear();
+        self.buffer.get_mut().extend_from_slice(bytes.as_ref());
+        self.buffer.set_position(0);
```
victor-dumitrescu, line 112: This introduces an extra clone coming from `extend_from_slice`. Since we're now operating on `impl AsRef<[u8]>` instead of concretely on `Vec<u8>`, that's _maybe_ unavoidable though?

pvm/src/storage.rs
```diff
@@ -38,14 +37,29 @@ pub enum StorageError {
...
-
     #[error("Committed chunk {0} not found")]
     ChunkNotFound(String),
+
+    #[error("Blob store error")]
+    BlobStore(#[from] BlobStoreError),
+}
+
+/// A subtrait for `BlobStore` to provide extra functionality required by the PVM storage to
+/// 'export' PVM snapshots.
```
victor-dumitrescu: Why "export" in quotes? (asking since that would suggest to me it's not exactly what this functionality does)

pvm/src/storage.rs
```diff
@@ -38,14 +37,29 @@ pub enum StorageError {
...
+pub trait PersistentBlobStore: BlobStore {
+    /// Initialise a store. Either create a new directory if `path` does not exist or initialise in
+    /// an existing directory.
+    ///
+    /// Should throw `StorageError::InvalidRepo` if `path` is a file.
```
victor-dumitrescu:
```suggestion
    /// Should throw [`StorageError::InvalidRepo`] if `path` is a file.
```

pvm/tests/test_pvm_storage.rs
```diff
@@ -62,15 +64,16 @@ fn test_repo() {
...
 
     // Check that exporting a snapshot creates a new repo which contains
     // the requested commit
     let snapshot_dir = tempfile::tempdir().unwrap();
     let (export_hash, export_data) = &test_data[0];
-    repo.export_snapshot(export_hash, &snapshot_dir).unwrap();
-    let snapshot_repo = Repo::load(tmp_dir.path()).unwrap();
+    repo.export_snapshot_chunked(export_hash, &snapshot_dir)
+        .unwrap();
+    let snapshot_repo = Repo::<Store>::load(tmp_dir.path()).unwrap();
```
victor-dumitrescu:
```suggestion
    let snapshot_repo = Repo::<Store>::load(snapshot_dir.path()).unwrap();
```

This is an existing bug, but it should check out the snapshotted repo, not the original one.

pvm/tests/test_pvm_storage.rs
```diff
@@ -127,15 +130,16 @@ fn test_repo_serialised() {
...
```
victor-dumitrescu:
```suggestion
    let snapshot_repo = Repo::<Store>::load(snapshot_dir.path()).unwrap();
```

Same here

pvm/src/storage.rs
```diff
@@ -38,14 +37,29 @@ pub enum StorageError {
...
+    fn init_from_path(path: impl AsRef<Path>) -> Result<Self, StorageError>
+    where
+        Self: Sized;
+
+    /// Copy a specific blob across to a different store. While this should be functionally
+    /// equivalent to using `blob_get` followed by `blob_set` to copy the blob across, it could in
+    /// many impls be more efficient (especially for large blobs) by using `std::fs::copy` or
+    /// equivalent instead.
+    fn export_blob(&self, other: &mut Self, hash: &Hash) -> Result<(), StorageError>;
```
emturner: if it's equivalent to `blob_get` then `blob_set` - do think it could absolutely have a default implementation already?

pvm/src/storage.rs
```diff
@@ -98,23 +97,24 @@ impl Store {
...
-            }
+impl PersistentBlobStore for Store {
+    fn init_from_path(path: impl AsRef<Path>) -> Result<Self, StorageError> {
+        let path = path.as_ref().to_path_buf();
+        if !path.exists() {
+            std::fs::create_dir(&path)?;
+        } else if path.metadata()?.is_file() {
+            return Err(StorageError::InvalidRepo);
+        }
+        Ok(Store {
```
emturner, line 116:
```suggestion

        Ok(Store {
```
(nit for readability - though I know this is moved code)

pvm/src/storage.rs
```diff
@@ -150,61 +150,115 @@ impl BlobStore for Store {
...
+        // Only export a snapshot to a new or empty directory
+        let path = path.as_ref();
+        if !path.exists() || path.read_dir()?.next().is_none() {
+            std::fs::create_dir_all(path)?;
+        } else {
+            return Err(StorageError::InvalidRepo);
+        };
+        let mut other = BS::init_from_path(path)?;
+        let bytes = self.backend.blob_get(*id)?;
+        let commit: Vec<Hash> = serialisation::deserialise(bytes.as_ref())?;
```
emturner, line 188: discussed offline - might well be worth just having this generic over something that implements `unfold` and `fold`?

pvm/src/storage.rs
thomasathorne: Sure, done [here](https://github.com/tezos/riscv-pvm/compare/edf618d0d18c6878b7de2dbbde5e0de2ea8cf58e..d50cd377457e2578827628bb17170442e4303670).

pvm/src/storage.rs
thomasathorne: Done [here](https://github.com/tezos/riscv-pvm/compare/d50cd377457e2578827628bb17170442e4303670..46f8603adce78f8c37aef47aff0f81d9b064d27c)

pvm/src/storage.rs
thomasathorne: This is `export_snapshot_chunked` which we want to keep with the old implementation (for benchmarking it). I've added `export_snapshot_folded` below, which isn't implemented yet in this PR but will be. See https://github.com/tezos/riscv-pvm/pull/994 where it is generic over something `Foldable` and `Unfoldable` as we discussed.

pvm/tests/test_pvm_storage.rs
thomasathorne: Ah good spot! 😅 Fixed [here](https://github.com/tezos/riscv-pvm/compare/46f8603adce78f8c37aef47aff0f81d9b064d27c..75fe9c768c3c1151f7135f93da1d11d94e99add3)

pvm/tests/test_pvm_storage.rs
thomasathorne: Fixed in same diff as comment above.

pvm/src/storage.rs
thomasathorne: Done [here](https://github.com/tezos/riscv-pvm/compare/75fe9c768c3c1151f7135f93da1d11d94e99add3..b9cb050521e103cfa6472c3283cccb37aeec0f55)

pvm/src/storage.rs
thomasathorne: Fair point, the quotes are misleading so I've removed them [here](https://github.com/tezos/riscv-pvm/compare/b9cb050521e103cfa6472c3283cccb37aeec0f55..5c4e07abaafe42fc01dc7a9082b18cdfceaee963)

pvm/src/storage/chunked_io.rs
thomasathorne: Ah yes, this is annoying but I can't see a way around it given the definition of the `BlobStore` trait. 🤔

Need to think about how bad this is, it's not great if we benchmark the old implementation vs the new one but we've made the old one worse in the process. 🤦‍♂️

kurtisc:
```
diff --git i/pvm/src/storage/chunked_io.rs w/pvm/src/storage/chunked_io.rs
index 9d00c1ffff..4115f04792 100644
--- i/pvm/src/storage/chunked_io.rs
+++ w/pvm/src/storage/chunked_io.rs
@@ -108,7 +108,7 @@ impl<'a, BS: BlobStore> ChunkedReader<'a, BS> {
         let bytes = self.store.blob_get(hash)?;
 
         self.buffer.get_mut().clear();
-        self.buffer.get_mut().extend_from_slice(bytes.as_ref());
+        self.buffer.get_mut().extend(bytes.as_ref());
         self.buffer.set_position(0);
 
         Ok(())
```
Should avoid a clone?

kurtisc: Probably doesn't matter though, see [u8 clone impl](https://github.com/rust-lang/rust/blob/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/clone.rs#L512)

emturner: > AsRef<[u8]> instead of concretely on Vec<u8>, that's maybe unavoidable though?

It might be possible to alter our various `impl AsRef<[u8]>` return types via a `Trait::Bytes` type?

Would need to make sure this functions correctly in conjunction with the implicit lifetime introduced by the persistence layer though.

emturner: @thomasathorne @victor-dumitrescu 

> This introduces an extra clone coming from extend_from_slice

The good news is it doesn't allocate - but yes, this does indeed introduce a memcpy. This is likely quite fast still, though. 

It might, however, be worth benchmarking it against the following as an option:

```
diff --git a/pvm/src/storage/chunked_io.rs b/pvm/src/storage/chunked_io.rs
index 9d00c1ffff..d76ccc0199 100644
--- a/pvm/src/storage/chunked_io.rs
+++ b/pvm/src/storage/chunked_io.rs
@@ -80,11 +80,25 @@
     }
 }

+struct BytesWrapper<'a>(Box<dyn AsRef<[u8]> + 'a>);
+impl<'a> BytesWrapper<'a> {
+    fn new<T: AsRef<[u8]> + 'a>(value: T) -> Self {
+        Self(Box::new(value))
+    }
+}
+impl<'a> AsRef<[u8]> for BytesWrapper<'a> {
+    fn as_ref(&self) -> &[u8] {
+        use std::ops::Deref;
+
+        self.0.deref().as_ref()
+    }
+}
+
 /// Just like [`ChunkWriter`], but for reading.
 pub struct ChunkedReader<'a, BS> {
     store: &'a BS,
     hashes: VecDeque<Hash>,
-    buffer: Cursor<Vec<u8>>,
+    buffer: Cursor<BytesWrapper<'a>>,
 }

 impl<'a, BS: BlobStore> ChunkedReader<'a, BS> {
@@ -95,7 +109,7 @@
         Ok(Self {
             store,
             hashes,
-            buffer: Cursor::new(Vec::with_capacity(CHUNK_SIZE)),
+            buffer: Cursor::new(BytesWrapper::new([])),
         })
     }

@@ -107,9 +121,7 @@

         let bytes = self.store.blob_get(hash)?;

-        self.buffer.get_mut().clear();
-        self.buffer.get_mut().extend_from_slice(bytes.as_ref());
-        self.buffer.set_position(0);
+        self.buffer = Cursor::new(BytesWrapper::new(bytes));

         Ok(())
     }
@@ -117,7 +129,7 @@

 impl<BS: BlobStore> io::Read for ChunkedReader<'_, BS> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        if self.buffer.position() as usize >= self.buffer.get_ref().len() {
+        if self.buffer.position() as usize >= self.buffer.get_ref().as_ref().len() {
             self.next_chunk().map_err(io::Error::other)?;
         }
```

This doesn't memcpy, but it does do a (very small) allocation.

emturner: I suspect for now, though - it's genuinely probably okay to merge as is. Yes it can _might_ be a slight slowdown - but that for now is acceptable given this code is run infrequently, and is known to not be very performant to begin with.

I suspect it makes sense to get to the more maintainable place sooner rather than later - and then adjust from there once we have actual data to go off of?

pvm/src/storage.rs
```diff
@@ -150,61 +156,115 @@ impl BlobStore for Store {
...
+        for chunk in commit {
+            self.backend.export_blob(&mut other, &chunk)?;
+        }
+        self.backend.export_blob(&mut other, id)?;
```
emturner, line 192: nit (feel free to resolve)

This is already read to memory? So could `blob_set` it at this point.

emturner (approved):

victor-dumitrescu (approved):

pvm/src/storage.rs
thomasathorne: Using `blob_set` here is annoying, we already have the data and its hash but we need to pass them to `blob_set` as a `HashedData`, which we can't construct without recalculating the hash. 🙄

I'm inclined to just leave it as it is now, since that's closer to the old implementation and in theory copying a file should be just as quick as (or maybe quicker than) creating a new file and writing to it (`export_blob` shouldn't be using the default implementation).

PR #994, opened 2026-04-13
feat(pvm-storage): new commit, checkout and export, with end-to-end test

thomasathorne:
Closes TZX-121.
Part of TZX-86.

# What

An end-to-end test that creates a (small) empty PVM, commits it to the PVM storage, checks it out, exports it and checks out the exported copy, checking that the round trip gives the same state. This isn't intended to be an exhaustive test at all, more of a quick check that everything is plumbed together and basically works.

This includes two very simple `Unfoldable` implementations that were missing (probably due to PRs on using `Vector` that merged since the original work on unfold). It also adds three new methods to `Repo`:

- `commit_folded` (analogous to `commit_serialised` for something that is `Foldable<BlobStoreFold<BS>>`),
- `checkout_folded` (analogous to `checkout_serialised` for any type that is `Unfoldable`),
- and `export_snapshot_folded` which uses the 'quick-fix' implementation described in TZX-121. 

# Why

Want to check everything actually works with the new folding approach. And we need to be able to export snapshots, even if it's a bit clunky in this first version.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/data/src/store/unfold.rs
+++ b/data/src/store/unfold.rs
@@ -24,10 +24,10 @@ pub struct BlobStoreUnfold<BS> {
     hash: Hash,
 }
 
-// TODO TZX-86: cfg annotation will be removed as part of wiring up the new PVM commit and checkout
-#[cfg(test)]
 impl<BS> BlobStoreUnfold<BS> {
-    fn new(store: Arc<BS>, hash: Hash) -> Self {
+    /// To construct a source for blob-store unfolding we need a cloneable reference to a store
+    /// together with a `hash` (the root hash of the tree, to start from) in the store.
+    pub fn new(store: Arc<BS>, hash: Hash) -> Self {
         Self { store, hash }
     }
 }
--- a/pvm/src/storage.rs
+++ b/pvm/src/storage.rs
@@ -9,16 +9,22 @@ use std::io;
 use std::io::Write;
 use std::path::Path;
 use std::path::PathBuf;
+use std::sync::Arc;
 
 use bincode::Decode;
 use bincode::Encode;
 use bincode::error::DecodeError;
 use bincode::error::EncodeError;
+use octez_riscv_data::foldable::Foldable;
+use octez_riscv_data::foldable::UnfoldError;
+use octez_riscv_data::foldable::Unfoldable;
 use octez_riscv_data::hash::Hash;
 use octez_riscv_data::hash::HashedData;
 use octez_riscv_data::serialisation;
 use octez_riscv_data::store::BlobStore;
 use octez_riscv_data::store::BlobStoreError;
+use octez_riscv_data::store::fold::BlobStoreFold;
+use octez_riscv_data::store::unfold::BlobStoreUnfold;
 use thiserror::Error;
 
 const CHUNK_SIZE: usize = 4096;
@@ -42,6 +48,9 @@ pub enum StorageError {
 
     #[error("Blob store error")]
     BlobStore(#[from] BlobStoreError),
+
+    #[error("Unfold error")]
+    UnfoldError(#[from] UnfoldError),
 }
 
 /// A subtrait for `BlobStore` to provide extra functionality required by the PVM storage to export
@@ -59,8 +68,8 @@ pub trait PersistentBlobStore: BlobStore {
     /// equivalent to using `blob_get` followed by `blob_set` to copy the blob across, it could in
     /// many impls be more efficient (especially for large blobs) by using `std::fs::copy` or
     /// equivalent instead.
-    fn export_blob(&self, other: &mut Self, hash: &Hash) -> Result<(), StorageError> {
-        let blob = self.blob_get(*hash)?;
+    fn export_blob(&self, other: &mut Self, hash: Hash) -> Result<(), StorageError> {
+        let blob = self.blob_get(hash)?;
         other.blob_set(&HashedData::from_data(blob.as_ref()))?;
 
         Ok(())
@@ -118,9 +127,9 @@ impl PersistentBlobStore for Store {
         })
     }
 
-    fn export_blob(&self, other: &mut Self, hash: &Hash) -> Result<(), StorageError> {
-        let source_path = self.path_of_hash(hash);
-        let target_path = other.path_of_hash(hash);
+    fn export_blob(&self, other: &mut Self, hash: Hash) -> Result<(), StorageError> {
+        let source_path = self.path_of_hash(&hash);
+        let target_path = other.path_of_hash(&hash);
         std::fs::copy(source_path, target_path)?;
         Ok(())
     }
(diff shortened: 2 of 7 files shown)


data/src/store/unfold.rs
```diff
@@ -24,10 +24,8 @@ pub struct BlobStoreUnfold<BS> {
     hash: Hash,
 }
 
-// TODO TZX-86: cfg annotation will be removed as part of wiring up the new PVM commit and checkout
```
emturner, line 27: guessing this PR should also close TZX-86 then?

data/src/store/unfold.rs
```diff
@@ -24,10 +24,8 @@ pub struct BlobStoreUnfold<BS> {
...
-#[cfg(test)]
 impl<BS> BlobStoreUnfold<BS> {
-    fn new(store: Arc<BS>, hash: Hash) -> Self {
+    pub fn new(store: Arc<BS>, hash: Hash) -> Self {
```
emturner, line 30: could you poss add a v short doc string for this now public function?

pvm/src/storage.rs
```diff
@@ -196,21 +213,29 @@ impl<BS: PersistentBlobStore> Repo<BS> {
...
-    pub fn export_snapshot_folded(
+    /// TODO (TZX-126): Improve this so that it doesn't need to deserialise the whole state being
+    /// snapshotted.
+    pub fn export_snapshot_folded<T: Unfoldable + Foldable<BlobStoreFold<BS>>>(
         &self,
-        _id: &Hash,
-        _path: impl AsRef<Path>,
+        id: &Hash,
+        path: impl AsRef<Path>,
     ) -> Result<(), StorageError> {
```
emturner:
```suggestion
    pub fn export_snapshot_folded<State>(
&self,
        id: &Hash,
        path: impl AsRef<Path>,
) -> Result<(), StorageError>
where
    State: Unfoldable + Foldable<BlobStoreFold<BS>> {
```

nit: usually find  `where` clauses a bit more readable once you have more than one constraint going on?

pvm/src/storage.rs
```diff
@@ -196,21 +213,29 @@ impl<BS: PersistentBlobStore> Repo<BS> {
...
-    ///
-    /// TODO (TZX-121)
-    pub fn export_snapshot_folded(
+    /// TODO (TZX-126): Improve this so that it doesn't need to deserialise the whole state being
+    /// snapshotted.
+    pub fn export_snapshot_folded<T: Unfoldable + Foldable<BlobStoreFold<BS>>>(
         &self,
-        _id: &Hash,
-        _path: impl AsRef<Path>,
+        id: &Hash,
```
emturner:
```suggestion
        id: Hash,
```

(feel free to resolve) - given that this requires this by value below, may as well take it by value?

pvm/tests/test_pvm_storage.rs
```diff
@@ -147,6 +152,32 @@ fn test_repo_serialised() {
     snapshot_dir.close().unwrap()
 }
 
+type TestPvm = Pvm<M1M, EmptyPageCache, DurableStorageDummy, Normal>;
+
+#[test]
+fn test_repo_folding() {
+    let tmp_dir = tempfile::tempdir().unwrap();
```
emturner: could you possibly pull in `octez_riscv_test_utils`? That way you can use `TestableTmpDir` which is a useful wrapper over this (e.g. it will keep the tmp_dir around if the test doesn't succeed)

pvm/tests/test_pvm_storage.rs
```diff
@@ -147,6 +152,32 @@ fn test_repo_serialised() {
...
+    let repo = Repo::<TestStore>::load(tmp_dir.path()).unwrap();
+
+    let empty_pvm = TestPvm::default();
+    let hash = repo.commit_folded(&empty_pvm).unwrap();
+    let checked_out = repo.checkout_folded::<TestPvm>(&hash).unwrap();
+
+    assert!(empty_pvm == checked_out);
```
emturner, line 162: nit - possibly worth a small comment on why you use `assert!` rather than `assert_eq!` (and below)

(I assume it's to avoid dumping the entire PVM state out when they don't match?)

data/src/store/unfold.rs
thomasathorne: It's part of that issue (added this to the PR description), but that's the benchmark issue so this PR is preparatory not the final one.

data/src/store/unfold.rs
thomasathorne: Good point, done so [here](https://github.com/tezos/riscv-pvm/compare/717a4248bb23fddf2cd43ff21218390c4563af57..90e58357c3141532887d137d25f5a251550ea9b2)

pvm/src/storage.rs
thomasathorne: Agree it's more readable, done [here](https://github.com/tezos/riscv-pvm/compare/90e58357c3141532887d137d25f5a251550ea9b2..680addec888f4f53324cb6d3138878886c69f51b)

pvm/src/storage.rs
thomasathorne: Switched in a few places [here](https://github.com/tezos/riscv-pvm/compare/680addec888f4f53324cb6d3138878886c69f51b..3c67ea4e75652a315b06f8bf4fb2bc11005cfeb8)

pvm/tests/test_pvm_storage.rs
thomasathorne: Sure, done [here](https://github.com/tezos/riscv-pvm/compare/4893277be14069cbb6a94f4eda3f98777d22b4a5..fa6dc56c206fe9bdd8a99ddd8698fa22e9d12624)

pvm/tests/test_pvm_storage.rs
thomasathorne: Done [here](https://github.com/tezos/riscv-pvm/compare/fa6dc56c206fe9bdd8a99ddd8698fa22e9d12624..ab9befca58cc53d55c2c7a36009b313e8e2b607c)

pvm/tests/test_pvm_storage.rs
```diff
@@ -135,22 +138,43 @@ fn test_repo_serialised() {
...
     let snapshot_repo = Repo::<Store>::load(snapshot_dir.path()).unwrap();
     let checked_out_data: Vec<u8> = snapshot_repo.checkout_serialised(export_hash).unwrap();
     assert_eq!(checked_out_data, *export_data);
+}
+
+type TestPvm = Pvm<M1M, EmptyPageCache, DurableStorageDummy, Normal>;
+
+#[test]
+fn test_repo_folding() {
+    let tmp_dir = tempfile::tempdir().unwrap();
```
emturner: I think this one was missed? (`TestableTmpDir`)

emturner (approved):

pvm/tests/test_pvm_storage.rs
thomasathorne: Sorry about that, had some issues with my `jj` worktree (other changes to this test in a separate branch). 🤦‍♂️ 

Sorted it [here](https://github.com/tezos/riscv-pvm/compare/2e71148a9ec735d17e741ad994258dd6ebf24efe..da3d286e2541348bb811f1c0c27367fd9dcc24fd)

pvm/src/storage.rs
```diff
@@ -157,37 +166,45 @@ impl BlobStore for Store {
...
     }
 
+    /// Makes sure an empty directory exists at the path, suitable for a new `Repo`, and
+    /// initialises a blob store there.
+    ///
+    /// If this is not possible (the directory isn't empty or is a file), we return
+    /// `Err(StorageError::InvalidRepo)`.
+    fn init_empty(path: impl AsRef<Path>) -> Result<BS, StorageError> {
+        let path = path.as_ref();
+        if !path.exists() || path.read_dir()?.next().is_none() {
```
victor-dumitrescu: Contrary to what the doc comment says, if the path is to a file, `path.read_dir()` will return an IO error before getting the chance to return an `InvalidRepo` error.

thomasathorne: Ah 🤦‍♂️ very good point. Fixed (and added a test case) [here](https://github.com/tezos/riscv-pvm/compare/da3d286e2541348bb811f1c0c27367fd9dcc24fd..decdefe72d094c24df51cb66b362cfe05b8387eb)

victor-dumitrescu (approved):

PR #1004, opened 2026-04-16
feat(durable-storage): Fix Prove Resolver

NSant215:
Closes RV-957.

# What and Why

The current implementation for creating a proof of the MerkleLayer is flawed. Due to the inherent ability for the MAVL tree to change structure, we are unable to fold the state below the level of the MerkleLayer in a way that we can match the initial and final state of the proof. 

Instead, we implement the initial-state creation (with MerkleProofFold::proof) using a combination of a frozen `initial_tree`, a track of accesses in the `ProveResolver`, and a `working_tree`. 

See RV-957 for more detail and design. 

# How

- Refactor ProveImpl to hold an initial_tree (immutable snapshot) and a
working_tree (mutable prove-mode projection). The MerkleProofFold is
now driven by walking the initial tree with access-set blinding, reading
per-field flags from the prove-mode nodes in the working tree.

This replaces the old Foldable<MerkleProofFold> impls on ProveNodeId,
ProveTreeId, and Tree<ProveNodeId> which folded the post-mutation tree
and produced incorrect proofs after AVL rotations.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -94,6 +94,11 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
         self.hash = OnceLock::new();
     }
 
+    /// Direct access to the left child identifier.
+    pub(crate) fn left_id(&self) -> &TreeId {
+        &self.left
+    }
+
     /// A mutable reference to the left branch.
     #[inline]
     pub(super) fn left_mut<NodeId>(
@@ -111,6 +116,11 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
         resolver: &impl TreeResolver<NodeId, TreeId>,
     ) -> Result<&Tree<NodeId>, OperationalError> {
         resolver.resolve(&self.left)
+    }
+
+    /// Direct access to the right child identifier.
+    pub(crate) fn right_id(&self) -> &TreeId {
+        &self.right
     }
 
     /// A mutable reference to the right branch.
--- a/durable-storage/src/avl/resolver.rs
+++ b/durable-storage/src/avl/resolver.rs
@@ -48,7 +48,6 @@ use octez_riscv_data::merkle_proof::DeserialiserNode;
 use octez_riscv_data::merkle_proof::FromProof;
 use octez_riscv_data::merkle_proof::Partial;
 use octez_riscv_data::merkle_proof::SuspendedResult;
-use octez_riscv_data::merkle_proof::proof_tree::MerkleProofFold;
 use octez_riscv_data::mode::Mode;
 use octez_riscv_data::mode::Normal;
 use octez_riscv_data::mode::Prove;
@@ -321,6 +320,12 @@ impl From<Hash> for LazyTreeId {
     }
 }
 
+impl From<Tree<LazyNodeId>> for LazyTreeId {
+    fn from(value: Tree<LazyNodeId>) -> Self {
+        Self(LazyId::new(value))
+    }
+}
+
 impl Default for LazyTreeId {
     fn default() -> Self {
         Self(LazyId {
@@ -450,6 +455,43 @@ impl<KV: KeyValueStore> Resolver<LazyTreeId, Tree<LazyNodeId>> for LazyResolver<
     }
 }
 
+/// Resolver that only accesses already-cached prove-mode values.
+///
+/// Returns [`OperationalError::ResolverInvariantViolated`] if a node or tree has not been
+/// resolved yet, since callers are expected to only traverse pre-resolved subtrees.
+pub(crate) struct CachedOnlyResolver;
+
+impl Resolver<ProveNodeId, Node<ProveTreeId, Prove<'static>>> for CachedOnlyResolver {
+    fn resolve<'a>(
+        &self,
+        id: &'a ProveNodeId,
+    ) -> Result<&'a Node<ProveTreeId, Prove<'static>>, OperationalError> {
+        id.cached_node()
+            .ok_or(OperationalError::ResolverInvariantViolated)
+    }
+
+    fn resolve_mut<'a>(
+        &mut self,
+        _id: &'a mut ProveNodeId,
+    ) -> Result<&'a mut Node<ProveTreeId, Prove<'static>>, OperationalError> {
+        Err(OperationalError::ResolverInvariantViolated)
+    }
+}
+
+impl Resolver<ProveTreeId, Tree<ProveNodeId>> for CachedOnlyResolver {
+    fn resolve<'a>(&self, id: &'a ProveTreeId) -> Result<&'a Tree<ProveNodeId>, OperationalError> {
+        id.inner()
+            .ok_or(OperationalError::ResolverInvariantViolated)
+    }
+
+    fn resolve_mut<'a>(
+        &mut self,
+        _id: &'a mut ProveTreeId,
+    ) -> Result<&'a mut Tree<ProveNodeId>, OperationalError> {
+        Err(OperationalError::ResolverInvariantViolated)
+    }
+}
+
 /// Identifier for a node resolved in [`Prove`] mode.
 ///
 /// This wrapper keeps the original [`LazyNodeId`] to allow delegating hash computation and
@@ -499,7 +541,6 @@ impl ProveNodeId {
     ///
     /// Used by the `MerkleLayer`-level fold to extract per-field read flags from a node that
     /// was resolved during the proof step.
-    #[expect(dead_code, reason = "used by MerkleLayer fold in follow-up commit")]
     pub(crate) fn cached_node(&self) -> Option<&Node<ProveTreeId, Prove<'static>>> {
         self.inner.get().map(|rc| rc.as_ref())
     }
@@ -510,15 +551,6 @@ impl Foldable<HashFold> for ProveNodeId {
         match self.inner.get() {
             Some(inner) => *inner.hash(),
             None => self.node.fold(builder),
-        }
-    }
-}
-
-impl Foldable<MerkleProofFold> for ProveNodeId {
-    fn fold(&self, builder: MerkleProofFold) -> <MerkleProofFold as Fold>::Folded {
-        match self.inner.get() {
-            Some(inner) => inner.as_ref().fold(builder),
-            None => builder.into_blind(Hash::from_foldable(&self.node)),
         }
     }
 }
@@ -555,18 +587,8 @@ impl Foldable<HashFold> for ProveTreeId {
 }
 
 impl ProveTreeId {
-    #[expect(dead_code, reason = "used by CachedOnlyResolver in follow-up commit")]
     pub(crate) fn inner(&self) -> Option<&Tree<ProveNodeId>> {
         self.inner.get()
-    }
-}
-
-impl Foldable<MerkleProofFold> for ProveTreeId {
-    fn fold(&self, builder: MerkleProofFold) -> <MerkleProofFold as Fold>::Folded {
-        match self.inner.get() {
-            Some(inner) => inner.fold(builder),
-            None => builder.into_blind(Hash::from_foldable(&self.tree)),
-        }
     }
 }
 
@@ -575,7 +597,6 @@ impl Foldable<MerkleProofFold> for ProveTreeId {
 /// Only `meta` and `data` carry the per-field read flags consumed by the
 /// `MerkleLayer`-level fold; subtree IDs and the hash cache are unnecessary.
 #[derive(Debug)]
-#[expect(dead_code, reason = "used by MerkleLayer fold in follow-up commit")]
 pub(crate) struct DeletedNodeFields {
     pub(crate) meta: Atom<Meta, Prove<'static>>,
     pub(crate) data: Bytes<Prove<'static>>,
@@ -612,29 +633,39 @@ pub(crate) struct ProveResolver<R> {
 
 impl<R> ProveResolver<R> {
     /// Start the prove resolution process.
-    pub(crate) fn start(inner: R) -> Self {
+    pub(crate) fn start(inner: R, root_tree_hash: Option<Hash>) -> Self {
+        let mut accessed_items = AccessedItems::default();
+        if let Some(root_hash) = root_tree_hash {
+            accessed_items.trees.insert(root_hash);
+        }
+
         Self {
             inner,
-            accessed_items: RefCell::default(),
+            accessed_items: RefCell::new(accessed_items),
             deleted_nodes: Rc::new(RefCell::new(BTreeMap::new())),
         }
     }
 
+    /// The wrapped resolver used for [`LazyNodeId`] / [`LazyTreeId`] resolution.
+    ///
+    /// The fold uses this directly to walk the initial tree, bypassing the access-tracking
+    /// methods.
+    pub(crate) fn inner(&self) -> &R {
+        &self.inner
+    }
+
     /// Whether a node with the given hash was accessed during the proof step.
-    #[expect(dead_code, reason = "used by MerkleLayer fold in follow-up commit")]
     pub(crate) fn was_node_accessed(&self, hash: &Hash) -> bool {
         self.accessed_items.borrow().nodes.contains(hash)
     }
 
     /// Whether a tree with the given hash was accessed during the proof step.
-    #[expect(dead_code, reason = "used by MerkleLayer fold in follow-up commit")]
     pub(crate) fn was_tree_accessed(&self, hash: &Hash) -> bool {
         self.accessed_items.borrow().trees.contains(hash)
     }
 
     /// Prove-mode projections of nodes unlinked from the working tree during the step, keyed by
     /// their initial-tree hash.
-    #[expect(dead_code, reason = "used by MerkleLayer fold in follow-up commit")]
     pub(crate) fn deleted_nodes(&self) -> impl Deref<Target = BTreeMap<Hash, DeletedNodeFields>> {
         self.deleted_nodes.borrow()
     }
@@ -1336,7 +1367,7 @@ mod tests {
         let lazy_tree: Tree<LazyNodeId> = Some(LazyNodeId::from(root_hash)).into();
         let lazy_resolver = LazyResolver::new(persistence_layer);
 
-        let prove_resolver = ProveResolver::start(lazy_resolver);
+        let prove_resolver = ProveResolver::start(lazy_resolver, None);
         let lazy_node_id = lazy_tree.root().expect("tree should have a root");
 
         let prove_id = lazy_node_id.clone().into_proof(&prove_resolver);
@@ -1355,7 +1386,7 @@ mod tests {
 
         let expected_hash = Hash::from_foldable(lazy_root);
 
-        let prove_resolver = ProveResolver::start(lazy_resolver);
+        let prove_resolver = ProveResolver::start(lazy_resolver, None);
         let prove_id = lazy_root.clone().into_proof(&prove_resolver);
 
         // Hash before resolve (delegates to lazy path).
@@ -1383,7 +1414,7 @@ mod tests {
         let lazy_tree: Tree<LazyNodeId> = Some(LazyNodeId::from(root_hash)).into();
         let lazy_resolver = LazyResolver::new(persistence_layer);
 
-        let prove_resolver = ProveResolver::start(lazy_resolver);
+        let prove_resolver = ProveResolver::start(lazy_resolver, None);
         let prove_root_id = lazy_tree
             .root()
             .expect("tree should have a root")
@@ -1420,7 +1451,7 @@ mod tests {
         let lazy_tree: Tree<LazyNodeId> = Some(LazyNodeId::from(root_hash)).into();
         let lazy_resolver = LazyResolver::new(persistence_layer.clone());
 
-        let prove_resolver = ProveResolver::start(lazy_resolver);
+        let prove_resolver = ProveResolver::start(lazy_resolver, None);
         let prove_id = lazy_tree
             .root()
             .expect("tree should have a root")
@@ -1448,7 +1479,7 @@ mod tests {
         let lazy_tree: Tree<LazyNodeId> = Some(LazyNodeId::from(root_hash)).into();
         let lazy_resolver = LazyResolver::new(persistence_layer);
 
-        let mut prove_resolver = ProveResolver::start(lazy_resolver);
+        let mut prove_resolver = ProveResolver::start(lazy_resolver, None);
         let mut prove_id = lazy_tree
             .root()
             .expect("tree should have a root")
@@ -1485,7 +1516,7 @@ mod tests {
         let lazy_left_node_hash = Hash::from_foldable(lazy_left_id);
 
         // Now get the same hash via the prove resolver path.
-        let prove_resolver = ProveResolver::start(lazy_resolver);
+        let prove_resolver = ProveResolver::start(lazy_resolver, None);
         let prove_root_id = lazy_root.clone().into_proof(&prove_resolver);
         let prove_node = prove_resolver
             .resolve(&prove_root_id)
@@ -1515,7 +1546,7 @@ mod tests {
         let empty_lazy_tree: LazyTreeId = LazyTreeId::default();
         let prove_tree_id = empty_lazy_tree.clone().into_proof();
 
-        let prove_resolver = ProveResolver::start(lazy_resolver);
+        let prove_resolver = ProveResolver::start(lazy_resolver, None);
         let tree = prove_resolver
             .resolve(&prove_tree_id)
             .expect("resolving empty prove tree should succeed");
--- a/durable-storage/src/avl/tree.rs
+++ b/durable-storage/src/avl/tree.rs
@@ -24,15 +24,17 @@ use octez_riscv_data::merkle_proof::Deserialiser;
 use octez_riscv_data::merkle_proof::DeserialiserNode;
 use octez_riscv_data::merkle_proof::FromProof;
 use octez_riscv_data::merkle_proof::Partial;
-use octez_riscv_data::merkle_proof::proof_tree::MerkleProofFold;
-use octez_riscv_data::merkle_proof::proof_tree::MinimumPresence;
+use octez_riscv_data::mode::Prove;
 use octez_riscv_data::mode::utils::not_found;
 use octez_riscv_data::serialisation::deserialise;
 use octez_riscv_data::serialisation::serialise;
 use perfect_derive::perfect_derive;
 
 use super::node::Node;
+use super::resolver::CachedOnlyResolver;
 use super::resolver::ProveNodeId;
+use super::resolver::ProveTreeId;
+use super::resolver::Resolver;
 use super::resolver::VerifyNodeId;
 use crate::avl::resolver::AvlResolver;
 use crate::avl::resolver::LazyNodeId;
@@ -207,7 +209,7 @@ impl<NodeId> Tree<NodeId> {
 
     #[inline]
     /// A reference to the root [`Node`].
-    pub(super) fn root(&self) -> Option<&NodeId> {
+    pub(crate) fn root(&self) -> Option<&NodeId> {
         self.0.as_ref()
     }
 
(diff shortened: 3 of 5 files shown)


durable-storage/src/avl/node.rs
```diff
@@ -207,6 +217,16 @@ impl<TreeId, M: BytesMode + AtomMode> Node<TreeId, M> {
         &self.meta.key
     }
 
+    /// The metadata of this node.
+    pub(crate) fn meta_atom(&self) -> &Atom<Meta, M> {
```
```suggestion
    pub(crate) fn metadata(&self) -> &Atom<Meta, M> {
```

(ditto with bytes - the type is in the return type)

durable-storage/src/avl/resolver.rs
```diff
@@ -465,6 +493,16 @@ impl From<Node<ProveTreeId, Prove<'static>>> for ProveNodeId {
...
 
+/// Access the cached prove-mode node, if populated.
+///
+/// Used by the `MerkleLayer`-level fold to extract per-field read flags from a node that
+/// was resolved during the proof step.
+impl ProveNodeId {
+    pub(crate) fn cached_node(&self) -> Option<&Node<ProveTreeId, Prove<'static>>> {
+        self.inner.get().map(|rc| rc.as_ref())
+    }
+}
```
```suggestion
impl ProveNodeId {
    /// Access the cached prove-mode node, if populated.
    ///
    /// Used by the `MerkleLayer`-level fold to extract per-field read flags from a node that
    /// was resolved during the proof step.
    pub(crate) fn cached_node(&self) -> Option<&Node<ProveTreeId, Prove<'static>>> {
        self.inner.get().map(|rc| rc.as_ref())
    }
}
```

durable-storage/src/avl/resolver.rs
```diff
@@ -67,6 +69,27 @@ pub trait Resolver<Id, Value> {
...
+///
+/// Hooks must only be triggered at the four semantic deletion sites in [`Tree::delete`]. Rotations
+/// and rebalancing reassign `NodeId` slots without removing any key, so invoking this hook from
+/// rotation paths would produce false-positive "deletions".
+///
+/// [`Tree`]: crate::avl::tree::Tree
+/// [`Tree::delete`]: crate::avl::tree::Tree::delete
+/// [`MerkleProofFold`]: octez_riscv_data::merkle_proof::proof_tree::MerkleProofFold
+/// [`MerkleLayer`]: crate::merkle_layer::MerkleLayer
+pub trait DeletionNotifier<NodeId> {
```
emturner, line 88: I'm fairly convinced that you don't need to use an extra trait as a mechanism - just a `Drop` implementation should be sufficient if the `ProveNodeId` can update `deleted_items` directly?

emturner: Ultimately, there's nothing that would enforce normal mode to actually call the notifier

durable-storage/src/avl/resolver.rs
```diff
@@ -529,12 +560,76 @@ impl Foldable<MerkleProofFold> for ProveTreeId {
...
+
+    /// Inner state for access tracking
+    accessed_items: RefCell<AccessedItems>,
+
+    /// Prove-mode projections of nodes that were unlinked from the working tree during the step.
+    ///
+    /// Keyed by the node's initial hash. The fold consults this map before searching the working
+    /// tree, so that a node which was deleted (or deleted then reinserted at the same key) still
+    /// contributes its correct per-field read flags to the proof.
+    deleted_nodes: RefCell<BTreeMap<Hash, ProveNodeId>>,
```
emturner, line 575:
```suggestion
    deleted_nodes: Rc<RefCell<BTreeMap<Hash, ProveNodeId>>>,
```
Would be sufficient, for example

emturner: You could also do it (perhaps a bit more nicely) using a `Send/Receive` channel - where the resolver holds the receive end. It can then read all the deletions out from the channel before starting the proof fold (perhaps in a `into_proof_fold -> impl Foldable<MerkleProofFold>` method or something?

durable-storage/src/avl/tree.rs
```diff
@@ -347,6 +351,35 @@ impl<NodeId> Tree<NodeId> {
     }
 }
 
+impl<NodeId> Tree<NodeId> {
+    /// Find the `NodeId` for a given key by traversing the tree.
+    ///
+    /// Returns `None` if the key is not present.
+    pub(crate) fn find_node<'a, TreeId, M: BytesMode + AtomMode + 'a>(
```
emturner, line 358: slightly concerned where this will be used

durable-storage/src/merkle_layer.rs
```diff
@@ -379,29 +421,198 @@ impl<KV> NormalImpl<KV> {
...
+///
+/// A proof is generated against the **initial** state that the step read from, not against the
+/// post-step state: AVL rotations on `set`/`delete`/`write` rewrite subtree structure, so folding
+/// the working tree would produce a proof whose shape does not match the initial root. Prove mode
+/// therefore holds both trees:
+///
+/// * `initial_tree` — an immutable snapshot of the Normal-mode tree, captured at `start_proof`
+///   time. This is what the [`MerkleProofFold`] implementation on [`MerkleLayer`] walks.
+/// * `working_tree` — a Prove-mode projection that the step mutates. Its root hash is the
+///   final-state hash. Its per-node [`Cell<bool>`] read flags are the source of truth for
```
emturner, line 440: where did `Cell<bool>` come from? 

Wouldn't all the various parts of the nodes themselves just be in proof mode?

durable-storage/src/merkle_layer.rs
```diff
@@ -379,29 +421,198 @@ impl<KV> NormalImpl<KV> {
...
     resolver: ProveResolver<LazyResolver<KV>>,
 }
 
-#[derive(Debug)]
-struct VerifyImpl {
-    tree: Tree<VerifyNodeId>,
-    resolver: VerifyResolver,
+// ---------------------------------------------------------------------------
+// Fold implementations for prove-mode MerkleLayer
+// ---------------------------------------------------------------------------
```
emturner: these types of comments (I've seen claude generate them) can be removed

durable-storage/src/merkle_layer.rs
```diff
@@ -379,29 +421,198 @@ impl<KV> NormalImpl<KV> {
...
+impl<KV: KeyValueStore> Foldable<HashFold> for MerkleLayer<KV, Prove<'_>> {
+    fn fold(&self, _builder: HashFold) -> <HashFold as Fold>::Folded {
+        self.inner.working_tree.hash()
+    }
+}
+
+impl<KV: KeyValueStore> Foldable<MerkleProofFold> for MerkleLayer<KV, Prove<'_>> {
+    fn fold(&self, builder: MerkleProofFold) -> <MerkleProofFold as Fold>::Folded {
+        // The root tree is always the entry point — it's never a `ProveTreeId` child, so it
+        // won't appear in `accessed_items.trees`. Fold it unconditionally; child subtrees
```
emturner, line 464: shouldn't it be? e.g. the root tree should be blinded in the initial proof, if durable storage is not touched (I think?)

emturner: Alternatively, the root tree could just place itself into `accessed_items` in the resolver automatically? would save having to carry the `is_root` field around iiuc

durable-storage/src/merkle_layer.rs
```diff
@@ -379,29 +421,198 @@ impl<KV> NormalImpl<KV> {
...
+        let deleted = self.prove_impl.resolver.deleted_node(&hash);
+        let prove_node = if let Some(deleted_id) = &deleted {
+            deleted_id
+                .cached_node()
+                .expect("Deleted ProveNodeId should have a cached prove-mode node")
+        } else {
+            let key = initial_node.key();
+            self.prove_impl
+                .working_tree
+                .find_node(key, &self.prove_impl.resolver)
```
emturner, line 558: this technically shouldn't require `resolver` I think? 

If anything get's resolved at this point IIUC it's a bug

durable-storage/src/merkle_layer.rs
```diff
@@ -379,29 +421,198 @@ impl<KV> NormalImpl<KV> {
...
+                .expect("Working-tree lookup should not fail for an accessed node")
+                .expect("Accessed node should still be present in the working tree")
+                .cached_node()
+                .expect("Working-tree ProveNodeId should have a cached prove-mode node")
+        };
+
+        // Fold meta + data from the prove-mode node (they carry per-field read flags).
```
emturner, line 565: oh, I see what this means now - they both carry `read` flags, but more importantly just are in general the proof trackers for these fields

durable-storage/src/merkle_layer.rs
```diff
@@ -379,29 +421,198 @@ impl<KV> NormalImpl<KV> {
...
+///
+/// If the node was not accessed during the step, it is blinded.  Otherwise, per-field presence
+/// (meta, data) is taken from the prove-mode `Node`'s read flags, and the children are folded
+/// recursively off the **initial tree**'s structure (not the working tree's).
+struct InitialNodeFold<'a, KV> {
+    node_id: &'a LazyNodeId,
+    prove_impl: &'a ProveImpl<KV>,
+}
+
+impl<KV: KeyValueStore> Foldable<MerkleProofFold> for InitialNodeFold<'_, KV> {
```
emturner, line 525: would be great to have tests that for proofs that are 'read only' they produce the same hash as the og `HashFold` over the initial tree

durable-storage/src/avl/node.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/6f7b0b4..574aa8de06981a041ad185503ee0b5242ed6dcd7)

durable-storage/src/avl/resolver.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/6f7b0b4..574aa8de06981a041ad185503ee0b5242ed6dcd7)

durable-storage/src/merkle_layer.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/6f7b0b4..7acb8c3)

durable-storage/src/merkle_layer.rs
```diff
@@ -231,10 +231,16 @@ impl MerkleLayerMode for Normal {
 
 impl MerkleLayerMode for Prove<'_> {
     fn try_clone_with<KV: KeyValueStore>(
```
NSant215, line 240: add a TODO here for multi-step - with a corresponding linear issue.

durable-storage/src/merkle_layer.rs
NSant215: changed [here](https://github.com/tezos/riscv-pvm/compare/63328d7b4bdd8f773809119728669aa064f779e1..298cd35b7f18330f39444860770742361043a6c1)

durable-storage/src/avl/tree.rs
NSant215: is this ok now? [diff](https://github.com/tezos/riscv-pvm/compare/63328d7b4bdd8f773809119728669aa064f779e1..298cd35b7f18330f39444860770742361043a6c1)

durable-storage/src/avl/resolver.rs
NSant215: I have done this now in the newest commit

durable-storage/src/avl/resolver.rs
NSant215: done this now in new commit

durable-storage/src/merkle_layer.rs
NSant215: added [here](https://github.com/tezos/riscv-pvm/compare/298cd35b7f18330f39444860770742361043a6c1..87aa50883937f124bcb4e38a03a71cf7e625e8be)

durable-storage/src/merkle_layer.rs
NSant215: maybe clearer [here](https://github.com/tezos/riscv-pvm/compare/87aa50883937f124bcb4e38a03a71cf7e625e8be..6ff9710ec2c42ec031fadd49176f9d87cba419e7)

durable-storage/src/avl/node.rs
```diff
@@ -131,6 +131,16 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
...
+
+    /// Direct access to the raw left child identifier.
+    pub(crate) fn left_id(&self) -> &TreeId {
+        &self.left
+    }
+
+    /// Direct access to the raw right child identifier.
+    pub(crate) fn right_id(&self) -> &TreeId {
+        &self.right
+    }
```
kurtisc: Please order these consistently with the other methods

durable-storage/src/avl/node.rs
```diff
@@ -207,6 +217,16 @@ impl<TreeId, M: BytesMode + AtomMode> Node<TreeId, M> {
         &self.meta.key
     }
 
+    /// The metadata of this node.
+    pub(crate) fn metadata(&self) -> &Atom<Meta, M> {
```
kurtisc: Nit: rename to `meta`

durable-storage/src/avl/node.rs
```diff
@@ -207,6 +217,16 @@ impl<TreeId, M: BytesMode + AtomMode> Node<TreeId, M> {
...
+        &self.meta
+    }
+
+    /// The data stored in this node.
+    pub(crate) fn data(&self) -> &Bytes<M> {
+        &self.data
+    }
+
```
kurtisc: Same ordering comment as above

durable-storage/src/avl/resolver.rs
```diff
@@ -75,7 +78,8 @@ trait_set! {
     pub trait TreeResolver<NodeId, TreeId> = Resolver<TreeId, Tree<NodeId>>;
 
     /// Specialised [`Resolver`] for MAVL nodes and trees
-    pub trait AvlResolver<NodeId, TreeId, M: Mode> = NodeResolver<NodeId, TreeId, M> + TreeResolver<NodeId, TreeId>;
+    pub trait AvlResolver<NodeId, TreeId, M: Mode> =
+        NodeResolver<NodeId, TreeId, M> + TreeResolver<NodeId, TreeId>;
```
kurtisc: Is this just a formatter update?

durable-storage/src/avl/resolver.rs
```diff
@@ -465,6 +469,57 @@ impl From<Node<ProveTreeId, Prove<'static>>> for ProveNodeId {
...
+    pub(crate) fn lazy_node_id(&self) -> &LazyNodeId {
+        &self.node
+    }
+}
+
+/// Dummy resolver that only accesses already-cached prove-mode values.
+///
+/// Returns [`OperationalError::ResolverInvariantViolated`] if a node or tree has not been
+/// resolved yet, since callers are expected to only traverse pre-resolved subtrees.
+pub(crate) struct DummyResolver;
```
kurtisc: I think 'Dummy' downplays the role of this resolver. It's real and used in release builds, it just only used for nodes that have already been resolved/cached.

durable-storage/src/avl/resolver.rs
```diff
@@ -465,6 +469,57 @@ impl From<Node<ProveTreeId, Prove<'static>>> for ProveNodeId {
...
+    /// was resolved during the proof step.
+    pub(crate) fn cached_node(&self) -> Option<&Node<ProveTreeId, Prove<'static>>> {
+        self.inner.get().map(|rc| rc.as_ref())
+    }
+
+    pub(crate) fn lazy_node_id(&self) -> &LazyNodeId {
+        &self.node
+    }
+}
+
```
kurtisc, line 548: You've sandwiched the `DummyResolver` code between this impl block and the 'impl for` blocks. Please reorder them to keep the convention of:

struct X

impl X

impl X for Y

durable-storage/src/avl/resolver.rs
```diff
@@ -505,6 +551,12 @@ impl Default for ProveTreeId {
     }
 }
 
+impl ProveTreeId {
+    pub(crate) fn cached_tree(&self) -> Option<&Tree<ProveNodeId>> {
```
kurtisc: I don't think this name is accurate if it's still optional. It's just a delegated getter.

durable-storage/src/avl/resolver.rs
```diff
@@ -545,11 +674,13 @@ impl<R: Resolver<LazyNodeId, Node<LazyTreeId, Normal>> + Resolver<LazyTreeId, Tr
         &self,
         id: &'b ProveNodeId,
     ) -> Result<&'b Node<ProveTreeId, Prove<'static>>, OperationalError> {
+        self.track_node(&id.node);
```
kurtisc: You have these `track_{node,tree}` methods called manually with Self and the ID. You then later use Self and the ID to resolve the nodes or trees.

Is it correct to say that a node/tree should be tracked as accessed every time it is resolved? Is there then a way to unify the calls?

durable-storage/src/avl/tree.rs
```diff
@@ -87,19 +88,22 @@ impl<NodeId: FromProof> Tree<NodeId> {
 impl<NodeId> Tree<NodeId> {
     /// Delete the [`Node`] in the [`Tree`] with a given key.
     ///
+    /// Each node unlinked from the tree is sent through `deletion_tx`. When no observer is
+    /// needed the receiver can simply be dropped — sends fail silently.
```
kurtisc: You can keep the em-dash only if you can tell me how you type one

durable-storage/src/avl/tree.rs
```diff
@@ -347,6 +363,32 @@ impl<NodeId> Tree<NodeId> {
     }
 }
 
+impl Tree<ProveNodeId> {
+    /// Find the `ProveNodeId` for a given key by traversing the tree.
+    ///
+    /// Returns `None` if the key is not present.
+    ///
+    /// Only traverses already-resolved nodes. Returns an error if an
+    /// unresolved node or tree is encountered.
```
kurtisc, line 359: a) it's only resolved nodes

b) it uses the same logic as lookups that we've called 'get' elsewhere

c) node is evident from the signature

`get_resolved`?

durable-storage/src/merkle_layer.rs
```diff
@@ -301,6 +356,16 @@ struct NormalImpl<KV> {
     tree: Tree<LazyNodeId>,
     persistence: Arc<KV>,
     resolver: LazyResolver<KV>,
+    deletion_tx: mpsc::Sender<LazyNodeId>,
+}
+
+/// Create a sender whose receiver is already dropped.
```
kurtisc: I'm not a huge fan of 'whose' for inanimate intangibles.

```
/// Create a sender with a dropped receiver
```

?

durable-storage/src/merkle_layer.rs
```diff
@@ -379,35 +448,202 @@ impl<KV> NormalImpl<KV> {
...
+#[derive(Debug)]
+struct VerifyImpl {
+    tree: Tree<VerifyNodeId>,
+    resolver: VerifyResolver,
+    deletion_tx: mpsc::Sender<VerifyNodeId>,
+}
+
+/// Prove-mode backing state for a [`MerkleLayer`].
+///
+/// A proof is generated against the **initial** state that the step read from, not against the
```
kurtisc: I think the ** ** distracts more than it adds in most cases this will be read.

durable-storage/src/merkle_layer.rs
```diff
@@ -379,35 +448,202 @@ impl<KV> NormalImpl<KV> {
...
+/// post-step state: AVL rotations on `set`/`delete`/`write` rewrite subtree structure, so folding
+/// the working tree would produce a proof whose shape does not match the initial root. Prove mode
```
kurtisc:
```suggestion
/// the working tree would produce a proof with a shape that does not match the initial root. Prove mode
```

durable-storage/src/merkle_layer.rs
```diff
@@ -379,35 +448,202 @@ impl<KV> NormalImpl<KV> {
...
+/// therefore holds both trees:
+///
+/// * `initial_tree` — an immutable snapshot of the Normal-mode tree, captured at `start_proof`
+///   time. This is what the [`MerkleProofFold`] implementation on [`MerkleLayer`] walks.
+/// * `working_tree` — a Prove-mode projection that the step mutates. Its root hash is the
+///   final-state hash and its per-node read flags are the source of truth for
+///   deciding which fields of an initial node were actually read.
+/// * `resolver` — a [`ProveResolver`] wrapping a [`LazyResolver`]. Its access set tells the fold
+///   which initial-tree nodes can be blinded, and its `deleted_nodes` map preserves the
+///   prove-mode projection of any node that was unlinked from the working tree during the step.
```
kurtisc: This says 'both trees' but then describes each field rather than two trees. This style of bullet point also feel unnatural/distracting.

kurtisc (changes requested):
Not finished reviewing but have a meeting

durable-storage/src/merkle_layer.rs
```diff
@@ -379,33 +451,193 @@ impl<KV> NormalImpl<KV> {
...
+            tree: &self.inner.initial_tree,
+            prove_impl: &self.inner,
+        };
+        wrapper.fold(builder)
+    }
+}
+
+/// Wrapper that folds an initial-tree [`Tree<LazyNodeId>`] into a [`MerkleProofFold`].
+///
+/// Drives the tree-level structure (occupied-bool + optional node) off the initial tree, then
```
kurtisc: I'm not sure what this means

Also, is there a reason this and the `InitialNodeFold` are part of the Merkle layer code? These seem like things that belong in the respective files.

kurtisc (changes requested):

durable-storage/src/merkle_layer.rs
NSant215: messy diff but have made this change now

durable-storage/src/avl/node.rs
NSant215: not quite sure what you mean - is [this](https://github.com/tezos/riscv-pvm/compare/2703226c084027f70163b902d7e04ed6f49b50fa..f0f601ddb4a58b7b8aa9e542b45cefe3789c12fc) what you wanted?

durable-storage/src/avl/node.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/2703226c084027f70163b902d7e04ed6f49b50fa..f0f601ddb4a58b7b8aa9e542b45cefe3789c12fc)

durable-storage/src/avl/node.rs
NSant215: is [this](https://github.com/tezos/riscv-pvm/compare/f0f601ddb4a58b7b8aa9e542b45cefe3789c12fc..7f5cad62af354bd23d04ff8d4ea78104037ee055) what you mean?

durable-storage/src/avl/resolver.rs
NSant215: ah I think claude did this as a pseudo-formatting correction - because it is in a macro it doesn't actually get formatted by cargo fmt - undone now

durable-storage/src/avl/resolver.rs
NSant215: [diff](https://github.com/tezos/riscv-pvm/compare/d9962be6005e588f9f5904de8d2a90c259873897..c93e0b9c19cc5edaf428ea9a34fc37ab4bc7e67a)

durable-storage/src/avl/tree.rs
NSant215: [diff](https://github.com/tezos/riscv-pvm/compare/c93e0b9c19cc5edaf428ea9a34fc37ab4bc7e67a..aa6c4213404de7c0cc722b15679d6a2746e1785b)

durable-storage/src/avl/tree.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/aa6c4213404de7c0cc722b15679d6a2746e1785b..8d6a2e6428356f4bcf42cc39583b583fae8c6f5c)

durable-storage/src/avl/tree.rs
NSant215: curiously I found another that I thought would be good to remove as well [here](https://github.com/tezos/riscv-pvm/compare/8d6a2e6428356f4bcf42cc39583b583fae8c6f5c..e57dc47a9327135ad0bc0c6ed048f1b5dca805b5) 😉

durable-storage/src/merkle_layer.rs
NSant215: diff [here](https://github.com/tezos/riscv-pvm/compare/8d6a2e6428356f4bcf42cc39583b583fae8c6f5c..e57dc47a9327135ad0bc0c6ed048f1b5dca805b5)

durable-storage/src/merkle_layer.rs
NSant215: removed [here](https://github.com/tezos/riscv-pvm/compare/8d6a2e6428356f4bcf42cc39583b583fae8c6f5c..e57dc47a9327135ad0bc0c6ed048f1b5dca805b5)

durable-storage/src/merkle_layer.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/8d6a2e6428356f4bcf42cc39583b583fae8c6f5c..e57dc47a9327135ad0bc0c6ed048f1b5dca805b5)

durable-storage/src/merkle_layer.rs
NSant215: removed [here](https://github.com/tezos/riscv-pvm/compare/8d6a2e6428356f4bcf42cc39583b583fae8c6f5c..e57dc47a9327135ad0bc0c6ed048f1b5dca805b5)

durable-storage/src/avl/resolver.rs
NSant215: changed to CachedOnlyResolver [here](https://github.com/tezos/riscv-pvm/compare/e57dc47a9327135ad0bc0c6ed048f1b5dca805b5..83dea5717e9ac049ba680174dc932da811457372)

durable-storage/src/avl/resolver.rs
NSant215: changed [here](https://github.com/tezos/riscv-pvm/compare/83dea5717e9ac049ba680174dc932da811457372..b1568d8363adb1d394c6e118aff35e5364ab00d7)

durable-storage/src/avl/resolver.rs
NSant215: good point - have unified [here](https://github.com/tezos/riscv-pvm/compare/b1568d8363adb1d394c6e118aff35e5364ab00d7..1da1defc067ddb43f3a620fc53ddd0aa67824f4d)

durable-storage/src/merkle_layer.rs
NSant215: added [here](https://github.com/tezos/riscv-pvm/compare/a28bb39f39270455684813aa13121ab25ba336dd..786e02caa5a53660c7bafdae131fd025d5ee89cf)

durable-storage/src/merkle_layer.rs
NSant215: removed [here](https://github.com/tezos/riscv-pvm/compare/786e02caa5a53660c7bafdae131fd025d5ee89cf..c496a16cdcc4d0f2d79898e5b84b616fa70879bd) - decided it wasn't adding much.

on the latter, it is only associated with making the initial state part of the proof, which is only possible at the merkle layer level. Moving it to tree/node code also involved importing merkle-layer specific structs like ProveImpl, which seemed more odd?

NSant215:
@kurtisc @emturner I have broken out the earlier commits into separate MRs.

NSant215:
@kurtisc @emturner I have split out the deletion mechanism from the huge commit - hopefully that will make reviewing easier

PR #1005, opened 2026-04-17
Implement `Verify` mode for `Database`

kurtisc:
Part of RV-964

# What
Implements `DatabaseMode` for `Verify`

# Why
Part of hooking up `Verify` mode to the `Registry` level

# How
- Promotes `data` and  `get` to non-test functionality
- Passes off each call to the `MerkleLayer` equivalent

`DatabaseMode::hash` is added, but still calls an unimplemented method. I'll change this PR to mark RV-964 as closed when RV-959 is.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/data/src/components/bytes.rs
+++ b/data/src/components/bytes.rs
@@ -214,6 +214,28 @@ impl Bytes<Verify> {
         }
     }
 
+    /// Returns the given range as a contiguous byte slice.
+    ///
+    /// Panics (via [`not_found`]) if the range extends beyond the length, or is not fully
+    /// contained within a single defined entry, indicating the Merkle proof did not include the
+    /// data required for this access.
+    pub fn partial_slice(&self, range: Range<usize>) -> &[u8] {
+        if range.is_empty() {
+            return &[];
+        }
+        if range.start >= self.len() {
+            // SAFETY: called only in `Verify` mode
+            unsafe { not_found() }
+        }
+        match self.bytes.data.contiguous_range(range) {
+            Some(slice) => slice,
+            None => {
+                // SAFETY: called only in `Verify` mode
+                unsafe { not_found() }
+            }
+        }
+    }
+
     /// Zero-initialise all undefined bytes in the given range.
     pub(crate) fn zero_init_range(&mut self, range: Range<usize>) {
         let mut offset = range.start;
--- a/data/src/partial_vec.rs
+++ b/data/src/partial_vec.rs
@@ -366,7 +366,7 @@ impl<T> PartialVec<T> {
     ///
     /// Returns `None` if the range spans multiple entries, includes any undefined data, or is empty
     /// and out-of-bounds.
-    pub fn contiguous_range(&self, range: Range<usize>) -> Option<&[T]> {
+    pub(crate) fn contiguous_range(&self, range: Range<usize>) -> Option<&[T]> {
         let idx = self
             .entries
             .partition_point(|entry| entry.end() < range.end);
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -134,6 +134,36 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
 }
 
 impl<TreeId, M: BytesMode + AtomMode> Node<TreeId, M> {
+    /// The data stored in a [`Node`] within the subtree of this [`Node`] with a given [`Key`] .
+    pub(super) fn get<'a, NodeId>(
+        mut node: &'a NodeId,
+        key: &Key,
+        resolver: &impl AvlResolver<NodeId, TreeId, M>,
+    ) -> Result<Option<&'a Bytes<M>>, OperationalError>
+    where
+        NodeId: 'a,
+        TreeId: 'a,
+    {
+        loop {
+            let resolved_node = resolver.resolve(node)?;
+            match resolved_node.key().cmp(key) {
+                Ordering::Equal => return Ok(Some(&resolved_node.data)),
+                Ordering::Greater => {
+                    let Some(left) = resolved_node.left_ref(resolver)?.root() else {
+                        return Ok(None);
+                    };
+                    node = left;
+                }
+                Ordering::Less => {
+                    let Some(right) = resolved_node.right_ref(resolver)?.root() else {
+                        return Ok(None);
+                    };
+                    node = right;
+                }
+            }
+        }
+    }
+
     /// Create a new leaf [`Node`] from the given key and data.
     pub(crate) fn new(key: Key, data: impl Into<Bytes<M>>) -> Self
     where
@@ -712,36 +742,6 @@ impl<TreeId, M: BytesMode + AtomMode> Node<TreeId, M> {
 
 #[cfg(test)]
 impl<TreeId, M: BytesMode + AtomMode> Node<TreeId, M> {
-    /// The data stored in a [`Node`] within the subtree of this [`Node`] with a given [`Key`] .
-    pub(super) fn get<'a, NodeId>(
-        mut node: &'a NodeId,
-        key: &Key,
-        resolver: &impl AvlResolver<NodeId, TreeId, M>,
-    ) -> Result<Option<&'a Bytes<M>>, OperationalError>
-    where
-        NodeId: 'a,
-        TreeId: 'a,
-    {
-        loop {
-            let resolved_node = resolver.resolve(node)?;
-            match resolved_node.key().cmp(key) {
-                Ordering::Equal => return Ok(Some(resolved_node.data())),
-                Ordering::Greater => {
-                    let Some(left) = resolved_node.left_ref(resolver)?.root() else {
-                        return Ok(None);
-                    };
-                    node = left;
-                }
-                Ordering::Less => {
-                    let Some(right) = resolved_node.right_ref(resolver)?.root() else {
-                        return Ok(None);
-                    };
-                    node = right;
-                }
-            }
-        }
-    }
-
     /// Returns true if the balance factors stored in the [`Node`]'s subtree are correct.
     pub(super) fn has_correct_balance_factors<NodeId>(
         &self,
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -12,6 +12,8 @@ pub(crate) mod value_ref;
 
 use std::convert::Infallible;
 use std::marker::PhantomData;
+use std::ops::Index;
+use std::ops::Range;
 use std::sync::Arc;
 
 use bytes::BufMut;
@@ -22,6 +24,7 @@ use octez_riscv_data::hash::HashFold;
 use octez_riscv_data::mode::Modal;
 use octez_riscv_data::mode::Mode;
 use octez_riscv_data::mode::Normal;
+use octez_riscv_data::mode::Verify;
 use tezos_smart_rollup_constants::core::MAX_FILE_CHUNK_SIZE;
 use tokio::runtime::Handle;
 
@@ -32,6 +35,7 @@ use crate::errors::Error;
 use crate::errors::InvalidArgumentError;
 use crate::errors::OperationalError;
 use crate::key::Key;
+use crate::merkle_layer::MerkleLayer;
 use crate::merkle_worker::BackgroundKeyValueStore;
 use crate::merkle_worker::BackgroundPersistentKeyValueStore;
 use crate::merkle_worker::MerkleWorker;
@@ -289,7 +293,7 @@ impl<KV> Modal for DatabaseTemplate<KV> {
 
     type Prove<'normal> = Infallible;
 
-    type Verify = Infallible;
+    type Verify = VerifyImpl<KV>;
 }
 
 /// Modes that support the operational API exposed by [`Database`].
@@ -380,6 +384,79 @@ struct NormalImpl<KV> {
     merkle: MerkleWorker<KV>,
 }
 
+/// Verify-mode implementation for the [`Database`].
+struct VerifyImpl<KV> {
+    merkle: MerkleLayer<KV, Verify>,
+}
+
+impl DatabaseMode for Verify {
+    fn get<KV: BackgroundKeyValueStore>(
+        this: &Database<KV, Self>,
+        key: &Key,
+    ) -> Result<impl ValueRef, Error> {
+        // Wraps the return type of [`MerkleLayer::get`] to allow returning the data as an
+        // [`impl ValueRef`] without allocating.
+        struct Wrapper<'a>(&'a octez_riscv_data::components::bytes::Bytes<Verify>);
+
+        impl Index<Range<usize>> for Wrapper<'_> {
+            type Output = [u8];
+
+            fn index(&self, range: Range<usize>) -> &[u8] {
+                // TODO RV-993: Reading data which is contained in multiple adjacent entries will
+                // yield `not_found`.
+                self.0.partial_slice(range)
+            }
+        }
+
+        impl ValueRef for Wrapper<'_> {
+            fn len(&self) -> usize {
+                self.0.len()
+            }
+        }
+
+        let value = this
+            .inner
+            .merkle
+            .get(key)?
+            .ok_or(InvalidArgumentError::KeyNotFound)?;
+
+        Ok(Wrapper(value))
+    }
+
+    fn set<KV: BackgroundKeyValueStore>(
+        this: &mut Database<KV, Self>,
+        key: Key,
+        data: Bytes,
+    ) -> Result<(), Error> {
+        this.inner.merkle.set(&key, &data)?;
+        Ok(())
+    }
+
+    fn write<KV: BackgroundKeyValueStore>(
+        this: &mut Database<KV, Self>,
+        key: Key,
+        offset: usize,
+        data: Bytes,
+    ) -> Result<usize, Error> {
+        let written = data.len();
+        this.inner.merkle.write(&key, offset, &data)?;
+        Ok(written)
+    }
+
+    fn delete<KV: BackgroundKeyValueStore>(
+        this: &mut Database<KV, Self>,
+        key: Key,
+    ) -> Result<(), OperationalError> {
+        this.inner.merkle.delete(&key)
+    }
+
+    fn hash<KV: BackgroundKeyValueStore>(
+        this: &Database<KV, Self>,
+    ) -> Result<Hash, OperationalError> {
+        Ok(this.inner.merkle.hash())
+    }
+}
+
 #[cfg(test)]
 impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
     /// Assert that a database contains the expected value for a given key.
@@ -399,6 +476,7 @@ mod tests {
 
     use bytes::Bytes;
     use octez_riscv_data::mode::Normal;
+    use octez_riscv_data::mode::Verify;
     use proptest::prelude::*;
     use proptest::prop_assert_eq;
     use proptest::proptest;
@@ -407,11 +485,13 @@ mod tests {
 
     use super::Database;
     use super::DatabaseMode;
+    use crate::avl::tree::Tree;
     use crate::database::MAX_VALUE_SIZE;
     use crate::errors::Error;
     use crate::errors::InvalidArgumentError;
     use crate::key::KEY_MAX_SIZE;
     use crate::key::Key;
+    use crate::merkle_layer::MerkleLayer;
     use crate::merkle_worker::BackgroundKeyValueStore;
     use crate::merkle_worker::BackgroundPersistentKeyValueStore;
     use crate::storage::TestKeyValueStoreSetup;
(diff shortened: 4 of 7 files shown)


data/src/partial_vec.rs
```diff
@@ -360,6 +360,17 @@ impl<T> PartialVec<T> {
     pub fn is_all_undefined(&self) -> bool {
         self.entries.iter().all(|entry| entry.data.is_empty())
     }
+
+    /// Returns the data as a contiguous slice if it consists of a single entry starting at index 0.
+    ///
+    /// Returns `None` if the data is sparse or does not start at the beginning.
+    pub fn as_contiguous_slice(&self) -> Option<&[T]> {
```
victor-dumitrescu, line 367: This is currently used once in `as_slice`. Unless this might be used again, I'd suggest merging the two, I think it would be a bit clearer.

durable-storage/src/database.rs
```diff
@@ -342,6 +344,69 @@ struct NormalImpl<KV> {
     merkle: MerkleWorker<KV>,
 }
 
+/// Verify-mode implementation for the [`Database`].
+struct VerifyImpl<KV> {
+    merkle: MerkleLayer<KV, Verify>,
+}
+
+impl DatabaseMode for Verify {
+    fn get<KV: BackgroundKeyValueStore>(
```
victor-dumitrescu, line 372: I'm not sure this function will work for verification. Say the proof contains a partial value at a given key. Reading that portion of the value should work via `read_bytes`. However, independent of mode, `read_bytes` calls `get`, which in `Verify` mode tries to read the whole value and fails if it can't.

durable-storage/src/database.rs
```diff
@@ -355,15 +354,22 @@ impl DatabaseMode for Verify {
...
-        let bytes: &ModeBytes<Verify> = this
-            .inner
-            .merkle
-            .get(key)?
-            .ok_or(InvalidArgumentError::KeyNotFound)?;
-        let len = bytes.len();
-        let mut buf = vec![0u8; len];
-        bytes.read(0, &mut buf);
-        Ok(buf)
+        // Wraps the return type of `[MerkleLayer::get]` to allow returning the data as
```
NSant215, line 376: is there a reason to separate this implementation in the commmit?

durable-storage/src/database.rs
emturner: I suspect it would be prudent to move away from `impl<AsRef<u8>>` directly? This is where the requirement to have contiguous slices comes from. 

Possibly a trait that gives more control over such as `impl Index<Range, Output = [u8]>` would be preferable?

durable-storage/src/avl/node.rs
```diff
@@ -134,6 +140,35 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
...
+                }
+                Ordering::Less => {
+                    let Some(right) = resolved_node.right_ref(resolver)?.root() else {
+                        return Ok(None);
+                    };
+                    node = right;
+                }
+            }
+        }
+    }
```
NSant215, line 171: think this needs an empty line after it for format to pass

durable-storage/src/database.rs
kurtisc: I don't think I understand the problem here:

>read_bytes calls get, which in Verify mode tries to read the whole value 

My understanding is that using the `AsRef<[u8]>` return type means that there is no reading of the underlying `Bytes` type until the part in the implementation of `read_bytes` that calls `as_ref`

durable-storage/src/avl/node.rs
kurtisc: Each commit should be passing formatting now

data/src/partial_vec.rs
kurtisc: But they're impls on different types?

durable-storage/src/database.rs
kurtisc: My thought process was that I would get it working sub-optimally (with a clone) then implement 1) the new functionality for `PartialVec`/`Bytes` and 2) avoiding the clone as separate concerns

durable-storage/src/database.rs
emturner: It should not be the case that `get` requires the whole of the value to be available, starting from offset zero, in order to implement `read` in proof mode

Otherwise the size of the proof will be unbounded: as it will contain the whole of the value

emturner (changes requested):
I suspect the `as_contiguous_slice` mechanism won't work in Verify mode as it requires the entire value to be present in the proof. 

We can discuss in the meeting tomorrow if needed 👍🏻

kurtisc (commented):
Sorry, some of these comments didn't post yesterday

data/src/partial_vec.rs
kurtisc: Changed slightly with the move to ValueRef

durable-storage/src/database.rs
kurtisc: Implemented using `ValueRef` as discussed in person

durable-storage/src/database.rs
```diff
@@ -375,6 +378,40 @@ impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
     }
 }
 
+/// Borrowed view of a value that can be sub-sliced by range.
+pub trait ValueRef: Index<Range<usize>, Output = [u8]> {
```
emturner: this shouldn't need to be `pub` outside of this module - but it needs to (ig) because of `DatabaseMode`.

Could you add a `TODO` on [RV-986](https://linear.app/tezos/issue/RV-986/modal-components-leak-private-interface) to make it private to the crate

durable-storage/src/database.rs
```diff
@@ -375,6 +378,40 @@ impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
...
+    }
+}
+
+impl<T: AsRef<[u8]>> ValueRef for AsRefValueRef<T> {
+    fn len(&self) -> usize {
+        self.0.as_ref().len()
+    }
+}
+
+impl ValueRef for octez_riscv_data::components::bytes::Bytes<Normal> {
```
emturner: given that this needs to be implemented on foreign types (I imagine it will for `PartialVec` in a subsequent commit) - could you move this into a module under `database`? thanks!

durable-storage/src/avl/node.rs
```diff
@@ -89,6 +89,12 @@ impl Node<LazyTreeId, Normal> {
 }
 
 impl<TreeId, M: Mode> Node<TreeId, M> {
+    #[inline]
+    /// The data stored in the [`Node`].
```
```suggestion
    /// The data stored in the [`Node`].
    #[inline]
```

nit: normally the attributes come after the doc comment - e.g. see [std::vec::Vec](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#438)

durable-storage/src/avl/node.rs
```diff
@@ -89,6 +89,12 @@ impl Node<LazyTreeId, Normal> {
...
+    pub(crate) fn data(&self) -> &Bytes<M> {
```
emturner: nit - does `data` need to be non-test? 

(making it non-test will cause issues with the stuff I'm working on w.r.t. the race condition)

Instead, in `get`, you could just return `&resolved_node.data` instead of calling the method 🙏🏻

durable-storage/src/database.rs
```diff
@@ -364,6 +366,54 @@ struct NormalImpl<KV> {
...
+        this.inner.merkle.write(&key, offset, &data)?;
+        Ok(written)
+    }
+
+    fn delete<KV: BackgroundKeyValueStore>(
+        this: &mut Database<KV, Self>,
+        key: Key,
+    ) -> Result<(), OperationalError> {
+        this.inner.merkle.delete(&key)?;
+        Ok(())
```
emturner:
```suggestion
        this.inner.merkle.delete(&key)
```
`delete` already returns the correct result type

durable-storage/src/database.rs
```diff
@@ -1191,4 +1252,82 @@ mod tests {
...
+        let mut database = new_verify_database();
+        let key = Key::new(&[1]).expect("Size less than KEY_MAX_SIZE");
+
+        database
+            .set(key.clone(), Bytes::copy_from_slice(&[]))
+            .expect("Setting should succeed");
+        assert!(
+            database
+                .exists(&key)
+                .expect("There should be no other `PersistenceLayerError`s")
```
emturner:
```
              .expect("Verify mode does not return `OperationError`s")
```
perhaps? might be slightly clearer possibly

durable-storage/src/database.rs
```diff
@@ -1191,4 +1252,82 @@ mod tests {
             ))
         ));
     }
+
+    #[test]
+    fn test_verify_database_delete() {
+        let mut database = new_verify_database();
+        let key = Key::new(&[1]).expect("Size less than KEY_MAX_SIZE");
```
emturner, line 1347: could also check deleting a non-existent value is also ok?

durable-storage/src/database.rs
```diff
@@ -1191,4 +1252,82 @@ mod tests {
...
+            .expect("Deleting should succeed");
+        assert!(
+            !database
+                .exists(&key)
+                .expect("There should be no other `PersistenceLayerError`s")
+        );
+    }
+
+    #[test]
+    fn test_verify_database_set_and_read() {
```
emturner: this could be a `proptest!` for different lengths

emturner: (and elsewhere)

durable-storage/src/database.rs
```diff
@@ -364,6 +366,54 @@ struct NormalImpl<KV> {
...
+struct VerifyImpl<KV> {
+    merkle: MerkleLayer<KV, Verify>,
+}
+
+impl DatabaseMode for Verify {
+    fn get<KV: BackgroundKeyValueStore>(
+        this: &Database<KV, Self>,
+        key: &Key,
+    ) -> Result<impl AsRef<[u8]>, Error> {
+        unimplemented!()
```
emturner: this is possibly a commit-ordering issue? because this is unimplemented, most of the tests below will fail on this commit

data/src/partial_vec.rs
```diff
@@ -360,6 +360,23 @@ impl<T> PartialVec<T> {
     pub fn is_all_undefined(&self) -> bool {
         self.entries.iter().all(|entry| entry.data.is_empty())
     }
+
+    /// Returns the given range as a contiguous slice if it is fully contained in a single
+    /// defined entry.
+    ///
+    /// Returns `None` if the range spans multiple entries or includes any undefined data.
+    pub fn contiguous_range(&self, range: Range<usize>) -> Option<&[T]> {
```
emturner: could this be `pub(crate)` instead? 

I'd also argue this should return `None` if `range.start > self.len()` - matching the behaviour of `[T]`, even if the index-range is empty

durable-storage/src/database.rs
```diff
@@ -306,15 +309,15 @@ pub trait DatabaseMode: Mode {
...
 }
 
 impl DatabaseMode for Normal {
     fn get<KV: BackgroundKeyValueStore>(
         this: &Database<KV, Self>,
         key: &Key,
-    ) -> Result<impl AsRef<[u8]>, Error> {
-        this.inner.persistent.get(key.as_ref())
+    ) -> Result<impl ValueRef, Error> {
+        Ok(AsRefValueRef(this.inner.persistent.get(key.as_ref())?))
```
emturner: could you possibly break this line up slightly?

data/src/components/bytes.rs
```diff
@@ -214,6 +214,25 @@ impl Bytes<Verify> {
         }
     }
 
+    /// Returns the given range as a contiguous byte slice.
+    ///
+    /// Panics (via [`not_found`]) if the range extends beyond the length, or is not fully
+    /// contained within a single defined entry, indicating the Merkle proof did not include the
+    /// data required for this access.
+    pub fn slice(&self, range: Range<usize>) -> &[u8] {
```
emturner:
```suggestion
    pub fn partial_slice(&self, range: Range<usize>) -> &[u8] {
```
perhaps? (just to make it v clear)

data/src/partial_vec.rs
```diff
@@ -360,6 +360,23 @@ impl<T> PartialVec<T> {
     pub fn is_all_undefined(&self) -> bool {
         self.entries.iter().all(|entry| entry.data.is_empty())
     }
+
+    /// Returns the given range as a contiguous slice if it is fully contained in a single
+    /// defined entry.
+    ///
+    /// Returns `None` if the range spans multiple entries or includes any undefined data.
```
emturner: could you possibly break this out into its own PR? 

Reasoning: it would be very useful to have this method tested properly with some proptests. Namely - we need to be confident that the following property holds:

- Bytes<Proof>: perform a sequence of writes within a range. Initially, these writes might not be overlapping, but by the end, the full range is covered
- ToProof -> bytes -> FromProof -> Bytes<Verify> roundtrip
- `contiguous_slice` over that full range succeeds

It would be okay to have an initial PR that adds a proptest that only covers single write -> contiguous range of the range that that write covered (or any sub-part of that range), and then add the full version as a follow up? 

The initial PR could be merged very quickly I think

And then, in this PR, would allow the `MerkleLayer<Verify>` to just be implemented in one go

kurtisc: Moved to #1023

kurtisc: I will break out `ValueRef` too

kurtisc:
Moved out of draft

durable-storage/src/database.rs
kurtisc: Resolving as done in another PR

durable-storage/src/avl/node.rs
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/1801bb0fabbbc59be46004eaead5282c9d4c3bc9..f1f9de6390091a8ccd0b6e81dd554020fd258b5d)

durable-storage/src/database.rs
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/f1f9de6390091a8ccd0b6e81dd554020fd258b5d..5491b3e28b4b7e6404c3468037eef5c5d4f11aa8)

durable-storage/src/database.rs
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/5491b3e28b4b7e6404c3468037eef5c5d4f11aa8..ea281413011a79cd4a2580abcdfc869a53145059)

durable-storage/src/database.rs
kurtisc: Resolving as no longer a distinct commit

durable-storage/src/database.rs
kurtisc: Resolving as done in another PR

data/src/partial_vec.rs
kurtisc: [New commit](https://github.com/tezos/riscv-pvm/pull/1005/changes/47cb2ea7fe8398bf1b0c9b10b6c63f7b5152fbc1), should have been done in another PR

data/src/components/bytes.rs
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/5e4031846802a018e6d1f2030eac56a4f71e4e9b..e1b2ff3211d5c0dda05e01cee87c4d10f6bd87a4)

durable-storage/src/database.rs
kurtisc: [Done](https://linear.app/tezos/issue/RV-992/make-valueref-private)

durable-storage/src/avl/node.rs
kurtisc: Done but munged diff

durable-storage/src/database.rs
kurtisc: Done but munged diff

durable-storage/src/database.rs
kurtisc: Done but munged diff

data/src/components/bytes.rs
```diff
@@ -214,6 +214,28 @@ impl Bytes<Verify> {
         }
     }
 
+    /// Returns the given range as a contiguous byte slice.
+    ///
+    /// Panics (via [`not_found`]) if the range extends beyond the length, or is not fully
+    /// contained within a single defined entry, indicating the Merkle proof did not include the
+    /// data required for this access.
+    pub fn partial_slice(&self, range: Range<usize>) -> &[u8] {
```
emturner, line 222: could be worth a couple of unit tests here just to uphold this?

durable-storage/src/database.rs
```diff
@@ -380,6 +384,76 @@ struct NormalImpl<KV> {
...
+                self.0.len()
+            }
+        }
+
+        Ok(Wrapper(
+            this.inner
+                .merkle
+                .get(key)?
+                .ok_or(InvalidArgumentError::KeyNotFound)?,
+        ))
```
emturner:
```suggestion
        let value = this.inner
                .merkle
                .get(key)?
                .ok_or(InvalidArgumentError::KeyNotFound)?;
                
        Ok(Wrapper(value))
```

nit: formatting/breaking up the expression

emturner (dismissed):

durable-storage/src/database.rs
```diff
@@ -380,6 +384,76 @@ struct NormalImpl<KV> {
...
+        key: &Key,
+    ) -> Result<impl ValueRef, Error> {
+        // Wraps the return type of [`MerkleLayer::get`] to allow returning the data as an
+        // [`impl ValueRef`] without allocating.
+        struct Wrapper<'a>(&'a octez_riscv_data::components::bytes::Bytes<Verify>);
+
+        impl Index<Range<usize>> for Wrapper<'_> {
+            type Output = [u8];
+
+            fn index(&self, range: Range<usize>) -> &[u8] {
```
victor-dumitrescu, line 404: I'd maybe add a comment about the limitations of `partial_slice` here, namely that reading data which is present but in multiple adjacent entries will yield `not_found`.

durable-storage/src/database.rs
```diff
@@ -1253,4 +1339,108 @@ mod tests {
         );
         assert_eq!(MAX_VALUE_SIZE, database.value_length(&key).unwrap());
     });
+
+    #[test]
+    fn test_verify_database_delete() {
```
victor-dumitrescu: These should be `kv_test!`s now 🫣

kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/6eda31ef3ae2b661692ac011662b0a7ea52eb390..a21593aaed3f1dfb1fb3b93ca56a1efb88ae1181)

durable-storage/src/database.rs
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/a21593aaed3f1dfb1fb3b93ca56a1efb88ae1181..1745d466b5c083bbcac7fbe2170e2c941af8e8fa) and ticket created

durable-storage/src/database.rs
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/1745d466b5c083bbcac7fbe2170e2c941af8e8fa..13dc6a1eb1f47fe65250644f2f8055825b7ec4f6)

victor-dumitrescu (approved):

data/src/components/bytes.rs
kurtisc: Some tests [here](https://github.com/tezos/riscv-pvm/compare/13dc6a1eb1f47fe65250644f2f8055825b7ec4f6..11d5da556cd257b0b3e40a58526f2f669de36109)

data/src/components/bytes.rs
```diff
@@ -214,6 +214,28 @@ impl Bytes<Verify> {
...
+        if range.end > self.len() {
```
emturner:
```suggestion
        if range.start >= self.len() {
```
I think? (only realised because of the tests showed up a case I wasn't expecting 😅 )

If the `range.end` is out of scope - then it's fine, it will be picked up by `contiguous_range` as an error case

kurtisc: [Done ](https://github.com/tezos/riscv-pvm/compare/3615e3e8b7ce9d26896078b1b806c7342398c3ac..86e4f3cc596d307209bf1cda7dfc561a7217e0f6)

emturner: you've pushed `start > self.len()` not `start >= self.len()`

kurtisc: [>=](https://github.com/tezos/riscv-pvm/compare/86e4f3cc596d307209bf1cda7dfc561a7217e0f6..a72ff26e83c995c1bb67a02a783332b3a311f0e1)

emturner (approved):

PR #1007, opened 2026-04-17
test(pvm-storage): benchmark commit and checkout of PVM

thomasathorne:
Closes TZX-86

# What

Add a benchmark to the `pvm` crate which commits and checks out a few PVM states in different ways.

The benchmark is run for the following 24 combinations: (rocks-db/no rocks-db) x (serialising/folding) x (empty PVM/randomised memory PVM) x (`M4K`/`M1M`/`M8M`). For each of the twelve PVM + serialising/folding combinations we also print statistics of bytes/blobs used by the store (including how many 'overlapping' bytes/blobs there were).

# Why

We want to see how the new 'folding' approach compares to the old 'serialising' approach, and how using rocks DB compares to not using it. We also want the benchmark so that we have something to aim against when implementing improvements in the [next project](https://linear.app/tezos/project/production-ready-rollup-node-part-ii-50952c55a053/overview).

A brief outline of what I've observed from running the benchmark locally: 

- folding is considerably slower than serialising,
- using rocks DB is generally a win,
- rocks DB is an even bigger improvement for the folding approach, and for larger PVMs.

# Manually Testing

Run the benchmark with

```
cargo bench --bench pvm_storage --features="rocksdb"
```

# Tasks for the Author


pvm/benches/pvm_storage.rs
```diff
@@ -0,0 +1,362 @@
+// SPDX-FileCopyrightText: 2025 TriliTech <contact@trili.tech>
```
kurtisc: 2026

pvm/benches/pvm_storage.rs
```diff
@@ -0,0 +1,362 @@
...
+
+impl CountingBlobStore {
+    /// Produces a summary of the statistics for the current state of the blob store.
+    fn totals(&self) -> TotalStats {
+        let store = self.blob_stats.lock().expect("Should not be poisoned");
+
+        store.values().map(TotalStats::from).sum()
+    }
+}
+
```
kurtisc: Conventionally we usually put `impl X` before the `impl X for Y` blocks

pvm/benches/pvm_storage.rs
thomasathorne: whoops, see [here](https://github.com/tezos/riscv-pvm/compare/5ca0c903757a4bea9ffb06bf2eee0ede4b7db6a6..83d4ea2ae16cc4cd904de6050cf9bd19c124d673)

pvm/benches/pvm_storage.rs
thomasathorne: Thanks, moved it [here](https://github.com/tezos/riscv-pvm/compare/83d4ea2ae16cc4cd904de6050cf9bd19c124d673..330b067176e0f94fe9525b0d31b83b34d58ff9e0)

pvm/benches/pvm_storage.rs
```diff
@@ -0,0 +1,362 @@
...
+fn setup_random_pvm<MC: MemoryConfig>(rng: &mut impl Rng) -> BenchPvm<MC> {
+    let mut pvm = BenchPvm::<MC>::default();
+
+    let length = MC::TOTAL_BYTES.get();
+    let mut v: Vec<u8> = Vec::with_capacity(length);
+    unsafe {
+        rng.fill(std::slice::from_raw_parts_mut(v.as_mut_ptr(), length));
+        v.set_len(length);
+    }
```
victor-dumitrescu, line 331:
```suggestion
    let mut v = vec![0u8; length];
    rng.fill(v.as_mut_slice());
```

Why use `unsafe` here?

pvm/benches/pvm_storage.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    pvm.machine_state
+        .core
+        .main_memory
+        .write_all(0, &v)
+        .expect("Should be able to write to memory");
+    pvm
+}
+
+fn bench_commit_and_checkout(c: &mut Criterion) {
```
victor-dumitrescu, line 352: Wouldn't it make sense to bench committing and checkout separately? I imagine their variations in performance would be correlated but perhaps not?

pvm/benches/pvm_storage.rs
thomasathorne: Doesn't the `vec![0u8; length]` followed by `fill` write every byte twice, first as 0 and then to a random value? While the unsafe version only writes once, with the random value. That's my understanding anyway.

The performance benefit probably isn't worth it for the memory config sizes in this PR, but I want to run this locally with much bigger PVMs. As I type this my laptop is in an epic struggle with an `M32G` PVM, `htop` has swap memory at 37.7G... 😅

I suppose the performance win on initialising the random PVM is small compared to the cost of the benchmark itself, but also the `unsafe` is a very simple case where it's quite easy to see that it's ok. 🤷‍♂️

pvm/benches/pvm_storage.rs
thomasathorne: Yes this could definitely be a follow-up improvement. I guess at the moment since we're already testing 4 different combinations of folding/serialising and rocksdb/no rocksdb I wanted to keep the total number of different benchmarks down, each one takes around 30 seconds so it already takes 12 minutes for a full run.

Added a note on TZX-129.

pvm/benches/pvm_storage.rs
victor-dumitrescu: Fair point. I was curious if this does make a real difference and at least on my machine over a few runs it doesn't really seem like it, especially compared to the cost of running the benchmark itself:

```
fill/8 MiB/unsafe/8388608
                        time:   [1.4778 ms 1.4848 ms 1.4927 ms]
fill/8 MiB/safe/8388608
                        time:   [1.5998 ms 1.6177 ms 1.6350 ms]

fill/32 MiB/unsafe/33554432
                        time:   [4.7321 ms 4.7455 ms 4.7601 ms]
fill/32 MiB/safe/33554432
                        time:   [5.2043 ms 5.2337 ms 5.2640 ms]

fill/128 MiB/unsafe/134217728
                        time:   [25.646 ms 25.958 ms 26.281 ms]
fill/128 MiB/safe/134217728
                        time:   [25.453 ms 25.754 ms 26.077 ms]

fill/512 MiB/unsafe/536870912
                        time:   [110.43 ms 111.96 ms 113.50 ms]
fill/512 MiB/safe/536870912
                        time:   [118.22 ms 119.77 ms 121.40 ms]

fill/1 GiB/unsafe/1073741824
                        time:   [243.59 ms 244.87 ms 246.15 ms]
fill/1 GiB/safe/1073741824
                        time:   [237.62 ms 238.99 ms 240.43 ms] 🤔

fill/2 GiB/unsafe/2147483648:
                        time:   [485.16 ms 487.47 ms 489.82 ms]
fill/2 GiB/safe/2147483648
                        time:   [492.07 ms 494.51 ms 497.08 ms]
```

I agree this is a really straightforward use of `unsafe` though, especially in a benchmark. I'd add a comment explaining this though. Feel free to resolve.

thomasathorne: Added comment [here](https://github.com/tezos/riscv-pvm/compare/a99acea31f70ec83ba0f45ab2cba34ffada3287e..5825d814c402287a2596eafb949eef5a7cf83ee8)

victor-dumitrescu (approved):
Needs rebasing on main, otherwise LGTM!

kurtisc (approved):

PR #1009, opened 2026-04-20
test(dursto): run MerkleLayer tests across backends using new macro when `rocksdb` is on

victor-dumitrescu:
Closes RV-970. Closes RV-973.

# What

This PR introduces a new `kv_test!` macro in the durable storage test suite, which, given a test function:
- runs it using the in-memory backend when `rocksdb` is off
- runs 2 tests, one using the in-memory backend and one using the persistence layer when `rocksdb` is on

As a first step, only `MerkleLayer` tests are converted to using `kv_test`.

Small caveat: While this refactoring is ongoing and not all tests which can be refactored to use `kv_test`  have been refactored, there will be a slight duplication of testing between running tests with `--no-default-features` and without. Once the refactor is complete, we can stop running tests with `--no-default-features` in CI, since those will be a strict subtset of tests ran without it.

# Why

We want to be to run tests over multiple configurations simultaneously in order to be able to check they are consistent with each other. This is a first step in that direction.

# How

In addition to the `kv_test!` macro, a new `TestKeyValueStoreSetup` trait is introduced which, for now, duplicates the functionality of the `TestKeyValueStore` selection mechanism in order to allow testing with multiple backends at once.

The tests themselves are unchanged as much as possible, except for what is required to wrap them with `kv_test`. The way repos are set up has also slightly changed, now going through a `TestKeyValueStoreSetup` function.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/storage.rs
+++ b/durable-storage/src/storage.rs
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: 2026 Trilitech <contact@trili.tech>
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
 //
 // SPDX-License-Identifier: MIT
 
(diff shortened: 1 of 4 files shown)


durable-storage/src/storage.rs
```diff
@@ -148,6 +149,67 @@ cfg_if::cfg_if! {
     }
 }
 
+// TODO: This trait currently duplicates the functionality of the `TestKeyValueStore` mechanism defined above.
```
emturner: you might be able to put all of this in a `cfg_if` block? then you can avoid having to put `cfg(test)` on each item individually

durable-storage/src/storage.rs
```diff
@@ -148,6 +149,67 @@ cfg_if::cfg_if! {
...
+}
+
+/// Macro that runs a test against every available KV backend.
+///
+/// Without the `rocksdb` feature this expands to a single `#[test]` using the in-memory store.
+/// With the `rocksdb` feature it expands to 2 tests, one using the in-memory store and one
+/// using the persistence layer.
+#[cfg(test)]
+macro_rules! kv_test {
+    (fn $name:ident<$KV:ident: $bound:path>() $body:block) => {
```
emturner: could you do a similar thing to `mode_test` where we bind the `KV` as a separate argument, so that the tests end up as:

```
kv_test!(test_etc, KV, {
  ...
})
```

This would help keep our test macros consistent, and possibly reduce boilerplate very slightly when writing these tests (they don't need to manually have `fn X<...>` correctly)

durable-storage/src/storage.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/66ae912189fdf325b673df94784fa193717492ac..141139f66be9c24419abd022d026603a839d0116).

durable-storage/src/storage.rs
victor-dumitrescu: Good point, done [here](https://github.com/tezos/riscv-pvm/compare/141139f66be9c24419abd022d026603a839d0116..5120e1353ca7ee726721b5f8fe0c1225f1114625).

thomasathorne (approved):

durable-storage/src/storage.rs
```diff
@@ -148,6 +149,70 @@ cfg_if::cfg_if! {
...
+
+                    #[cfg(feature = "rocksdb")]
+                    #[test]
+                    fn [<$fun_name _rocksdb>]() {
+                        fn inner<$ty_name: $crate::storage::TestKeyValueStoreSetup $(+ $ty_bound)?>() $body
+                        inner::<$crate::persistence_layer::PersistenceLayer>();
+                    }
+                }
+            };
+        }
```
kurtisc, line 212:
```suggestion
        macro_rules! kv_test {
            (
                $ty_name:ident $(: [$($bound:tt)+])?,
                $(#[$attr:meta])*
                fn $fun_name:ident() $body:block
            ) => {
                paste::paste! {
                    $(#[$attr])*
                    #[test]
                    fn [<$fun_name _in_memory>]() {
                        fn inner<$ty_name: $crate::storage::TestKeyValueStoreSetup $(+ $($bound)+)?>() $body
                        inner::<$crate::storage::in_memory::InMemoryKeyValueStore>();
                    }

                    $(#[$attr])*
                    #[cfg(feature = "rocksdb")]
                    #[test]
                    fn [<$fun_name _rocksdb>]() {
                        fn inner<$ty_name: $crate::storage::TestKeyValueStoreSetup $(+ $($bound)+)?>() $body
                        inner::<$crate::persistence_layer::PersistenceLayer>();
                    }
                }
            };
        }

```

 - can use attributes like `#[ignore]`
 - can expand the bounds to include lifetimes and multiple type bounds (but unfortunately wrapped in '[]')
 - evocative of output

example use:
```
    kv_test2!(KV: ['a + 'b + crate::storage::TestKeyValueStoreSetup],
        #[ignore]
        fn test_fun2() {
            let keys = [
                Key::new(&[2]),
                Key::new(&[4]),
                Key::new(&[0]),
                Key::new(&[6]),
                Key::new(&[5]),
                Key::new(&[1]),
                Key::new(&[3]),
            ]
            .map(|r| r.expect("Sizes less than KEY_MAX_SIZE"));
            test_mavl_delete_keys::<KV>(&repo, &keys);
        }
    );

```

Design principles taken from [here](https://rust-lang.github.io/api-guidelines/macros.html#macros), which doesn't apply to us but does provide a useful base, IME

kurtisc: I noticed the bounds aren't currently used, so maybe they can be removed, but I guess you have unpublished work that uses them

victor-dumitrescu: Thanks for the suggestion, regarding each point:
- I've added attributes [here](https://github.com/tezos/riscv-pvm/compare/dba7dc5bbfd79ccab823afa3a16fcb9ad684d156..ff7a21a93bc73e5a60ebbb5fa0a79fbd6ac65475).
- Since this isn't (yet) used with multiple bounds, I'd rather leave it as it is to avoid wrapping single bounds in `[]`, but obviously could revisit if it becomes necessary.
- My initial version was more evocative of the output, wrapping a whole test function as in your suggestion. I revised it following [this suggestion](https://github.com/tezos/riscv-pvm/pull/1009#discussion_r3111399487). Beyond consistency with `mode_test`, I think I prefer this snappier invocation.

emturner (approved):

PR #1015, opened 2026-04-22
feat(durable): limit value sizes to 64MiB

emturner:
- Part of RV-960

# What

Limits sizes of values in durable storage to 64MiB

# Why

Large enough to allow all reasonable kernels to be stored in a single slot (e.g. jstz is 18M) - but small enough to not eat up all memory in one go/cause issues with rocksdb.

# How

- `set` is covered as `MAX_FILE_CHUNK_SIZE < MAX_VALUE_SIZE`
- `write`: ensure that `offset + data.len() =< MAX_VALUE_SIZE` *NB* if `data.len()` is too large `IoRequestTooLarge` is still the preferred error.

# Manually Testing

```
make all
```

checkout the first commit and run:
```
λ cargo nextest run -p octez-riscv-durable-storage --run-ignored only
```
The new test  `test_database_write_value_too_large` will fail. This test is enabled in the second commit.

# Tasks for the Author

### DIFF
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -35,6 +35,10 @@ use crate::storage::KeyValueStore;
 use crate::storage::PersistentKeyValueStore;
 use crate::storage::StoreOptions;
 
+/// The maximum possible length of a value in durable storage.
+pub const MAX_VALUE_SIZE: usize =
+    const { 64_usize.checked_shl(20).expect("usize overflow on 64MiB") };
+
 /// An isolated key-space, independent from other [`Database`]s, on which database operations can
 /// be performed, e.g. read, write, delete.
 ///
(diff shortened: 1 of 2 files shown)


durable-storage/src/database.rs
```diff
@@ -215,6 +219,13 @@ impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
             Err(InvalidArgumentError::IoRequestTooLarge)?;
         }
 
+        const {
+            assert!(
+                MAX_FILE_CHUNK_SIZE < MAX_VALUE_SIZE,
```
kurtisc: <=?

kurtisc (approved):

thomasathorne (approved):

durable-storage/src/database.rs
emturner: Ah, thanks! see [diff](https://github.com/tezos/riscv-pvm/compare/8047c9e456c2af0134910a6b3dff01b57e80bf5f..349d55b97199e6525202436e78f5259a2ba69b0f)

PR #1018, opened 2026-04-23
feat(dursto): Add access-tracking to the ProveResolver.

NSant215:
Part of RV-957.

# What

Add access tracking to the ProveResolver, as part of the refactor of prove and verify modes for the MAVL tree.

# Why

The MerkleProofFold implementation over the Merkle Layer will use to check how the initial tree has been accessed as it is traversed.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/resolver.rs
+++ b/durable-storage/src/avl/resolver.rs
@@ -27,6 +27,8 @@
 //! [`Tree`]: crate::avl::tree::Tree
 //! [`Node`]: crate::avl::node::Node
 
+use std::cell::RefCell;
+use std::collections::BTreeSet;
 use std::rc::Rc;
 use std::sync::Arc;
 use std::sync::OnceLock;
@@ -465,6 +467,17 @@ impl From<Node<ProveTreeId, Prove<'static>>> for ProveNodeId {
     }
 }
 
+/// Access the cached prove-mode node, if populated.
+///
+/// Used by the `MerkleLayer`-level fold to extract per-field read flags from a node that
+/// was resolved during the proof step.
+impl ProveNodeId {
+    #[expect(dead_code, reason = "used by MerkleLayer fold in follow-up commit")]
+    pub(crate) fn cached_node(&self) -> Option<&Node<ProveTreeId, Prove<'static>>> {
+        self.inner.get().map(|rc| rc.as_ref())
+    }
+}
+
 impl Foldable<HashFold> for ProveNodeId {
     fn fold(&self, builder: HashFold) -> <HashFold as Fold>::Folded {
         match self.inner.get() {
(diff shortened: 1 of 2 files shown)


durable-storage/src/avl/resolver.rs
```diff
@@ -523,18 +536,80 @@ impl Foldable<MerkleProofFold> for ProveTreeId {
     }
 }
 
+/// State of the [`ProveResolver`] that tracks which nodes and trees have been accessed via the
+/// corresponding resolution methods
```
emturner: IIUC, this is to track which nodes/trees are accessed, _in the initial tree_? could you possibly add this to the comment?

durable-storage/src/avl/resolver.rs
```diff
@@ -523,18 +536,80 @@ impl Foldable<MerkleProofFold> for ProveTreeId {
...
+    }
+
+    /// Whether a tree with the given hash was accessed during the proof step.
+    #[expect(dead_code, reason = "used by MerkleLayer fold in follow-up commit")]
+    pub(crate) fn was_tree_accessed(&self, hash: &Hash) -> bool {
+        self.accessed_items.borrow().trees.contains(hash)
+    }
+
+    /// Track node access and resolve the underlying lazy node in one step.
+    fn resolve_tracking_node<'a>(
```
```suggestion
    fn resolve_and_track_node<'a>(
```
perhaps? (ditto below)

emturner (approved):

durable-storage/src/avl/resolver.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/7f9e1a16ddf40f68f988ddd1eed646115e7de541..7e237a5eb6a3e39e49498e33779153547c66cf0c)

durable-storage/src/avl/resolver.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/7f9e1a16ddf40f68f988ddd1eed646115e7de541..7e237a5eb6a3e39e49498e33779153547c66cf0c)

durable-storage/src/avl/resolver.rs
```diff
@@ -549,7 +624,7 @@ impl<R: Resolver<LazyNodeId, Node<LazyTreeId, Normal>> + Resolver<LazyTreeId, Tr
             return Ok(inner);
         }
 
-        let resolved: &Node<LazyTreeId, Normal> = self.0.resolve(&id.node)?;
+        let resolved = self.resolve_and_track_node(&id.node)?;
```
victor-dumitrescu, line 627: We only get here if we haven't returned `inner` earlier. Is that because we expect the node to have already been tracked when it was previously resolved? If so, is that a correct assumption given we can `start` a `ProveResolver` with any existing resolver, in which that node could have already been resolved?

durable-storage/src/avl/resolver.rs
```diff
@@ -523,18 +536,80 @@ impl Foldable<MerkleProofFold> for ProveTreeId {
...
+    /// Inner state for access tracking
+    accessed_items: RefCell<AccessedItems>,
+}
 
 impl<R> ProveResolver<R> {
-    /// Create a new prove-mode resolver wrapping the given resolver.
-    pub(crate) fn new(resolver: R) -> Self {
-        Self(resolver)
+    /// Start the prove resolution process.
+    pub fn start(inner: R) -> Self {
```
victor-dumitrescu: I think the constructor can keep being `pub(crate)`?

durable-storage/src/avl/resolver.rs
NSant215: I think that is still fine here. The `inner` field of `ProveNodeId` can only have been populated with a ProveResolver (as that is the only resolver type implemented to resolve `ProveNodeId`). Within the `resolve_and_track_node` method, we populate the `inner` of the `LazyNodeId`, but not the `ProveNodeId::inner` field. Does that answer the question?

durable-storage/src/avl/resolver.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/f869402b6d9efc349f78d9c50555511b6de845bb..71449fd444dffbf79cb466c50a0c3752f95964bb)

victor-dumitrescu (approved):

PR #1020, opened 2026-04-23
test(dursto): run all Database tests across backends when `rocksdb` is on

victor-dumitrescu:
Closes RV-971. Closes RV-975.

# What

Adds a test-only implementation of `PersistentKeyValueStore` for `InMemoryKeyValueStore`. Uses it to convert all*) remaining `Database` tests to the `kv_test` macro.

# Why

In order to run all tests across both the in-memory backend and the persistence layer.

# How

A slight refactoring was needed in production code:
- The requirement that the repo be a `DirectoryManager` for `Database` functions is relaxed so that they can be implemented with the in-memory backend repo.
- As a consequence, the `DirectoryManager`-based default implementations in `PersistentKeyValueStore` were instead moved to the implementation for `PersistenceLayer`.

In test-only code:
- `InMemoryRepo` now tracks commits. This is used for the `PersistentKeyValueStore` implementation for `InMemoryKeyValueStore`. Committing and checking out to/from a path are not supported.
- Adapting helpers and converting tests to `kv_test` is mostly mechanical.

*) The only test not be converted is `test_database_checkout_missing_root_blob_fails_operationally`, since it relies on RocksDB-specific functionality.

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author

### DIFF
--- a/durable-storage/src/storage.rs
+++ b/durable-storage/src/storage.rs
@@ -18,7 +18,6 @@ use tempfile::TempDir;
 use crate::commit::CommitId;
 use crate::errors::Error;
 use crate::errors::OperationalError;
-use crate::repo::DirectoryManager;
 
 /// Types that implement this trait can be used as the underlying key-value store
 pub trait KeyValueStore: Sized {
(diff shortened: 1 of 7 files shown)


durable-storage/src/storage.rs
```diff
@@ -71,25 +70,8 @@ pub trait PersistentKeyValueStore: KeyValueStore + Sized {
     /// Implementations will treat the path as a directory.
     fn commit_to_path(&self, path: &Path) -> Result<(), OperationalError>;
 
-    /// Write the current key-value state to disk.
-    fn commit(&self, repo: &DirectoryManager, id: &CommitId) -> Result<(), OperationalError> {
```
thomasathorne, line 75: I'm not sure I understand your rationale for removing the default implementations of `commit` and `checkout` and putting them explicitly into the `PersistenceLayer` implementation. They still make sense to me as defaults and in theory we could have another implementation of `PersistentKeyValueStore` which _does_ implement `commit_to_path` and `checkout_from_path` where these defaults would make just as much sense?

Not a big deal, just curious if you've got some argument for this I'm not appreciating. 🤔

thomasathorne (approved):

durable-storage/src/storage.rs
victor-dumitrescu: Ah yes, I haven't explained this in the PR description. It's related to the change in interface for these functions, which can now work on any `Repo` instead of just `DirectoryManager` in order to support being implemented for `InMemoryKeyValueStore`. The `PersistenceLayer` implementations are unchanged and still use `DirectoryManager` specifically but it means they can't be the default implementations any more. Updated PR description.

thomasathorne: Ah of course, I didn't notice the change of type signature. 🤦‍♂️

Interesting new feature for the rust type-system, 'partial default' method implementations with extra constraints... 🤔

thomasathorne (approved):

emturner (approved):

PR #1023, opened 2026-04-23
feat(PartialVec): access without allocation

kurtisc:
Part of [RV-964](https://linear.app/tezos/issue/RV-964)

# What
Adds support for accessing a sub-range of a `PartialVec` without requiring access to the whole value (in case e.g. it is blinded/absent in `Verify` mode)

# Why
This is needed to implement `DatabaseMode::get` for `Verify` mode. See #1005 

# How
Small diff

More in-depth testing will be done in a later PR.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/data/src/partial_vec.rs
+++ b/data/src/partial_vec.rs
@@ -311,7 +311,7 @@ impl<T> PartialVec<T> {
     /// Retrieve chunks that make up a continuous range.
     ///
     /// If there are gaps in the range, `None` is returned.
-    pub fn continuous_defined_range(&self, range: Range<usize>) -> Option<Vec<&[T]>> {
+    pub(crate) fn continuous_defined_range(&self, range: Range<usize>) -> Option<Vec<&[T]>> {
         self.range(range)
             .map(|chunk| match chunk {
                 RangeEntry::Undefined { .. } => None,
@@ -359,6 +359,21 @@ impl<T> PartialVec<T> {
     /// Is nothing in the partial vector defined?
     pub fn is_all_undefined(&self) -> bool {
         self.entries.iter().all(|entry| entry.data.is_empty())
+    }
+
+    /// Returns the given range as a contiguous slice if it is fully contained in a single
+    /// defined entry.
+    ///
+    /// Returns `None` if the range spans multiple entries, includes any undefined data, or is empty
+    /// and out-of-bounds.
+    pub fn contiguous_range(&self, range: Range<usize>) -> Option<&[T]> {
+        let idx = self
+            .entries
+            .partition_point(|entry| entry.end() < range.end);
+        let entry = self.entries.get(idx).filter(|e| e.start <= range.start)?;
+        let offset = range.start - entry.start;
+        let end = range.end - entry.start;
+        Some(&entry.data[offset..end])
     }
 }
 
--- a/data/src/partial_vec/tests.rs
+++ b/data/src/partial_vec/tests.rs
@@ -20,6 +20,25 @@ fn data_vec() -> impl Strategy<Value = Vec<u8>> {
 }
 
 proptest! {
+    /// `contiguous_range` returns the correct slice for any subrange of a single defined write.
+    #[test]
+    fn contiguous_range_single_write(
+        offset in 0usize..1000,
+        data in vec(any::<u8>(), 1..128),
+        a in any::<proptest::sample::Index>(),
+        b in any::<proptest::sample::Index>(),
+    ) {
+        let mut vec = PartialVec::empty();
+        vec.define(offset, data.clone());
+
+        let a = a.index(data.len() + 1);
+        let b = b.index(data.len() + 1);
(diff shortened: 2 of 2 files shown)


thomasathorne (approved):

data/src/partial_vec.rs
```diff
@@ -311,7 +311,7 @@ impl<T> PartialVec<T> {
     /// Retrieve chunks that make up a continuous range.
     ///
     /// If there are gaps in the range, `None` is returned.
-    pub fn continuous_defined_range(&self, range: Range<usize>) -> Option<Vec<&[T]>> {
+    pub fn(crate) continuous_defined_range(&self, range: Range<usize>) -> Option<Vec<&[T]>> {
```
NSant215:
```suggestion
    pub(crate) fn continuous_defined_range(&self, range: Range<usize>) -> Option<Vec<&[T]>> {
```

XD

NSant215 (changes requested):

data/src/partial_vec.rs
NSant215, line 345:
```suggestion
    /// In other words, `self.contiguous_range(range)` would succeed (return `Some`).
```

data/src/partial_vec.rs
kurtisc: Forgot to push :( [here](https://github.com/tezos/riscv-pvm/compare/4d9cad30cc1b8b475d6522d9a41530d9b8186e66..777c211aca019479238aef089ee1e853ff1607b4)

data/src/partial_vec.rs
kurtisc: Remove the newline? It's consistent with the method below

data/src/partial_vec.rs
```diff
@@ -360,6 +360,23 @@ impl<T> PartialVec<T> {
     pub fn is_all_undefined(&self) -> bool {
         self.entries.iter().all(|entry| entry.data.is_empty())
     }
+
+    /// Returns the given range as a contiguous slice if it is fully contained in a single
+    /// defined entry.
+    ///
+    /// Returns `None` if the range spans multiple entries or includes any undefined data.
+    pub fn contiguous_range(&self, range: Range<usize>) -> Option<&[T]> {
+        if range.is_empty() {
```
emturner: If the range is empty, but still out of bounds - this should return None I think?

kurtisc: Done [here](https://github.com/tezos/riscv-pvm/compare/01a9539914cffa9de77be1c41a243019ccdd1591..fc6cea24fb2b1ddf595159caf9fe7a4b8a370ff8)

emturner (approved):

data/src/partial_vec.rs
```diff
@@ -360,6 +360,21 @@ impl<T> PartialVec<T> {
...
+    /// Returns the given range as a contiguous slice if it is fully contained in a single
+    /// defined entry.
+    ///
+    /// Returns `None` if the range spans multiple entries, includes any undefined data, or is empty
+    /// and out-of-bounds.
+    pub fn contiguous_range(&self, range: Range<usize>) -> Option<&[T]> {
+        let entry = self
+            .entries
+            .iter()
+            .find(|entry| entry.start <= range.start && range.end <= entry.end())?;
```
victor-dumitrescu: Could use `partition_point` here (like in `range`) to do a binary instead of a linear search.

data/src/partial_vec.rs
```diff
@@ -360,6 +360,21 @@ impl<T> PartialVec<T> {
     pub fn is_all_undefined(&self) -> bool {
         self.entries.iter().all(|entry| entry.data.is_empty())
     }
+
+    /// Returns the given range as a contiguous slice if it is fully contained in a single
+    /// defined entry.
```
victor-dumitrescu, line 365: We ultimately want this to return a slice when all of the data in `range` is present in the partial vector, whether contained in a single entry or not, right? Of course, if it is split we couldn't really return a single slice over it. IIUC it is possible for `PartialVec` to contain contiguous entries. If that's the case, how would we tackle that? (Perhaps a question for a follow-up PR)

NSant215 (approved):

data/src/partial_vec.rs
kurtisc: Nice, [done](https://github.com/tezos/riscv-pvm/compare/fc6cea24fb2b1ddf595159caf9fe7a4b8a370ff8..88526905de2a5125c8fb9f1622282ddf93127a43)

data/src/partial_vec.rs
kurtisc: I think it makes sense to support adjacent ranges, so I've opened #1028 (draft) so I can work on it in parallel.

>Of course, if it is split we couldn't really return a single slice over it

N.B. this is OK because it would mean the data isn't present and verification should fail

victor-dumitrescu (approved):

PR #1026, opened 2026-04-24
feat(dursto): Add deletion tracking

NSant215:
Part of RV-957. 

# What

This is a precursor to #1004 . It introduces a way for the Merkle Layer to keep track of deleted nodes during a proof. 

# Why

Deleted nodes still need to have presence in the initial tree. Once deleted, we cannot track their presence.

# How

We introduce an additional operation to the `Drop` of any `ProveNodeId`, by preserving it in the `ProveResolver::deleted_nodes` field. 

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -94,8 +94,8 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
         self.hash = OnceLock::new();
     }
 
-    #[inline]
     /// A mutable reference to the left branch.
+    #[inline]
     pub(super) fn left_mut<NodeId>(
         &mut self,
         resolver: &mut impl TreeResolver<NodeId, TreeId>,
@@ -104,8 +104,8 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
         resolver.resolve_mut(&mut self.left)
     }
 
-    #[inline]
     /// An immutable reference to the left branch.
+    #[inline]
     pub(super) fn left_ref<NodeId>(
         &self,
         resolver: &impl TreeResolver<NodeId, TreeId>,
@@ -113,8 +113,8 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
         resolver.resolve(&self.left)
     }
 
-    #[inline]
     /// A mutable reference to the right branch.
+    #[inline]
     pub(super) fn right_mut<NodeId>(
         &mut self,
         resolver: &mut impl TreeResolver<NodeId, TreeId>,
@@ -123,8 +123,8 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
         resolver.resolve_mut(&mut self.right)
     }
 
-    #[inline]
     /// An immutable reference to the right branch.
+    #[inline]
     pub(super) fn right_ref<NodeId>(
         &self,
         resolver: &impl TreeResolver<NodeId, TreeId>,
@@ -189,27 +189,33 @@ impl<TreeId, M: BytesMode + AtomMode> Node<TreeId, M> {
         self.hash.get_or_init(|| Hash::from_foldable(self))
     }
 
+    /// The metadata of this [`Node`].
     #[inline]
+    pub(crate) fn meta(&self) -> &Atom<Meta, M> {
+        &self.meta
+    }
+
     /// The difference in heights between child branches.
+    #[inline]
     pub(crate) fn balance_factor(&self) -> i64 {
         self.meta.balance_factor
     }
 
-    #[inline]
     /// A mutable reference to the difference in heights between child branches.
+    #[inline]
     pub(super) fn balance_factor_mut(&mut self) -> &mut i64 {
         &mut self.meta.balance_factor
     }
 
-    #[inline]
     /// The [`Key`] used for determining the [`Node`].
+    #[inline]
     pub(crate) fn key(&self) -> &Key {
         &self.meta.key
     }
 
     /// Retrieve the value associated with this node.
-    #[cfg(test)]
-    pub(crate) fn value(&self) -> &Bytes<M> {
+    #[inline]
+    pub(crate) fn data(&self) -> &Bytes<M> {
         &self.data
     }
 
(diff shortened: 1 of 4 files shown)


durable-storage/src/avl/tree.rs
```diff
@@ -110,19 +113,25 @@ impl<NodeId> Tree<NodeId> {
...
                 (None, Some(right)) => {
-                    *node = right.clone();
+                    let replacement = right.clone();
+                    resolver.on_delete(node);
+                    *node = replacement;
                     Ok(true)
                 }
                 (Some(_), Some(_)) => {
                     let (new_node, shrank) = Node::replace_with_successor(node, resolver)?;
+                    resolver.on_delete(node);
```
victor-dumitrescu: At this point `node` is no longer the original node, but the one replaced by its successor, so we're not capturing the initial state.

NSant215: you're right thanks! changed [here](https://github.com/tezos/riscv-pvm/compare/558dfc7991bdf5ce8baa65d07664b3fae90fe045..a5d6c9d4d864dcfbaaaf7e41b2016fb2ba0958cd)

NSant215: offline discussion with Emma - have changed tactic to exercising the add of the ProveNodeId to within the drop implementation - less chance of wrong placement like here.

durable-storage/src/avl/resolver.rs
```diff
@@ -453,6 +455,23 @@ impl<KV: KeyValueStore> Resolver<LazyTreeId, Tree<LazyNodeId>> for LazyResolver<
...
+        // to be available for the `MerkleLayer`-level fold after the step finishes. Non-resolved
+        // nodes are not relevant for the fold.
+        let Some(inner) = self.inner.get() else {
+            return;
+        };
+        let hash = Hash::from_foldable(&self.node);
+        let cloned = inner.as_ref().clone();
+
+        let displaced = self.deleted_nodes.borrow_mut().insert(hash, cloned);
+        drop(displaced);
```
emturner: discussed offline: we can avoid the recursive `drop` by just keeping `Meta` and `Bytes` in the `deleted_nodes`. `Left` and `Right` get picked up out the resolution already

durable-storage/src/avl/resolver.rs
```diff
@@ -582,6 +614,15 @@ impl<R> ProveResolver<R> {
         self.accessed_items.borrow().trees.contains(hash)
     }
 
+    /// Prove-mode projections of nodes unlinked from the working tree during the step, keyed by
+    /// their initial-tree hash.
+    #[expect(dead_code, reason = "used by MerkleLayer fold in follow-up commit")]
+    pub(crate) fn deleted_nodes(
+        &self,
+    ) -> std::cell::Ref<'_, HashMap<Hash, Node<ProveTreeId, Prove<'static>>>> {
```
emturner: this essentially leaks that a `RefCell` is being used under the hood. Could maybe consider something like `impl Deref` in the return position?

durable-storage/src/merkle_layer.rs
```diff
@@ -1571,13 +1571,13 @@ mod tests {
             .expect("setting node should succeed");
 
         // `Prove` mode
-        let prove_tree = normal_ml.inner.tree.into_proof();
+        let resolver = ProveResolver::start(LazyResolver::new(normal_ml.inner.persistence.clone()));
```
```suggestion
        let resolver = LazyResolver::new(normal_ml.inner.persistence.clone());
        let resolver = ProveResolver::start(resolver);
```
nit: I'd possibly advocate for breaking up these larger/more nested initialisations

durable-storage/src/avl/resolver.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/a2684fdf53c771c577064379bab2106e97a3f2b9..c49f301)

durable-storage/src/avl/resolver.rs
NSant215: ah good point done [here](https://github.com/tezos/riscv-pvm/compare/c49f30190b55d2caa7997d6c25aea39cf0f04d55..5153affbc199f67314758f77afe4cf605ad39ab7)

durable-storage/src/merkle_layer.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/5153affbc199f67314758f77afe4cf605ad39ab7..441ef296c03459dbbc4c1875210200a14bb230b0)

durable-storage/src/avl/resolver.rs
NSant215: removed explicit `drop` [here](https://github.com/tezos/riscv-pvm/compare/441ef296c03459dbbc4c1875210200a14bb230b0..39447c04c139bad28385e7c6157a687e5e316374)

durable-storage/src/avl/node.rs
```diff
@@ -189,6 +189,12 @@ impl<TreeId, M: BytesMode + AtomMode> Node<TreeId, M> {
         self.hash.get_or_init(|| Hash::from_foldable(self))
     }
 
+    #[inline]
+    /// The metadata of this [`Node`].
```
```suggestion
    /// The metadata of this [`Node`].
    #[inline]
```

(and elsewhere)

durable-storage/src/avl/resolver.rs
```diff
@@ -453,6 +459,26 @@ impl<KV: KeyValueStore> Resolver<LazyTreeId, Tree<LazyNodeId>> for LazyResolver<
...
+        // to be available for the `MerkleLayer`-level fold after the step finishes. Non-resolved
+        // nodes are not relevant for the fold.
+        let Some(inner) = self.inner.get() else {
+            return;
+        };
+        let hash = Hash::from_foldable(&self.node);
+        let node = inner.as_ref();
+        let fields = DeletedNodeFields {
+            meta: node.meta().clone(),
+            data: node.data().clone(),
```
emturner, line 477: nit - you should be able to just replace data with an empty bytes, and meta with an empty meta? 

But this would allow to avoid a new allocation for this, as `Key::new(&[])` shouldn't allocate, and likewise `Bytes::new()` shouldn't either? 

Then, you can `mem::swap` to extract the two values we care about and send them along

emturner (approved):

durable-storage/src/avl/node.rs
NSant215: [diff](https://github.com/tezos/riscv-pvm/compare/4de10f3cf85172b2a31a44cad60940542bcde15a..fe7f0c36ecfb164ad50fc121122e9c2efe9b08b1)

durable-storage/src/avl/resolver.rs
NSant215: I am finding an issue in trying this 
- I am trying to create the `DeletedNodeFields` with empty fields, and then swapping them directly with the `node` about to be dropped.
- this requires the Node to be taken mutably
- this is an issue, as the Node to be dropped could have been cloned e.g. line 515 of `rotate_left`.

emturner: ah thanks!

emturner (approved):

victor-dumitrescu (approved):

PR #1028, opened 2026-04-27
Coalesce PartialVec range

kurtisc:
Closes RV-993

# What
Allows `PartialVec::contiguous_range` to return slices which cover multiple entries.

# Why
Currently this requires the whole range to live inside a single entry.

# How
`PartialVec::define` coalesces any adjacent entries it touches using two pointers followed by a drain.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/data/src/partial_vec.rs
+++ b/data/src/partial_vec.rs
@@ -231,6 +231,7 @@ impl<T> PartialVec<T> {
         let mut start_idx = self
             .entries
             .partition_point(|entry| entry.end() <= new_entry.start);
+        let initial_idx = start_idx;
 
         while !new_entry.data.is_empty() {
             // Insert or merge the new entry at the current index.
@@ -238,6 +239,43 @@ impl<T> PartialVec<T> {
             self.insert_entry(start_idx, &mut new_entry);
             start_idx += 1;
         }
+
+        self.coalesce_adjacent(
+            initial_idx.saturating_sub(1),
+            (start_idx + 1).min(self.entries.len()),
+        );
+    }
+
+    /// Merge any consecutive entries in `[from, to_exclusive)` whose ranges touch.
+    ///
+    /// Entries outside that window are assumed to already satisfy the no-adjacency invariant.
+    ///
+    /// Merging leaves emptied entries scattered through the window, but the `swap` below pushes
+    /// them all past `write_ptr` so a single `drain` at the end cleans them up.
+    fn coalesce_adjacent(&mut self, from: usize, to_exclusive: usize) {
+        let to = to_exclusive.min(self.entries.len());
+        if from + 1 >= to {
+            return;
+        }
+
+        let mut write_ptr = from;
+        let mut read_ptr = from + 1;
+
+        while read_ptr < to {
+            if self.entries[write_ptr].end() == self.entries[read_ptr].start {
+                let mut absorbed = std::mem::take(&mut self.entries[read_ptr].data);
+                self.entries[write_ptr].data.append(&mut absorbed);
+            } else {
+                // Keeps emptied entries behind `write_ptr`; see the partition note above.
+                write_ptr += 1;
+                if write_ptr != read_ptr {
+                    self.entries.swap(write_ptr, read_ptr);
+                }
+            }
+            read_ptr += 1;
+        }
+
+        self.entries.drain(write_ptr + 1..to);
     }
 
     /// Mark everything beyond `keep_length` as undefined.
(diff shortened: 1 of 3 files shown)


data/src/partial_vec/tests.rs
```diff
@@ -49,6 +50,21 @@ proptest! {
...
+    /// `entries[i].end()` must be strictly less than `entries[i + 1].start`.
+    #[test]
+    fn entries_never_touching(init_data in vec((any::<usize>(), data_vec()), ..1024)) {
+        let mut vec: PartialVec<u8> = PartialVec::empty();
+
+        for (index, data) in init_data {
+            vec.define(index, data);
+        }
+
+        for pair in vec.entries.windows(2) {
```
emturner, line 63:
```suggestion
        for pair in vec.entries.windows(2) {
        // `DefinedEntry::end` is an exclusive end of the contained range
```

emturner: just for completeness/ease of review

data/src/partial_vec/tests.rs
```diff
@@ -654,3 +670,45 @@ fn contiguous_range_empty() {
...
+fn define_extension_meets_right_neighbour() {
+    let mut vec: PartialVec<u8> = PartialVec::empty();
+    vec.define(0, vec![1, 2, 3]);
+    vec.define(10, vec![11]);
+    vec.define(2, vec![3, 4, 5, 6, 7, 8, 9, 10, 11]);
+
+    assert_eq!(vec.entries.len(), 1);
+    assert_eq!(
+        vec.contiguous_range(0..11),
+        Some(&[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11][..])
```
emturner, line 713: it would be great to see some overwriting in action? e.g. partially or fully overwriting the left entry, partially or fully overwriting the right entry. 

Does this apply for a subsequent entry to the right if it's big enough etc

data/src/partial_vec.rs
```diff
@@ -231,13 +231,47 @@ impl<T> PartialVec<T> {
...
+            initial_idx.saturating_sub(1),
+            (start_idx + 1).min(self.entries.len()),
+        );
+    }
+
+    /// Merge any consecutive entries in `[from, to_exclusive)` whose ranges touch.
+    ///
+    /// Entries outside that window are assumed to already satisfy the no-adjacency invariant.
+    fn coalesce_adjacent(&mut self, from: usize, to_exclusive: usize) {
+        let to = to_exclusive.min(self.entries.len());
```
NSant215, line 256: is this line needed if already checked in the call? or is it just safer to have even though it shouldn't be exercised in practice based on the call?

data/src/partial_vec.rs
```diff
@@ -231,13 +231,47 @@ impl<T> PartialVec<T> {
...
+
+        while read < to {
+            if self.entries[write].end() == self.entries[read].start {
+                let mut absorbed = std::mem::take(&mut self.entries[read].data);
+                self.entries[write].data.append(&mut absorbed);
+            } else {
+                write += 1;
+                if write != read {
+                    self.entries.swap(write, read);
+                }
```
NSant215: I don't think this is wrong as the test coverage includes it, but I am not following the algorithm here and can't identify the test that covers it? What is the idea behind this part and which test covers it?

thomasathorne: It took me quite a bit of thought to work out that this logic is (as far as I can tell) correct. I was worrying about cases where we coalesce a set of more than two entries followed by a gap and then more coalescing later. Perhaps worth adding a unit test for just the `coalesce_adjacent` function that covers this?

I'm struggling to think of a concise comment that could explain the `swap` and `drain` combination, but maybe some kind of comment here and below would help. Something like "this algorithm leaves a load of emptied entries along the way, but because of the `swap` they all end up _after_ the `write` head in the `entries` vec and hence are drained at the end" maybe? 🤔

thomasathorne (approved):

data/src/partial_vec.rs
NSant215: ok happy with it now after offline discussion - maybe we could call them `write_ptr` and `read_ptr`, and a comment over the swap saying that we are moving `drained` portions to the end and future targets to the front.

kurtisc: Both suggestions implemented [here](https://github.com/tezos/riscv-pvm/compare/3cfac37dfa8d7dae75d9cfd6ee9e5f2a3e238f99..c7b6c04e810840c3da8a52e3ffeb7e8a870347a3#diff-bf531aca23479236c870ba1879e3b8fb8522a7f4c683903bc801f1c575a6f611R252)

data/src/partial_vec/tests.rs
kurtisc: [Done](https://github.com/tezos/riscv-pvm/compare/3cfac37dfa8d7dae75d9cfd6ee9e5f2a3e238f99..c7b6c04e810840c3da8a52e3ffeb7e8a870347a3#diff-5a6a577cb209ba6ce09b8b020afee1b4f7366feef94f64687190bb1bf48ae0b0R64)

data/src/partial_vec/tests.rs
kurtisc: More tests [here](https://github.com/tezos/riscv-pvm/compare/3cfac37dfa8d7dae75d9cfd6ee9e5f2a3e238f99..c7b6c04e810840c3da8a52e3ffeb7e8a870347a3#diff-5a6a577cb209ba6ce09b8b020afee1b4f7366feef94f64687190bb1bf48ae0b0R716)

data/src/partial_vec.rs
kurtisc: I'd prefer to keep defensive checks like these because there's nothing enforcing it in the type checker so removing it increases the area where changes could cause bugs. That it practically won't be called without being checked is potentially temporary.

NSant215 (approved):

emturner (approved):

PR #1034, opened 2026-04-29
test(dursto): `kv_test` runs PBTs on same inputs and compares traces

victor-dumitrescu:
Part of RV-979.

# What

Expands the `kv_test!` macro in preparation for comparing traces across test runs:

- A new variant is added for property-based tests which lifts input generation at the macro level.
- Another variant is also added for property-based tests which sets up the test before running the proptest rloop.
- The existing variant, now only used for unit tests, produces a single test function which runs using both backends instead of one test function per backend.
- Both variants compare traces produced by each test runs. These are currently just unit.

Producing meaningful traces will be added in future PRs:
- #1036 

# Why

We want to be able to compare execution traces between tests which run over multiple configurations in order to assess semantic consistency.

# How

The `kv_test!` macro is significantly reworked, mostly because using a defined `inner` function for the test body is no longer possible. On one hand, for the proptest variant, we want to run test bodies with varying number and types of arguments to account for the proptest inputs. On the other, both variants will need to support getting a return value from running a test. I originally wanted to pin this to something like `-> impl Debug + PartialEq` but that doesn't work, so I resorted to keeping the bounds check as a separate function and not constraining the return type at all in the macro. This allows all the existing tests which return unit to work with no further modifications and, in the future, for tests in each module to return different types of traces.

Apart from the expanded macro, proptest-based tests are rewritten to use the new macro variant.

# Manually Testing

```
make all
```

The diff for the second commit is inflated by the formatting changes in PBTs. I suggest reviewing using `git diff -w` or similar.

# Tasks for the Author

### DIFF
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -479,7 +479,6 @@ mod tests {
     use octez_riscv_data::mode::Verify;
     use proptest::prelude::*;
     use proptest::prop_assert_eq;
-    use proptest::proptest;
     use tezos_smart_rollup_constants::core::MAX_FILE_CHUNK_SIZE;
     use tokio::runtime::Handle;
 
@@ -493,6 +492,7 @@ mod tests {
     use crate::merkle_layer::new_verify_layer;
     use crate::merkle_worker::BackgroundKeyValueStore;
     use crate::merkle_worker::BackgroundPersistentKeyValueStore;
+    use crate::storage::KeyValueStore;
     use crate::storage::TestKeyValueStoreSetup;
     use crate::storage::kv_test;
 
@@ -503,14 +503,15 @@ mod tests {
         Database::try_new(handle, repo).expect("Creating a test database should succeed")
     }
 
-    fn new_verify_database<KV: TestKeyValueStoreSetup>() -> Database<KV, Verify> {
+    fn new_verify_database<KV: KeyValueStore>(repo: &KV::Repo) -> Database<KV, Verify> {
         Database {
             inner: super::VerifyImpl {
-                merkle: new_verify_layer::<KV>(),
+                merkle: new_verify_layer::<KV>(repo),
             },
         }
     }
 
+    #[cfg(feature = "rocksdb")]
     fn new_persistent_database<KV>() -> (
         tokio::runtime::Runtime,
         KV::Keepalive,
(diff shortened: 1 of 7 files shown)


durable-storage/src/storage.rs
```diff
@@ -151,32 +151,105 @@ cfg_if::cfg_if! {
...
+                            type $ty_name = $crate::storage::in_memory::InMemoryKeyValueStore;
+                            _bound_check::<$ty_name>();
+                            $body
+                        };
+
+                        #[cfg(feature = "rocksdb")]
+                        ::proptest::prop_assert_eq!(
+                            rocksdb_trace, in_memory_trace,
+                            "trace mismatch"
+                        );
```
emturner:
```suggestion
                        let _in_memory_trace = {
                            type $ty_name = $crate::storage::in_memory::InMemoryKeyValueStore;
                            _bound_check::<$ty_name>();
                            $body
                        };

                        #[cfg(feature = "rocksdb")]
                        ::proptest::prop_assert_eq!(
                            rocksdb_trace, _in_memory_trace,
                            "trace mismatch"
                        );
```
nit - feel free to resolve, but in this sort of macro/test code - just preceding with underscore is likely good enough?

durable-storage/src/database.rs
```diff
@@ -472,59 +471,55 @@ mod tests {
...
+            checked_out.assert_database_value(&key, value.as_ref());
+        }
     });
 
-    kv_test!(test_database_delete, KV: BackgroundKeyValueStore, {
+    kv_test!(test_database_delete, KV: BackgroundKeyValueStore, [
+        keys in prop::collection::vec(prop::collection::vec(any::<u8>(), 0..=KEY_MAX_SIZE), 0..100),
+        data in prop::collection::vec(prop::collection::vec(any::<u8>(), 0..200), 0..100),
+    ], {
         let runtime = tokio::runtime::Builder::new_multi_thread()
```
emturner: it would be great, if possible, to keep this sort of pre-amble/setup code outside of the main proptest loop? 

Otherwise, it potentially makes each loop that much more expensive, making the whole test-suite run a bit slower.

 `TestableTmpDir` would also not keep the temp state around on failures - since it's designed to work with panics. So ideally it exists before the call to `proptest!`

durable-storage/src/storage.rs
victor-dumitrescu: I think you're right, removed  the `expect`s from the latest version.

durable-storage/src/database.rs
victor-dumitrescu: Oh, that's a good point! It's possible to do, although I haven't been able to figure out a way which doesn't add quite a bit of complexity to the macro. I've done this in the latest version.

I think that's fine though because we're always setting up more or less the same things, with a few variations, and I think the usability of the macro isn't affected too much, but users will have to pull out setup code from the main test body. Let me know what you think.

emturner (approved):

durable-storage/src/storage.rs
```diff
@@ -149,34 +149,163 @@ cfg_if::cfg_if! {
...
+        ///    - `setup_runtime |handle, repo| = { ... }` for tests that need a Tokio
+        ///      runtime. The block must return `(Runtime, Handle, KV::Keepalive, KV::Repo)`.
+        ///      Inside the test body, `handle: &Handle` and `repo: &KV::Repo` are exposed.
+        ///    - `setup |repo| = { ... }` for tests that only need a repo. The block
+        ///      must return `(KV::Keepalive, KV::Repo)`. Inside the test body only
+        ///      `repo: &KV::Repo` is exposed.
+        ///
+        ///    Note: In order to run the test body twice, the generated inputs are cloned.
+        ///    For this to be valid, cloning must create a deep copy. This means that proptests over
+        ///    pointers should not be run with this macro.
```
NSant215: when do we use pointers in proptests? or what could be a situation where we would use them? not a problem just curious

durable-storage/src/storage.rs
```diff
@@ -149,34 +149,163 @@ cfg_if::cfg_if! {
...
+                    });
+                }
+            };
+
+            // Unit test variant
+            ($(#[$attr:meta])* $fun_name:ident, $ty_name:ident $(: $ty_bound:path)?, $body:block) => {
+                $(#[$attr])*
+                #[test]
+                fn $fun_name() {
+                    fn _bound_check<
```
NSant215, line 300: what is the need for this part?

durable-storage/src/storage.rs
```diff
@@ -149,34 +149,163 @@ cfg_if::cfg_if! {
...
+                            let $setup_values = &rocksdb_setup;
+                            $(let $arg = ::std::clone::Clone::clone(&$arg);)*
+                            $body
+                        };
+
+                        let _in_memory_trace = {
+                            type $ty_name = $crate::storage::in_memory::InMemoryKeyValueStore;
+                            let $setup_values = &in_memory_setup;
+                            $body
+                        };
```
NSant215, line 284: would a test failure tell us whether it occurred in memory or rocksdb? If not could we add eprints to do that?

durable-storage/src/storage.rs
NSant215: like what is the bound check doing

durable-storage/src/storage.rs
victor-dumitrescu: Good point, you couldn't really tell unless you looked at which line panicked in the expanded file. I've added print statements [here](https://github.com/tezos/riscv-pvm/compare/17ace7c3c33f172c6a5de33ab877153ba328019f..7d4c3dd696d1bbd1b48270f2729a6d302b4be09d).

durable-storage/src/storage.rs
victor-dumitrescu: You can pass a type bound for the `KV` type when you call the macro, for example

```rust
    kv_test!(test_merkle_layer_checkout_lazy_from_commit, KV: PersistentKeyValueStore,
```

In addition, all tests expect that `KV: TestKeyValueStoreSetup` and that `KV::Repo: RegistryRepo`.

Previously, these were checked by construing an `inner` function using the body of the test whose type encoded these constraints, but this isn't possible any more in the reworked macro (explained in PR body). For PBTs, this constraint is still checked as part of the setup function. For unit tests, this `_bound_check` declaration is generated, which only needs to type check.

durable-storage/src/storage.rs
victor-dumitrescu: It's not used at the moment, but proptest does support it. This is just to document that using it in that way might have unexpected results, if, for example, the test mutates the input data.

NSant215 (approved):

PR #1036, opened 2026-04-29
test(dursto): database tests produce execution traces

victor-dumitrescu:
Closes RV-979.

# What

Adds a test-only mechanism for recording execution traces from running operations on a `Database`.

# Why

We want to be able to compare execution traces between tests which run over multiple configurations in order to assess semantic consistency.

# How

In test code, `TracedDatabase` is introduced, which wraps a `Database` exposing the same interface. Using it instead of `Database` records a `TraceEntry` with the inputs and outcome of every operation applied to it. When run as `kv_test!` with `rocksdb` on, database tests run with both backends and fail if traces don't match between executions.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -527,6 +527,8 @@ impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
     }
 }
 
+mod traced_database;
+
 #[cfg(test)]
 mod tests {
     use std::collections::HashSet;
@@ -543,7 +545,7 @@ mod tests {
     use tokio::runtime::Handle;
 
     use super::Database;
-    use super::DatabaseMode;
+    use super::traced_database::TracedDatabase;
     use crate::avl::tree::Tree;
     use crate::database::MAX_VALUE_SIZE;
     use crate::errors::Error;
@@ -551,7 +553,6 @@ mod tests {
     use crate::key::KEY_MAX_SIZE;
     use crate::key::Key;
     use crate::merkle_layer::MerkleLayer;
-    use crate::merkle_layer::new_verify_layer;
     use crate::merkle_worker::BackgroundKeyValueStore;
     use crate::merkle_worker::BackgroundPersistentKeyValueStore;
     use crate::storage::KeyValueStore;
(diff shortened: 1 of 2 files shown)


durable-storage/src/database.rs
```diff
@@ -470,6 +470,297 @@ impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
     }
 }
 
+#[cfg(test)]
+pub(crate) use traced_database::*;
+
+#[cfg(test)]
```
emturner: the `database` module is getting pretty large - might be worth moving traced_database into its own file?

durable-storage/src/database.rs
```diff
@@ -470,6 +470,297 @@ impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
...
+            result: Result<Hash, String>,
+        },
+        Commit {
+            result: Result<CommitId, String>,
+        },
+    }
+
+    pub(crate) struct TracedDatabase<KV, M: Mode = Normal> {
+        inner: Database<KV, M>,
+        trace: Rc<RefCell<Vec<TraceEntry>>>,
```
emturner: why an Rc? is this to cope with `copy` in future? 

I guess for now that works as we're still single-threaded

durable-storage/src/database.rs
```diff
@@ -470,6 +470,297 @@ impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
...
+
+        pub(crate) fn into_trace(self) -> Trace {
+            match Rc::try_unwrap(self.trace) {
+                Ok(cell) => cell.into_inner(),
+                Err(rc) => rc.borrow().clone(),
+            }
+        }
+
+        // Test helpers
+
```
emturner:
```suggestion
```

(nit - in general I'm not the biggest fan of section headers. But could either put a proper multiline one in or just get rid. The floating single-line comment looks a little odd I think?)

emturner (approved):

durable-storage/src/database.rs
victor-dumitrescu: Indeed, moved to a separate module.

durable-storage/src/database.rs
victor-dumitrescu: Yes this is for a future PR for `TracedRegistry`, I removed it for now.

durable-storage/src/database.rs
victor-dumitrescu: Removed

durable-storage/src/database.rs
```diff
@@ -788,24 +768,25 @@ mod tests {
...
 
-        let derived_reloaded =
-            Database::<KV, _>::checkout(handle, &repo, derived_commit)
-                .expect("Checkout should succeed");
+        let derived_reloaded = TracedDatabase::<KV>::checkout(handle, &repo, derived_commit)
+            .expect("Checkout should succeed");
         derived_reloaded.assert_database_value(&persisted_key, b"after");
         derived_reloaded.assert_database_value(&derived_key, b"new");
+
+        original.into_trace()
```
NSant215: can we add enforcing of the trace within the macro? or add this line itself to the body via the macro?

NSant215 (approved):

emturner (approved):

durable-storage/src/database.rs
emturner: I guess you could do it by requiring the body returns an `impl Into<Trace...>` from the block?
But then you'd end with

```
original
```

At the end of each function - which might look even more odd?

durable-storage/src/database.rs
```diff
@@ -1792,6 +1829,8 @@ mod tests {
             database.read_bytes(&key, 0, MAX_FILE_CHUNK_SIZE + 1),
             Err(Error::InvalidArgument(InvalidArgumentError::IoRequestTooLarge))
         ));
+
+        database.into_trace()
     });
 
     kv_test!(test_prove_database_read_bytes_records_only_accessed_range, KV: BackgroundKeyValueStore, {
```
NSant215, line 1836: is this test missing a trace?

durable-storage/src/database.rs
NSant215: my concern was that it would be easy to miss putting the trace at the end of the tests as well - could it be added automatically in the macro instead?

durable-storage/src/database.rs
victor-dumitrescu: I added a trace for `prove_db` [here](https://github.com/tezos/riscv-pvm/compare/d09b8382ab74018c98a4e930548075c06ad0cb68..4abe7f926b6ccda43f92227896089c9b2d88be8e), more for completeness / as an example since in this test specifically the trace will just contain a read, which the test checks too.

durable-storage/src/database.rs
victor-dumitrescu: I'm not sure there's a good way of doing this, ultimately each test needs to decide what to trace. Most (but not all) tests in this module trace a single database, but that can also trace more than one, or the name can be different from `database`. We'll also trace other things next, such as registries.

PR #1038, opened 2026-04-29
fix(dursto): Fix Prove and Verify Implementations of the Merkle Layer.

NSant215:
Closes RV-957.
Closes RV-961.

  # What

  Fix the `Prove` and `Verify` implementations of `MerkleLayer` so that proof generation
  and verification round-trip correctly across AVL-rebalancing operations
  (`set` on existing/new keys, `delete`, `write`).

  # Why

  The previous prove-mode fold walked the *working* tree, so any AVL rotation triggered
  by the step rearranged the structure relative to the proof and produced root hashes
  that no longer matched the Normal-mode hash. On the verify side, sub-proofs decoded
  into `VerifyNodeId` / `VerifyTreeId` were folded against the inherited parent proof
  rather than the sub-proof they originated from — this caused incorrect `prev_hash`
  substitutions or `PartialHash::InvalidProof` once nodes had been moved by rebalancing.

  # How

  Two structural fixes plus tests:

  1. **Verify side**: each `VerifyNodeId::Present` / `VerifyTreeId::Present` now carries
     the `MerkleProof` sub-tree it was deserialised from, captured up-front via the
     new `Deserialiser::capture_owned_proof()` hook. Their `Foldable<PartialHashFold>`
     impls override the inherited proof with this captured sub-proof before delegating,
     so each node folds against its initial-tree position even after rebalancing.
  2. **Prove side**: `start_proof()` snapshots the Normal-mode tree as an immutable
     `initial_tree`. The `MerkleProofFold` impl on `MerkleLayer<KV, Prove<'_>>` walks
     that snapshot — not the mutating working tree — using `was_node_accessed`/
     `was_tree_accessed` to decide what to blind, and pulling per-field read flags from
     the prove-mode `working_tree` / `deleted_nodes` map.

  # Per-commit breakdown

  1. Plumbing:
     - `PartialHashFold.proof` becomes `Option<Arc<MerkleProof>>` so callers can attach
       a node-local override without deep-cloning.
     - Adds `PartialHashFold::with_proof()` (replace outright, unlike
       `map_reference_proof` which requires `Some`).
     - Adds `Deserialiser::capture_owned_proof()` (default `None`; overridden by
       `ProofTree` to clone the in-memory sub-proof). Stream deserialisers can't
       reconstruct without re-parsing, so the default is correct.

  2. Use the new override:
     - `VerifyNodeId::Present` / `VerifyTreeId::Present` become struct variants
       carrying `original_proof: Option<Arc<MerkleProof>>`, captured before
       `into_node()` consumes the deserialiser.
     - Their `Foldable<PartialHashFold>` impls call `builder.with_proof(...)` before
       delegating to the inner `Node` / `Tree`, so the fold sees each node's
       own sub-proof rather than the parent's view.
     - Resolver impls and tests updated for the new struct layout.

  3. Walk the initial tree, not the mutating one:
     - Removes the broken `Foldable<MerkleProofFold>` impls on `Tree<ProveNodeId>` and
       `ProveNodeId`.
     - Adds `Foldable<MerkleProofFold>` for `MerkleLayer<KV, Prove<'_>>`, implemented
       via two wrappers (`InitialTreeFold` / `InitialNodeFold`) that recurse through the
       initial `LazyTreeId` snapshot, blinding any node/tree the resolver did not
       touch and pulling read-flag-bearing `meta` / `data` from either the working tree
       (via `Tree::get_resolved` + the new `CachedOnlyResolver`) or `deleted_nodes` for
       keys removed during the step.
     - `MerkleLayer<Normal>::start_proof()` is the new entry point: it clones the
       Normal tree, derives the prove-mode `working_tree` from it, and seeds the
       `ProveResolver` with the root hash so the root tree is always considered
       accessed (`ProveResolver::start` now takes `root_tree_hash`).
     - `ProveImpl` is now `{ initial_tree, working_tree, resolver }`. Existing
       prove-mode write-paths and tests adapted to the rename.
     - Removes `#[expect(dead_code, …)]` on `cached_node`, `inner`, `was_*_accessed`,
       `deleted_nodes`, `DeletedNodeFields` now that the fold consumes them.

  4. PBTs:
     - `assert_prove_mode_correct` asserts (1) `proof.root_hash() == initial Normal
       hash`, (2) prove-final == Normal-final after replaying the same op, (3)
       verify-mode replay over the deserialised proof reaches the same final hash.
     - Five proptest cases: random single-op (set/delete/write), writes-only,
       deletes-only, new-key insertions, and read-only.
     - Single-op cases note the current `ops_count = 1` limitation
       (multi-step proofs not yet supported).

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author

### DIFF
--- a/data/src/merkle_proof/proof_tree.rs
+++ b/data/src/merkle_proof/proof_tree.rs
@@ -515,6 +515,13 @@ impl<'t> Deserialiser for ProofTree<'t> {
     fn into_node(self) -> Result<Self::DeserialiserNode, Self::Error> {
         Ok(self.as_node()?.map_present(Vec::into_iter))
     }
+
+    fn capture_owned_proof(&self) -> Option<MerkleProof> {
+        match self {
+            ProofPart::Absent => None,
+            ProofPart::Present(proof) => Some((*proof).clone()),
+        }
+    }
 }
 
 /// Similar to [`ProofTree`], but owns the underlying [`MerkleProof`]
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -25,13 +25,12 @@ use octez_riscv_data::merkle_proof::DeserialiserNode;
 use octez_riscv_data::merkle_proof::FromProof;
 use octez_riscv_data::mode::Mode;
 use octez_riscv_data::mode::Normal;
-use octez_riscv_data::mode::Prove;
 use octez_riscv_data::serialisation::deserialise;
 use octez_riscv_data::serialisation::serialise;
 use perfect_derive::perfect_derive;
 
 use super::resolver::LazyTreeId;
-use super::resolver::ProveTreeId;
+use super::resolver::ProveNode;
 use super::resolver::TreeResolver;
 use super::tree::Tree;
 use crate::avl::resolver::AvlResolver;
@@ -76,8 +75,8 @@ pub struct Node<TreeId, M: Mode> {
 }
 
 impl Node<LazyTreeId, Normal> {
-    /// Converts the [`Node`] to [`Prove`] mode.
-    pub fn into_proof(self) -> Node<ProveTreeId, Prove<'static>> {
+    /// Converts the [`Node`] to prove mode.
+    pub fn into_proof(self) -> ProveNode {
         Node {
             meta: self.meta.into_proof(),
             data: self.data.into_proof(),
@@ -92,6 +91,11 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
     /// Mark the hash of this node as dirty.
     fn invalidate_hash(&mut self) {
         self.hash = OnceLock::new();
+    }
+
+    /// Direct access to the left child identifier.
+    pub(crate) fn left_id(&self) -> &TreeId {
+        &self.left
     }
 
     /// A mutable reference to the left branch.
@@ -111,6 +115,11 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
         resolver: &impl TreeResolver<NodeId, TreeId>,
     ) -> Result<&Tree<NodeId>, OperationalError> {
         resolver.resolve(&self.left)
+    }
+
+    /// Direct access to the right child identifier.
+    pub(crate) fn right_id(&self) -> &TreeId {
+        &self.right
     }
 
     /// A mutable reference to the right branch.
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -134,20 +134,21 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
 }
 
 impl<TreeId, M: BytesMode + AtomMode> Node<TreeId, M> {
-    /// The data stored in a [`Node`] within the subtree of this [`Node`] with a given [`Key`] .
+    /// Find the id of the [`Node`] within the subtree of this [`Node`] with a given [`Key`].
     pub(super) fn get<'a, NodeId>(
         mut node: &'a NodeId,
         key: &Key,
         resolver: &impl AvlResolver<NodeId, TreeId, M>,
-    ) -> Result<Option<&'a Bytes<M>>, OperationalError>
+    ) -> Result<Option<&'a NodeId>, OperationalError>
     where
         NodeId: 'a,
         TreeId: 'a,
+        M: 'a,
     {
         loop {
             let resolved_node = resolver.resolve(node)?;
             match resolved_node.key().cmp(key) {
-                Ordering::Equal => return Ok(Some(&resolved_node.data)),
+                Ordering::Equal => return Ok(Some(node)),
                 Ordering::Greater => {
                     let Some(left) = resolved_node.left_ref(resolver)?.root() else {
                         return Ok(None);
--- a/durable-storage/src/avl/tree.rs
+++ b/durable-storage/src/avl/tree.rs
@@ -398,6 +398,14 @@ impl Foldable<MerkleProofFold> for Tree<ProveNodeId> {
 
 impl Foldable<PartialHashFold> for Tree<VerifyNodeId> {
     fn fold(&self, builder: PartialHashFold) -> PartialHash {
+        // SAFETY: Extra care is required in folding for `Verify` in the MAVL tree, due to
+        // consequences from restructuring due to AVL balancing. This fold correctness relies
+        // on:
+        // (a) the `bool_leaf` always folding to `PartialHash::Present(...)` so its `prev_hash` is
+        // never substituted from the popped proof child.
+        // (b) the inner `VerifyNodeId` overrides the proof with its own captured sub-proof, before
+        // delegating to `Node::fold`. Otherwise, using `previous` could result in incorrect
+        // substitution.
         let mut node = builder.into_node_fold();
 
         let present = self.0.is_some();
--- a/durable-storage/src/avl/tree.rs
+++ b/durable-storage/src/avl/tree.rs
@@ -156,12 +156,12 @@ impl<NodeId> Tree<NodeId> {
     }
 
     #[inline]
-    /// The data stored in a [`Node`] in the [`Tree`] with a given [`Key`].
-    pub(crate) fn get<'a, TreeId: 'a, M: BytesMode + AtomMode>(
+    /// Find the id of the [`Node`] in the [`Tree`] with a given [`Key`].
+    pub(crate) fn get<'a, TreeId: 'a, M: BytesMode + AtomMode + 'a>(
         &'a self,
         key: &Key,
         resolver: &impl AvlResolver<NodeId, TreeId, M>,
-    ) -> Result<Option<&'a Bytes<M>>, OperationalError> {
+    ) -> Result<Option<&'a NodeId>, OperationalError> {
         let Some(node) = self.root() else {
             return Ok(None);
         };
@@ -983,7 +983,10 @@ mod tests {
             for operation in operations {
                 match operation {
                     Operation::Get(key) => {
-                        let tree_value = tree.get(&key, &resolver)?;
+                        let tree_value = match tree.get(&key, &resolver)? {
+                            Some(node_id) => Some(resolver.resolve(node_id)?.data()),
+                            None => None,
+                        };
 
                         // The values in the Options are comparable, but the Options themselves are
                         // not. So we need to awkwardly match on both Options to compare them.
@@ -1081,10 +1084,14 @@ mod tests {
             Err(Error::InvalidArgument(InvalidArgumentError::OffsetTooLarge))
         ));
 
-        let value = tree
+        let node_id = tree
             .get(&existing, &resolver)
             .expect("Resolver failure not expected.")
             .expect("Pre-existing key should still be present.");
+        let value = resolver
+            .resolve(node_id)
+            .expect("Resolver failure not expected.")
+            .data();
         let mut buf = [0u8; 11];
         value.read(0, &mut buf);
         assert_eq!(&buf, b"zero offset");
@@ -1092,10 +1099,14 @@ mod tests {
 
         tree.write(&existing, 1, b"o", &mut resolver)
             .expect("Writing to an existing key with a non-zero offset should succeed.");
-        let value = tree
+        let node_id = tree
             .get(&existing, &resolver)
             .expect("Resolver failure not expected.")
             .expect("Pre-existing key should still be present.");
+        let value = resolver
+            .resolve(node_id)
+            .expect("Resolver failure not expected.")
+            .data();
         let mut buf = [0u8; 11];
         value.read(0, &mut buf);
         assert_eq!(&buf, b"zoro offset");
--- a/durable-storage/src/avl/tree.rs
+++ b/durable-storage/src/avl/tree.rs
@@ -24,8 +24,6 @@ use octez_riscv_data::merkle_proof::Deserialiser;
 use octez_riscv_data::merkle_proof::DeserialiserNode;
 use octez_riscv_data::merkle_proof::FromProof;
 use octez_riscv_data::merkle_proof::Partial;
-use octez_riscv_data::merkle_proof::proof_tree::MerkleProofFold;
-use octez_riscv_data::merkle_proof::proof_tree::MinimumPresence;
 use octez_riscv_data::mode::utils::not_found;
 use octez_riscv_data::serialisation::deserialise;
 use octez_riscv_data::serialisation::serialise;
@@ -230,7 +228,7 @@ impl<NodeId> Tree<NodeId> {
 
     #[inline]
     /// A reference to the root [`Node`].
-    pub(super) fn root(&self) -> Option<&NodeId> {
+    pub(crate) fn root(&self) -> Option<&NodeId> {
         self.0.as_ref()
     }
 
@@ -380,23 +378,6 @@ impl<NodeId: Foldable<HashFold>> Foldable<HashFold> for Tree<NodeId> {
 
         let present = self.0.is_some();
         node.add(&Hash::hash_encodable(present).expect("Hashing a bool should never fail"));
-
-        if let Some(inner) = self.0.as_ref() {
-            node.add(inner);
-        }
-
-        node.done()
-    }
-}
-
-impl Foldable<MerkleProofFold> for Tree<ProveNodeId> {
-    fn fold(&self, builder: MerkleProofFold) -> <MerkleProofFold as Fold>::Folded {
-        let mut node = builder.into_node_fold();
-
-        let present = self.0.is_some();
-        let bool_data = serialise(present).expect("Serialising a bool should never fail");
-        let bool_leaf = MerkleProofFold::new_leaf(MinimumPresence::Present, bool_data);
-        node.add(&bool_leaf);
 
         if let Some(inner) = self.0.as_ref() {
             node.add(inner);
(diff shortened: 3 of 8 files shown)


data/src/merkle_proof/proof_tree.rs
```diff
@@ -515,6 +515,13 @@ impl<'t> Deserialiser for ProofTree<'t> {
     fn into_node(self) -> Result<Self::DeserialiserNode, Self::Error> {
         Ok(self.as_node()?.map_present(Vec::into_iter))
     }
+
+    fn capture_owned_proof(&self) -> Option<MerkleProof> {
```
emturner, line 519: I guess we might want to follow up on all of this, to see if we can avoid extra allocations - but works for now

durable-storage/src/avl/tree.rs
```diff
@@ -348,6 +349,37 @@ impl<NodeId> Tree<NodeId> {
...
+impl Tree<ProveNodeId> {
+    /// Find the `ProveNodeId` and its resolved [`Node`] for a given key by traversing the tree.
+    ///
+    /// Returns `None` if the key is not present.
+    ///
+    /// Only traverses already-resolved nodes. Returns an error if an
+    /// unresolved node or tree is encountered.
+    pub(crate) fn get_resolved(
+        &self,
+        key: &Key,
```
emturner: could this not just use `get` above? just need to pass the `CachedOnlyResolver`

durable-storage/src/merkle_layer.rs
```diff
@@ -1975,4 +2023,264 @@ mod tests {
...
+            StepOp::Delete(key) => {
+                ml.delete(key).expect("verify delete should succeed");
+            }
+            StepOp::Write(key, data) => {
+                if ml.get(key).expect("verify get should succeed").is_some() {
+                    ml.write(key, 0, data).expect("verify write should succeed");
+                }
+            }
+        }
+    }
```
emturner: I'm 99% sure these three methods can be squashed into a mode-generic one?

durable-storage/src/merkle_layer.rs
```diff
@@ -1975,4 +2023,264 @@ mod tests {
...
+            .expect("proof deserialisation should succeed")
+            .into_result();
+        let VerifyTreeId::Present {
+            tree: verify_tree, ..
+        } = verify_tree_id
+        else {
+            panic!("expected Present tree from proof");
+        };
+
+        let mut verify_ml: MerkleLayer<KV, Verify> = MerkleLayer::from_verify_tree(verify_tree);
```
emturner: could also assert the partial hash at this point matches the initial hash?

data/src/merkle_proof/proof_tree.rs
NSant215: added issue [here](https://github.com/tezos/riscv-pvm/compare/60a334910fa48cb2aa29006d2cabd1e28a4af6a7..7fb176528acd2974ef66aa5fce1334988b4c77cc)

durable-storage/src/avl/tree.rs
NSant215: unfortunately not - `get` specifically only returns the `data` of the node that has the key, rather than the node id. Changed name to make that more clear [here](https://github.com/tezos/riscv-pvm/compare/7fb176528acd2974ef66aa5fce1334988b4c77cc..2c030210eafc5a3f5af94d71bc4e3fe54eb46fc8) 

I could also change the get method itself - wasn't sure what made more sense but happy either way

durable-storage/src/merkle_layer.rs
NSant215: is this better? [diff](https://github.com/tezos/riscv-pvm/compare/2c030210eafc5a3f5af94d71bc4e3fe54eb46fc8..1f40a9351380079f1f5a114933e6b02b6c6503cd)

durable-storage/src/merkle_layer.rs
NSant215: This made me realise that I need to implement `hash` for Verify as well, which means that the `MerkleProof` MUST be a field on Verify. Pulling back to draft whilst I work on that.

NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/1f40a9351380079f1f5a114933e6b02b6c6503cd..9278f76f74e92f46f04572d24057290379644e17)

durable-storage/src/merkle_layer.rs
emturner: I think just something like this is okay? Is there any particular reason we have to do the `get` first?
```
    fn apply_step_op<KV: KeyValueStore, M: MerkleLayerMode>(ml: &mut MerkleLayer<KV, M>, op: &StepOp) {
        match op {
            StepOp::Set(key, data) => {
                ml.set(key, data).expect("set should succeed");
            }
            StepOp::Delete(key) => {
                ml.delete(key).expect("delete should succeed");
            }
            StepOp::Write(key, data) => {
                ml.write(key, 0, data).expect("write should succeed");
            }
            // TODO: add `get` to step op
        }
    }
```

(which, really becomes 'two steps' for the write at that point)

NSant215: lost diff but done now - this change highlighted a bug where writing a new Node would fail the test, as the tree itself wouldn't be accessed, and therefore would be blinded in the proof, and wouldn't be able to later be made present.

durable-storage/src/avl/tree.rs
```diff
@@ -47,6 +48,16 @@ use crate::storage::Loadable;
 use crate::storage::Storable;
 use crate::storage::StoreOptions;
 
+/// Hash of the empty AVL tree (`Tree::<NodeId>(None)`).
+/// The empty-tree hash is independent of the `NodeId` type parameter.
+static EMPTY_TREE_HASH: LazyLock<Hash> = LazyLock::new(|| {
+    Hash::from([
```
emturner: this shouldn't be hardcoded - this can be calculated directly

durable-storage/src/merkle_layer.rs
emturner: fixed offline

durable-storage/src/avl/tree.rs
```diff
@@ -888,6 +896,16 @@ mod tests {
...
+    fn empty_tree_hash_constant_matches_default() {
+        assert_eq!(
+            *EMPTY_TREE_HASH,
+            Hash::from_foldable(&Tree::<LazyNodeId>(None)),
+            "EMPTY_TREE_HASH literal is stale — recompute it from \
+             `Hash::from_foldable(&Tree::<LazyNodeId>(None))` and update the byte array.",
+        );
+    }
+
     #[test]
```
kurtisc: Not needed now

durable-storage/src/merkle_layer.rs
```diff
@@ -83,6 +89,37 @@ impl<KV> MerkleLayer<KV, Normal> {
...
+    ///
+    /// The proof is retained so [`MerkleLayer::hash`] can resolve `prev_hash` substitutions for
+    /// blinded subtrees that fold to [`PartialHash::Previous`].
+    #[cfg_attr(
+        not(test),
+        expect(
+            dead_code,
+            reason = "non-test callers wired in follow-up PVM integration (RV-957)"
+        )
+    )]
```
kurtisc: There is already a test module in this file, you may find it cleaner to simply move this function into it, then move it out after RV-957

durable-storage/src/merkle_layer.rs
```diff
@@ -250,12 +287,22 @@ impl MerkleLayerMode for Verify {
...
                 resolver: VerifyResolver,
+                proof: this.inner.proof.clone(),
             },
         }
     }
 
-    fn hash<KV>(_this: &MerkleLayer<KV, Self>) -> Hash {
-        unimplemented!("Blocked by RV-961")
+    fn hash<KV>(this: &MerkleLayer<KV, Self>) -> Hash {
+        // The wrapping `original_proof: None` is intentional: it lets each per-node sub-proof
```
kurtisc: ~~So make it explicit?~~

Should be removed now

durable-storage/src/merkle_layer.rs
```diff
@@ -400,15 +450,20 @@ impl<KV> MerkleLayer<KV, Verify> {
...
 }
 
+/// Construct a verify-mode [`MerkleLayer`] for an empty initial state. Used in tests that exercise
+/// verify-mode mutation primitives in isolation.
+#[cfg(test)]
+pub(crate) fn empty_verify_layer<KV>() -> MerkleLayer<KV, Verify> {
+    let proof = MerkleProof::from_foldable(&Tree::<ProveNodeId>::default());
+    MerkleLayer::from_proof(proof).expect("The proof should be deserialisable")
+}
+
```
kurtisc, line 608: Move to module below

durable-storage/src/merkle_layer.rs
```diff
@@ -400,15 +450,20 @@ impl<KV> MerkleLayer<KV, Verify> {
     }
 }
 
+/// Construct a verify-mode [`MerkleLayer`] for an empty initial state. Used in tests that exercise
+/// verify-mode mutation primitives in isolation.
+#[cfg(test)]
+pub(crate) fn empty_verify_layer<KV>() -> MerkleLayer<KV, Verify> {
```
kurtisc: `new_verify_layer` for consistency with the other helpers?

durable-storage/src/avl/tree.rs
kurtisc: Would it be possible to swap `get` to something that returns the ID, then have another method on the ID to get the data?

durable-storage/src/merkle_layer.rs
```diff
@@ -1798,4 +2052,216 @@ mod tests {
...
+    /// 2. The prove-mode final hash matches Normal-mode after an identical operation (prove-mode
+    ///    mutations are faithful).
+    /// 3. Replaying the same operation against the Verify-mode tree decoded from the proof
+    ///    produces the same final hash as prove mode (the proof carries enough information for
+    ///    full verification).
+    fn assert_prove_mode_correct<KV: KeyValueStore + TestKeyValueStoreSetup>(
+        setup_keys: &[[u8; 2]],
+        step_ops: &[StepOp],
+    ) {
+        // ---- Normal: build initial tree ----
```
kurtisc: Humanise pls

durable-storage/src/merkle_layer.rs
```diff
@@ -2003,4 +2052,216 @@ mod tests {
...
+            }
+            StepOp::Delete(key) => {
+                ml.delete(key).expect("delete should succeed");
+            }
+            StepOp::Write(key, data) => {
+                ml.write(key, 0, data).expect("write should succeed");
+            } // TODO RV-895: add `get` to step op
+        }
+    }
+    /// Core assertion: prove-mode proof and hash are consistent with Normal mode and the proof
```
kurtisc: Formatting error. Also, it is preferable (to me) for the helper functions to be located at the top of the tests module with the others, not interspersed with tests

durable-storage/src/merkle_layer.rs
```diff
@@ -2003,4 +2052,216 @@ mod tests {
...
+                    0 => step_ops.push(StepOp::Set(key, data)),
+                    1 => step_ops.push(StepOp::Delete(key)),
+                    _ => step_ops.push(StepOp::Write(key, data)),
+                }
+            }
+
+            assert_prove_mode_correct::<KV>(&setup_keys, &step_ops);
+        });
+
+        // prove_verify_writes_only (no structural change)
```
kurtisc: Explain why these don't cause structural changes.

The underscore item-like prefixes don't seem to add much to these comments

durable-storage/src/merkle_layer.rs
```diff
@@ -2003,4 +2052,216 @@ mod tests {
...
+        proptest!(|(setup_keys in prop::collection::vec(any::<[u8; 2]>(), 1..30))| {
+            let keys: Vec<Key> = setup_keys
+                .iter()
+                .map(|b| Key::new(b).expect("key should be valid"))
+                .collect();
+
+            let step_ops: Vec<StepOp> = keys
+                .iter()
+                .enumerate()
+                .map(|(i, key)| StepOp::Write(key.clone(), vec![0xAA; 1 + (i % 8)]))
```
kurtisc: Is there any significance to this pattern? Resolve if not

kurtisc (changes requested):

durable-storage/src/avl/tree.rs
kurtisc: This was resolved by Emma, re-resolve if the suggestion doesn't work

durable-storage/src/merkle_layer.rs
```diff
@@ -391,6 +440,9 @@ struct ProveImpl<KV> {
 struct VerifyImpl {
     tree: Tree<VerifyNodeId>,
     resolver: VerifyResolver,
+    /// Original proof the verify-mode tree was deserialised from. Required by [`MerkleLayer::hash`]
+    /// to resolve `prev_hash` substitutions in [`PartialHashFold::previous`] for blinded subtrees.
+    proof: Arc<MerkleProof>,
```
```suggestion
    original_proof: Arc<MerkleProof>,
```

durable-storage/src/merkle_layer.rs
kurtisc: Used elsewhere

durable-storage/src/merkle_layer.rs
```diff
@@ -2005,4 +2054,216 @@ mod tests {
...
+        op: &StepOp,
+    ) {
+        match op {
+            StepOp::Set(key, data) => {
+                ml.set(key, data).expect("set should succeed");
+            }
+            StepOp::Delete(key) => {
+                ml.delete(key).expect("delete should succeed");
+            }
+            StepOp::Write(key, data) => {
```
NSant215: add test for non-zero offset writes

durable-storage/src/avl/tree.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/bb108e9..ca8368c62fcc638f4f917a843499d0bba809fed1)

durable-storage/src/merkle_layer.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/ca8368c62fcc638f4f917a843499d0bba809fed1..c4b802a893196d56fdec8b55312023fdbe4cf2a4)

durable-storage/src/avl/tree.rs
NSant215: lost diff but gone I think

durable-storage/src/merkle_layer.rs
NSant215: lost diff but think I have done now

durable-storage/src/merkle_layer.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/aa3f044efd392e4870cad164da863fda899c5e71..f59ac40769316a5f244a00783160264c3792f099)

durable-storage/src/merkle_layer.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/859ac24edb6fdedec014332d4bfe5722cc3f20e5..21a6a1827a2f76c693423552fd37d37bf5025acd)

durable-storage/src/merkle_layer.rs
NSant215: I find the separation useful? Can you elaborate XD

durable-storage/src/merkle_layer.rs
NSant215: fixed [here](https://github.com/tezos/riscv-pvm/compare/27f7ac2d2a061d877050de1010103705389e90a6..03da16a609d3abf9046755f7e11749dcb2981937)

durable-storage/src/merkle_layer.rs
NSant215: moved (lost diff)

durable-storage/src/merkle_layer.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/03da16a609d3abf9046755f7e11749dcb2981937..5f4f0bb1b7cb4289ddcd8c892ed138267bddfb81)

emturner (approved):

durable-storage/src/avl/node.rs
```diff
@@ -134,20 +143,21 @@ impl<TreeId, M: Mode> Node<TreeId, M> {
...
     where
         NodeId: 'a,
         TreeId: 'a,
+        M: 'a,
     {
         loop {
             let resolved_node = resolver.resolve(node)?;
             match resolved_node.key().cmp(key) {
-                Ordering::Equal => return Ok(Some(&resolved_node.data)),
+                Ordering::Equal => return Ok(Some(node)),
```
kurtisc, line 160: One more thing, please break out these changes (and anything else that cascades) into their own commit

kurtisc (approved):

durable-storage/src/avl/node.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/25ef8e42a4fc7cc07fa6832017b5a92651fb1298..d3e6305e269eb4cb40a3071ae2508f0bad22b9ed)

PR #1041, opened 2026-05-04
refactor(durable): pass data as generic DataId in avl tree

emturner:
- Part of [RV-955](https://linear.app/tezos/issue/RV-955)Part of [RV-955](https://linear.app/tezos/issue/RV-955)

# What

Refactor the avl `Node` to be generic over the `DataId`. When needing access to the underlying `Bytes`, it must be resolved.

# Why

This is a first step towards lazy resolution of Bytes. This will allow `delete` to function when the node's data has already been deleted from the KV store (as `delete` will no longer load/resolve the data).

This _should_ also make the database more performant when it's been recently checked out - as intermediate values won't be loaded from the database even when the nodes are resolved (turning two roundtrips to the DB to one)

# How

For now, resolution does nothing: the bytes are still eagerly loaded. Subsequent MR will introduce `LazyBytes`, and use it alongside the `LazyResolver`.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/resolver.rs
+++ b/durable-storage/src/avl/resolver.rs
@@ -60,6 +60,7 @@ use super::node::Node;
 use super::tree::Tree;
 use crate::avl::node::Meta;
 use crate::errors::OperationalError;
+use crate::key::Key;
 use crate::storage::KeyValueStore;
 use crate::storage::Loadable;
 use crate::storage::Storable;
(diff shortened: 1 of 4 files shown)


victor-dumitrescu (approved):

durable-storage/src/avl/resolver.rs
```diff
@@ -702,8 +775,10 @@ impl<R> ProveResolver<R> {
     }
 }
 
-impl<R: Resolver<LazyNodeId, Node<LazyTreeId, Normal>> + Resolver<LazyTreeId, Tree<LazyNodeId>>>
-    Resolver<ProveNodeId, ProveNode> for ProveResolver<R>
+impl<
+    R: Resolver<LazyNodeId, Node<LazyTreeId, Bytes<Normal>, Normal>>
```
thomasathorne, line 780: Could put these constraints into a `where` clause for readability?

thomasathorne (approved):

NSant215 (approved):

durable-storage/src/avl/resolver.rs
emturner: done - see [diff](https://github.com/tezos/riscv-pvm/compare/9cedbac88874639d097a5074f9f2ed95659abd5c..5fa2e10d0e4997867ae69f63dcef1b2ba93248ee)

PR #1042, opened 2026-05-04
test(durable): do not mix insertion/updates in multi-step proof tests

emturner:
- Related to RV-961

# What

Prevent accidentally including 'update' operations in the multi-step 'insertion only' prove-verify roundtrip test

# Why

Mixing the operation kinds is not currently supported in multi-step mode

# How

Filter out 'new keys' that are part of the setup keys. Just use a simple linear scan as we're checking very few keys overall.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/merkle_layer.rs
+++ b/durable-storage/src/merkle_layer.rs
@@ -2258,6 +2258,10 @@ mod tests {
         proptest!(|(setup_keys in prop::collection::vec(any::<[u8; 2]>(), 1..20), new_keys in prop::collection::vec(any::<[u8; 2]>(), 1..10))| {
             let step_ops: Vec<Operation> = new_keys
                 .iter()
+                // ensure we don't accidentally mix insertions with overwritting old keys
+                // - the proof system doesn't currently support this
+                // TODO (RV-985): remove this line
+                .filter(|bytes| !setup_keys.contains(bytes))
                 .enumerate()
                 .map(|(i, bytes)| {
                     let key = Key::new(bytes).expect("key should be valid");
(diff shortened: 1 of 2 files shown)


victor-dumitrescu (approved):
Maybe worth checking in this regression in `durable-storage/proptest-regressions/merkle_layer.txt` to make sure this case is exercised when this is fixed:

```
# Seeds for failure cases proptest has generated in the past. It is
# automatically read and these particular cases re-run before any
# novel cases are generated.
#
# It is recommended to check this file in to source control so that
# everyone who runs the test benefits from these saved cases.
cc 6f3816de0c63e15b8d291d0dab0e77f3a6e6cc72bee54754a26a670e0ffd7acf # shrinks to setup_keys = [[174, 0], [173, 202], [0, 0], [0, 1], [0, 2]], new_keys = [[0, 3], [173, 202]]
```

durable-storage/src/merkle_layer.rs
```diff
@@ -2258,6 +2258,10 @@ mod tests {
         proptest!(|(setup_keys in prop::collection::vec(any::<[u8; 2]>(), 1..20), new_keys in prop::collection::vec(any::<[u8; 2]>(), 1..10))| {
             let step_ops: Vec<Operation> = new_keys
                 .iter()
+                // ensure we don't accidentally mix insertions with overwritting old keys
+                // - the proof system doesn't currently support this
+                // TODO (RV-985): remove this line
+                .filter(|bytes| !setup_keys.contains(&bytes))
```
NSant215:
```suggestion
                .filter(|bytes| !setup_keys.contains(bytes))
```

NSant215 (approved):

durable-storage/src/merkle_layer.rs
emturner: thanks! done [diff](https://github.com/tezos/riscv-pvm/compare/385e0898291dded004a5db87f86966dfefd4dca9..10533c1f6820fb24f4642e3c25e7aeeec1b07f4d)

PR #1047, opened 2026-05-06
fix(durable): deletion immediately after checkout crashes merkle worker

emturner:
- closes RV-955
- closes RV-956

# What

Fix a crash in the merkle worker, that would occur when deleting a value immediately after a checkout.

# Why

Operations in the database happen first in the persistence layer, before being sent to the merkle layer. This is to ensure that we only send succeeding operations to the merkle layer.

The database should be able to handle deletes.

# How

The issue occurred for delete, however, where first the value would be deleted out the persistence layer, and only then would the deletion be sent to the merkle layer. Since the relevant node had not yet been resolved, however, the merkle layer would attempt to load the data from the persistence layer, which was now missing!

Instead, we switch to use a `LazyDataId` wrapper over the bytes. For now, we still attempt to eagerly resolve the data, but if it fails for `NotFound` we don't crash the worker. Instead, we crash the worker on any subsequent attempt to read the bytes. 

Since, for now, we are addressing the main case of not found for a node that will be immediately deleted, this is okay.  There is still the race condition captured by [RV-987](https://linear.app/tezos/issue/RV-987/race-condition-in-database-between-merklelayer-and-persistencelayer) - which will address further cases where this can occur. 

This will lead us to making `LazyDataId` fully lazy (only resolving when absolutely needed) - but it does add further complexity around `impl Storeable for Node`, and `into_proof` etc. 

This MR represents a good-enough stopping point to address an immediate bug, without adding too much more complexity - which can then be reviewed on its own subsequently.

# Manually Testing

```
make all
```
checkout the first commit, and run the following to see the new deletion test does in fact fail before the fix is applied:
```
cargo nextest run -p octez-riscv-durable-storage --run-ignored only
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -55,6 +55,7 @@ pub(crate) struct Meta {
 #[derive(Encode, Decode)]
 struct StoredNode {
     meta: Meta,
+    data: Hash,
     left: Hash,
     right: Hash,
 }
@@ -849,6 +850,7 @@ impl<TreeId: Storable> Storable for Node<TreeId, Bytes<Normal>, Normal> {
         // should be written to the KV store separately.
         let repr = StoredNode {
             meta: self.meta.deref().clone(),
+            data: Hash::from_foldable(&self.data),
             left: Hash::from_foldable(&self.left),
             right: Hash::from_foldable(&self.right),
         };
@@ -875,7 +877,12 @@ impl<TreeId: Storable> Storable for Node<TreeId, Bytes<Normal>, Normal> {
 
 impl<TreeId: Loadable> Loadable for Node<TreeId, Bytes<Normal>, Normal> {
     fn load(id: Hash, store: &impl KeyValueStore) -> Result<Self, OperationalError> {
-        let StoredNode { meta, left, right } = {
+        let StoredNode {
+            meta,
+            left,
+            right,
+            data: _,
+        } = {
             let bytes =
                 store
                     .blob_get(id)
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -29,12 +29,14 @@ use octez_riscv_data::serialisation::deserialise;
 use octez_riscv_data::serialisation::serialise;
 use perfect_derive::perfect_derive;
 
+use super::resolver::LazyDataId;
 use super::resolver::LazyTreeId;
 use super::resolver::ProveNode;
 use super::resolver::TreeResolver;
 use super::tree::Tree;
 use crate::avl::resolver::AvlResolver;
 use crate::errors::Error;
+use crate::errors::InvalidArgumentError;
 use crate::errors::OperationalError;
 use crate::key::Key;
 use crate::storage::KeyValueStore;
@@ -75,7 +77,7 @@ pub struct Node<TreeId, DataId, M: Mode> {
     hash: OnceLock<Hash>,
 }
 
-impl Node<LazyTreeId, Bytes<Normal>, Normal> {
+impl Node<LazyTreeId, LazyDataId, Normal> {
     /// Converts the [`Node`] to prove mode.
     pub fn into_proof(self) -> ProveNode {
         Node {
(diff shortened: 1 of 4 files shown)


durable-storage/src/avl/node.rs
```diff
@@ -913,6 +934,59 @@ impl<TreeId: Loadable> Loadable for Node<TreeId, Bytes<Normal>, Normal> {
     }
 }
 
+impl<TreeId: Loadable> Loadable for Node<TreeId, LazyDataId, Normal> {
```
NSant215: this implementation is very similar to `Bytes<Normal>` - do we need both / could we use a helper on the code?

NSant215 (approved):

durable-storage/src/database.rs
```diff
@@ -657,6 +657,35 @@ mod tests {
         assert_eq!(before, after);
     });
 
+    kv_test!(test_database_delete_after_checkout, KV: BackgroundPersistentKeyValueStore, {
```
victor-dumitrescu, line 731: Maybe add a comment saying this test is to verify the fix for RV-955?

durable-storage/src/avl/node.rs
```diff
@@ -920,6 +934,59 @@ impl<TreeId: Loadable> Loadable for Node<TreeId, Bytes<Normal>, Normal> {
...
+        //
+        // TODO (RV-998): go all in on laziness
+        let data = match store.get(meta.key.as_ref()) {
+            Ok(bytes) => {
+                let bytes = Bytes::from(bytes.as_ref());
+                LazyDataId::from(bytes)
+            }
+            Err(Error::InvalidArgument(InvalidArgumentError::KeyNotFound)) => {
+                LazyDataId::from(data)
+            }
```
victor-dumitrescu: This case is purely to avoid the crash, right? As in, we don't fail here, but we'll never be able to read this data. If that's the case perhaps worth adding a comment?

durable-storage/src/avl/resolver.rs
```diff
@@ -335,6 +337,132 @@ impl Loadable for LazyNodeId {
...
+
+impl From<Bytes<Normal>> for LazyDataId {
+    fn from(value: Bytes<Normal>) -> Self {
+        Self(LazyId::new(value))
+    }
+}
+
+impl From<Hash> for LazyDataId {
+    fn from(value: Hash) -> Self {
+        LazyDataId(LazyId::from(value))
```
victor-dumitrescu:
```suggestion
        Self(LazyId::from(value))
```

victor-dumitrescu (approved):

durable-storage/src/avl/node.rs
emturner: I'll add a comment 👍🏻 

Technically you could still take the `hash` at this point, but reading the actual data would indeed fail.

durable-storage/src/avl/resolver.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/8b089eb086b00a988166baf3aa4fcd6049d299a4..ae509f281c3b45a490e9d9cbd93c8549f4724d2b)

durable-storage/src/avl/node.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/8b089eb086b00a988166baf3aa4fcd6049d299a4..ae509f281c3b45a490e9d9cbd93c8549f4724d2b)

durable-storage/src/database.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/8b089eb086b00a988166baf3aa4fcd6049d299a4..ae509f281c3b45a490e9d9cbd93c8549f4724d2b)

durable-storage/src/avl/node.rs
emturner: I pushed a diff [here](https://github.com/tezos/riscv-pvm/compare/ae509f281c3b45a490e9d9cbd93c8549f4724d2b..1a6a13cceb38031e5a38c23bddb3c0743b071393). Let me know what you think?

(the diff is a little messed up, as the `DataLoadable` trait is put right at the bottom of `Node`. I did this as I plan to add more 'Datatrait' abstractions in the follow up for the data race fix)

NSant215 (approved):

PR #1049, opened 2026-05-07
test(data): benchmark that searches for biggest proof

thomasathorne:
Closes RV-989.

# What

Add a benchmark that generates random valid proofs on a bytes component (of 64MiB) and prints the largest proof it finds.

# Why

This benchmark itself isn't really very useful, for proof size we'll mostly calculate the maximum we expect analytically and use a proptest instead to check no proofs are bigger.

This PR introduces the functions that we can use to search for maximally bad proofs on other metrics, such as verification time or memory allocations. Follow-up PRs will add those benchmarks too. We really just use `proof_size` as a simple example of an evaluation function.

# How

There are two commits. The first is a very mechanistic move of some code from a test module into a new `test_utils` module so that it can be accessed from the benchmark code. The second commit is where all the new stuff is.

# Manually Testing

To run the benchmark:

```
cargo bench --bench proof
```

# Tasks for the Author

### DIFF
--- a/data/Cargo.toml
+++ b/data/Cargo.toml
@@ -30,3 +30,9 @@ optional = true
 [dev-dependencies]
 proptest.workspace = true
 rand.workspace = true
+humansize.workspace = true
+
+[[bench]]
+name = "proof"
+harness = false
+required-features = ["unstable-test-utils"]
--- a/data/benches/proof.rs
+++ b/data/benches/proof.rs
@@ -1,0 +1,85 @@
+// SPDX-FileCopyrightText: 2026 TriliTech <contact@trili.tech>
+//
+// SPDX-License-Identifier: MIT
+
+use humansize::BINARY;
+use humansize::format_size;
+use octez_riscv_data::components::bytes::Bytes;
+use octez_riscv_data::components::bytes::test_utils::BytesMutOp;
+use octez_riscv_data::merkle_proof::proof_tree::MerkleProof;
+use octez_riscv_data::mode::Normal;
+use octez_riscv_data::mode::Provable;
+use octez_riscv_data::serialisation::serialise;
+use proptest::prelude::Strategy;
+use proptest::strategy::ValueTree;
+use proptest::test_runner::TestRunner;
+
+/// 64 MiB: the maximum size of a `Bytes` component in the durable storage
+const LENGTH: usize = 1024 * 1024 * 64;
+
+/// Given a `strategy` to generate instances of type `A` and an `eval` function to evaluate those
+/// according to some metric. We always make at least one attempt, followed by some number of
+/// `repeats`, to find the worst case, returning the example with the worst outcome we found.
+///
+/// Another function `setup` is provided to create some background state once instead of on every
+/// repeat, in case this provides a performance benefit.
+fn find_worst<A, State, Metric: Ord>(
+    strategy: impl Strategy<Value = A>,
+    setup: impl Fn() -> State,
+    eval: impl Fn(&State, &A) -> Metric,
+    repeats: usize,
+) -> (A, Metric) {
+    let mut runner = TestRunner::default();
+
+    let state = setup();
+
+    let mut gen_a = || {
+        strategy
+            .new_tree(&mut runner)
+            .expect("can use `strategy` to generate values")
+            .current()
+    };
+    let mut max_a = gen_a();
+    let mut max_eval = eval(&state, &max_a);
+    for _ in 0..repeats {
+        let a = gen_a();
+        let ev = eval(&state, &a);
+
+        if ev > max_eval {
+            max_a = a;
+            max_eval = ev;
+        }
+    }
+
+    (max_a, max_eval)
+}
+
+/// Initialise a `Bytes<Normal>` component for generating proofs over.
+fn init_state() -> Bytes<Normal> {
+    let v = [1u8, 2, 3, 4, 5]
+        .iter()
+        .copied()
+        .cycle()
+        .take(LENGTH)
+        .collect::<Vec<_>>();
+    Bytes::from(&v[..])
+}
+
+/// Produce the proof for a given bytes component and operation.
+fn produce_proof(state: &Bytes<Normal>, op: &BytesMutOp) -> Vec<u8> {
+    let mut bytes_prove = state.start_proof();
+    let _result = op.run(&mut bytes_prove);
+
+    let proof_tree = MerkleProof::from_foldable(&bytes_prove);
+    serialise(proof_tree).expect("can serialise proof")
+}
+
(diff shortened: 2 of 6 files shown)


NSant215 (approved):

data/Cargo.toml
```diff
@@ -19,7 +19,7 @@ memmap2.workspace = true
 range-collections.workspace = true
 bytes.workspace = true
 unty.workspace = true
+proptest.workspace = true
```
emturner: ideally this would not become a non `dev-dependency`

data/src/components/bytes/tests.rs
```diff
@@ -391,83 +386,6 @@ mode_test!(append_adds_to_end, F, {
...
-pub(crate) enum BytesOpResult {
-    Read { read: usize, data: Vec<u8> },
-    Wrote { wrote: usize },
-    Len { len: usize },
-    Void,
-}
-
 // Bytes behaves the same across different modes
 #[test]
 fn bytes_are_same_across_modes() {
```
emturner, line 392: so - technically these tests do not need to be inside the `bytes` module - they could just as well be integration tests.

Could we instead, move the test utils to be under the external tests? (With the view of eventually adding `octez-riscv-data-test-utils`, which could be used from everywhere?).

data/benches/proof.rs
```diff
@@ -0,0 +1,85 @@
...
+use octez_riscv_data::merkle_proof::proof_tree::MerkleProof;
+use octez_riscv_data::mode::Normal;
+use octez_riscv_data::mode::Provable;
+use octez_riscv_data::serialisation::serialise;
+use proptest::prelude::Strategy;
+use proptest::strategy::ValueTree;
+use proptest::test_runner::TestRunner;
+
+/// 1024 * 1024 * 64, or 64 MiB: the maximum size of a `Bytes` component in the durable storage
+const LENGTH: usize = 67108864;
```
emturner:
```suggestion
const LENGTH: usize = 64 * 1024 * 1024
```

nit: just inline the multiplication? when used as a `const`, any overflow should be a compilation error.

data/benches/proof.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let mut gen_a = || {
+        strategy
+            .new_tree(&mut runner)
+            .expect("can use `strategy` to generate values")
+            .current()
+    };
+    let mut max_a = gen_a();
+    let mut max_eval = eval(&state, &max_a);
+    for _ in 0..(attempts - 1) {
```
emturner:
```suggestion
for _ in 0..attempts
```

I think? otherwise e.g. 1 attempt would be `0..0` which would attempt zero times

data/benches/proof.rs
```diff
@@ -0,0 +1,85 @@
...
+    serialise(proof_tree).expect("can serialise proof")
+}
+
+/// An evaluation function that measures the length of the proof for a given `op`.
+fn proof_size(state: &Bytes<Normal>, op: &BytesMutOp) -> usize {
+    produce_proof(state, op).len()
+}
+
+fn main() {
+    let (worst_op, eval) = find_worst(BytesMutOp::any(LENGTH), init_state, proof_size, 1000);
```
emturner, line 83: could you make `run_attempts` configurable by environment variable?

data/benches/proof.rs
victor-dumitrescu: Seems passing 1 still does 1 attempt, but passing 0 would hang.

data/benches/proof.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/// Initialise a `Bytes<Normal>` component for generating proofs over.
+fn init_state() -> Bytes<Normal> {
+    let v = [1u8, 2, 3, 4, 5]
+        .iter()
+        .copied()
+        .cycle()
+        .take(LENGTH)
+        .collect::<Vec<_>>();
```
victor-dumitrescu, line 64: Is there a benefit to generating the starting state deterministically? I don't think the values of the individual bytes should influence any of the metrics we care about. If that is the case though, I'd argue generating random bytes from a set seed would be better.

data/benches/proof.rs
thomasathorne: Done [here](https://github.com/tezos/riscv-pvm/compare/f44dc6dc2f8b9319129b0f95dc23c7f4d1d326b2..2f780fbb92a4f47b6352dbc6f89d8f635190fc7f)

data/benches/proof.rs
thomasathorne: The 'extra' attempt is the one in the two lines above, we do that outside of the loop to initialise the two `max` mutable variables.

data/benches/proof.rs
thomasathorne: I would like to make the benchmark take a number of different command line arguments, I will add a follow-up issue for this.

data/benches/proof.rs
thomasathorne: I see Victor's point, passing zero leads to different behaviour to what you'd expect: I've fixed that [here](https://github.com/tezos/riscv-pvm/compare/2f780fbb92a4f47b6352dbc6f89d8f635190fc7f..1015bb7ee5f275363edc94e6f94c47d4542c067e).

data/benches/proof.rs
thomasathorne: I should certainly try randomly generated starting states as well, but for this PR I wanted to keep it the same because it's a bit of a performance hit generating a new 64MiB state for every attempt.

data/Cargo.toml
thomasathorne: As discussed offline I will take the first commit of this PR and expand it into a new PR that solves this.

data/src/components/bytes/tests.rs
thomasathorne: As discussed offline I will take the first commit of this PR and expand it into a new PR that solves this.

data/benches/proof.rs
thomasathorne: Actually decided a simpler approach would be [this](https://github.com/tezos/riscv-pvm/compare/1015bb7ee5f275363edc94e6f94c47d4542c067e..0874b9ed8a2703448ced704546ac4d4b435aa259).

victor-dumitrescu (approved):

data/Cargo.toml
thomasathorne: New PR: https://github.com/tezos/riscv-pvm/pull/1058

This PR now points to that branch.

data/src/components/bytes/tests.rs
thomasathorne: See comment above for new PR.

emturner (approved):

PR #1050, opened 2026-05-07
test(data): benchmark that searches for slowest proof

thomasathorne:
Closes RV-990

# What

A benchmark similar to https://github.com/tezos/riscv-pvm/pull/1049 that reports the slowest (in verification) proof it can find.

# Why

We want to explore the limits of proof space for this metric because if proofs can be very slow to verify that is a big issue---verification happens in L1 nodes.

This PR is only exploring a rather limited set of proofs---the valid and reachable ones. For this metric we need to explore invalid proofs too but that will happen in follow-up work.

# Manually Testing

```
cargo bench --bench proof
```

# Tasks for the Author

### DIFF
--- a/data/benches/proof.rs
+++ b/data/benches/proof.rs
@@ -2,10 +2,16 @@
 //
 // SPDX-License-Identifier: MIT
 
+use std::hint::black_box;
+use std::time::Duration;
+use std::time::Instant;
+
 use humansize::BINARY;
 use humansize::format_size;
 use octez_riscv_data::components::bytes::Bytes;
 use octez_riscv_data::components::bytes::test_utils::BytesMutOp;
+use octez_riscv_data::hash::PartialHash;
+use octez_riscv_data::merkle_proof::proof_binary;
 use octez_riscv_data::merkle_proof::proof_tree::MerkleProof;
 use octez_riscv_data::mode::Normal;
 use octez_riscv_data::mode::Provable;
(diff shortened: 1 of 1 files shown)


data/benches/proof.rs
```diff
@@ -79,7 +84,35 @@ fn proof_size(state: &Bytes<Normal>, op: &BytesMutOp) -> usize {
...
+    let _after_verify_hash =
+        PartialHash::from_foldable(parsed_proof_tree.clone(), &bytes_verify).to_hash();
+}
+
+/// An evaluation function that measures the time taken by the proof verification.
+fn proof_time(state: &Bytes<Normal>, op: &BytesMutOp) -> Duration {
+    let proof = produce_proof(state, op);
+
+    let start = Instant::now();
+    verify_proof(&proof, op);
```
emturner: you'll likely need a `black_box` here - otherwise there's a good chance some of the above is optimised out

emturner: also - possibly using criterion to some extent would be useful: should probably take at least a few samples to avoid one thing being way slower than the others?

emturner: it's okay if it's long running

thomasathorne: Good point with `black_box`, added that [here](https://github.com/tezos/riscv-pvm/compare/3687a76ad39e0aedeaabd446f2382c6bb7e2b9b8..139dbc165c38059adf7f2215ba8c827e8f23684c).

I'll have a look at integrating criterion.

thomasathorne: It seems fiddly to integrate with criterion without using `criterion_main!` (which is definitely how criterion is intended to be used). Instead I've implemented a simple sampling loop that takes 10 samples (after one warm-up) and disposes of the min and max before taking an average, see [here](https://github.com/tezos/riscv-pvm/compare/139dbc165c38059adf7f2215ba8c827e8f23684c..9743e683b8f199e6047b1b203c5c894b663502ad).

emturner:
Tried locally - saw that at it's slowest was a massive resize. 

Could possibly add a TODO/follow up: resizing should only happen by up to `2KiB` at a time? 

(though I suspect this won't make things _all_ that faster)

emturner (approved):

victor-dumitrescu (approved):
LGTM. Some comments about the benchmarks more broadly:
- The command to run them is now `cargo bench --bench proof --features unstable-test-utils`, perhaps worth adding some Makefile support for it?
- Benches are never run in CI. Once they're configurable, it would perhaps be good to run them with smaller parameters at least to make sure they don't become stale.

thomasathorne:
@victor-dumitrescu Thanks for those suggestions, have added them in RV-1003 as I think they make sense there.

PR #1051, opened 2026-05-07
test(data): benchmark that searches for most allocation-hungry proof

thomasathorne:
Closes RV-991

# What

A set of three benchmarks similar to https://github.com/tezos/riscv-pvm/pull/1049 that report the worst proofs they can find in various metrics related to allocation:

- total number of allocations
- total number of bytes allocated
- single largest allocation.

# Why

We want to explore the limits of proof space for these metrics because if proof verification uses a lot of memory that is a big issue---verification happens in L1 nodes.

This PR is only exploring a rather limited set of proofs---the valid and reachable ones. For this metric we need to explore invalid proofs too but that will happen in follow-up work.

# Manually Testing

```
cargo bench --bench proof
```

# Tasks for the Author

### DIFF
--- a/data/benches/proof.rs
+++ b/data/benches/proof.rs
@@ -2,7 +2,12 @@
 //
 // SPDX-License-Identifier: MIT
 
+use std::alloc::GlobalAlloc;
+use std::alloc::Layout;
+use std::alloc::System;
 use std::hint::black_box;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
 use std::time::Duration;
 use std::time::Instant;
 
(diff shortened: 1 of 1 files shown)


emturner (commented):
rather than reporting 'the most allocs' (or the most alloc bytes) - instead could report `greatest allocs - deallocs` / `greatest allocs_bytes - deallocs_bytes`? (ie more about memory pressure rather than raw allocations?)

data/benches/proof.rs
```diff
@@ -122,10 +218,56 @@ fn proof_time(state: &Bytes<Normal>, op: &BytesMutOp) -> Duration {
     durations.into_iter().skip(1).take(8).sum::<Duration>() / 8
 }
 
+/// An evaluation function that counts the total number of allocations made by the proof
+/// verification.
+fn proof_allocs(state: &Bytes<Normal>, op: &BytesMutOp) -> usize {
+    let proof = produce_proof(state, op);
+
+    GLOB_ALLOC.reset();
+    verify_proof(&proof, op);
```
victor-dumitrescu, line 268: `verify_proof` is also cloning `parsed_proof_tree` twice, which has a bearing on allocations. I _think_ this doesn't impact the search for the most allocation-intensive proof, but the absolute values for allocations / bytes allocated will be overestimated.

thomasathorne:
@emturner followed your suggestion of adding 'maximum simultaneous' alloc metrics [here](https://github.com/tezos/riscv-pvm/compare/4f788b2d90114ac4d0347f6cdd557836da1baacb..36ae18dbc68f6d44f008446b255a78e142cc1a27). That diff also includes a few more doc-strings on the `Counter` struct.

data/benches/proof.rs
thomasathorne: Good point, I've removed the final `clone` [here](https://github.com/tezos/riscv-pvm/compare/748e17cdd78ddfd77aa20b697d95b2a0b5dc73a4..a7a1919587b59986b5a9a5c048cacdb1dde812c0). I think the earlier one can't be avoided (and would be needed in production anyway). 🤔

Testing this the different is actually rather negligible, for a write operation I just tried it is 10 allocations totally less than 10 KiB.

victor-dumitrescu (approved):

emturner (approved):

PR #1052, opened 2026-05-07
Add `kv_test!`-based equivalent for registry integration test

victor-dumitrescu:
Closes RV-977.

# What

Adds `test_durable_storage_end_to_end` as a `kv_test!` for the registry which replicates the durable storage integration test.

# Why

To have a property-based test which generates `Registry` operations and runs across different configurations. Next, execution will produce traces which will be compared for consistency.

# How

Most of the helpers and strategies defined in `integration_test.rs` are moved to a new source module `test_helpers.rs`. This module is enabled in test and with a new `test-utils` feature flag (`#[cfg(test)]` is not enough since they need to be exposed in the crate API in order for the integration test to use it).

In addition to just moving:
- `test_durable_storage_inner` is renamed to `run_operations` and can now run with both RocksDB and in-memory backends
- the core of `test_durable_storage` (the `Vec<OperationView>` to `Vec<OperationView>` mapping) is taken out and renamed to `make_operations`

I opted to keep the integration test as is rather than replace with the `kv_test` because the integration test exercises the public API directly, whereas the new `kv_test!` will in the future go through a test-only `Registry` layer for recording traces.

Some of the moved declarations are now public but behind a test-only feature flag and largely self-explanatory so I didn't add any extra documentation.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/Cargo.toml
+++ b/durable-storage/Cargo.toml
@@ -9,6 +9,7 @@ workspace = true
 
 [features]
 default = ["rocksdb"]
+unstable-test-utils = ["dep:proptest"]
 
 [dependencies]
 bincode.workspace = true
@@ -31,9 +32,14 @@ trait-set.workspace = true
 workspace = true
 optional = true
 
+[dependencies.proptest]
+workspace = true
+optional = true
+
 [dev-dependencies]
 proptest.workspace = true
 criterion.workspace = true
+octez-riscv-durable-storage = { path = ".", features = ["unstable-test-utils"] }
 octez-riscv-test-utils.workspace = true
 serde.workspace = true
 serde_json.workspace = true
--- a/durable-storage/src/lib.rs
+++ b/durable-storage/src/lib.rs
@@ -39,3 +39,4 @@ pub mod persistence_layer;
 pub mod registry;
 pub mod repo;
 pub mod storage;
+pub mod test_helpers;
--- a/durable-storage/src/registry.rs
+++ b/durable-storage/src/registry.rs
@@ -1440,6 +1440,19 @@ pub(super) mod tests {
 
         assert!(registry.resize_tick(5).is_err());
     });
+
+    kv_test!(test_durable_storage_end_to_end, KV: BackgroundPersistentKeyValueStore,
+    [
+        generated in crate::test_helpers::operations_strategy(1usize..100)
+    ],
+    {
+        // Every test iteration expects an empty repo, so not setting it in a `setup` block
+        let (_keepalive, repo) = KV::setup_repo();
+
+        let (keys, values, ops) = generated;
+        let operations = crate::test_helpers::make_operations(keys, values, ops);
+        crate::test_helpers::run_operations::<KV>(repo, operations)
+    });
 }
 
 #[cfg(feature = "rocksdb")]
--- a/durable-storage/src/storage/in_memory.rs
+++ b/durable-storage/src/storage/in_memory.rs
@@ -21,10 +21,10 @@ use crate::errors::OperationalError;
 /// Will never write to disk.
 #[derive(Debug, Default, Clone)]
 pub struct InMemoryRepo {
-    #[cfg(test)]
+    #[cfg(any(test, feature = "unstable-test-utils"))]
     commits: std::sync::Arc<RwLock<HashMap<crate::commit::CommitId, InMemorySnapshot>>>,
 
-    #[cfg(test)]
+    #[cfg(any(test, feature = "unstable-test-utils"))]
     registry_commits: std::sync::Arc<RwLock<HashMap<crate::commit::CommitId, Vec<u8>>>>,
 }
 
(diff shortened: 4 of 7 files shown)


durable-storage/Cargo.toml
```diff
@@ -31,13 +32,21 @@ trait-set.workspace = true
...
 proptest.workspace = true
 criterion.workspace = true
 octez-riscv-test-utils.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 
+[[test]]
+name = "integration_test"
+required-features = ["test-utils"]
+
```
kurtisc: I mostly run durable storage tests from inside the subdirectory, using `cargo test`. From what I can tell, this change causes the integration test to be skipped unless the `test-utils` feature is specified. I don't like the ergonomics of doing that, or gating the feature through Make, and we had big issues around December with tests being merged that failed when run with certain features.

An idiom to work around this is to add this crate as a dev-dependency of itself, with `test-utils` enabled. e.g.:

```
diff --git i/durable-storage/Cargo.toml w/durable-storage/Cargo.toml
index bbae0e9dc5..4f1a281fe6 100644
--- i/durable-storage/Cargo.toml
+++ w/durable-storage/Cargo.toml
@@ -39,14 +39,11 @@ optional = true
 [dev-dependencies]
 proptest.workspace = true
 criterion.workspace = true
+octez-riscv-durable-storage = { path = ".", features = ["test-utils"] }
 octez-riscv-test-utils.workspace = true
 serde.workspace = true
 serde_json.workspace = true

-[[test]]
-name = "integration_test"
-required-features = ["test-utils"]
-
 [[bench]]
 name = "avl_tree"
 harness = false
```

kurtisc (changes requested):
PSA this is much easier to review locally with [histogram](https://adamj.eu/tech/2024/01/18/git-improve-diff-histogram/)

@victor-dumitrescu you've missed one of the check boxes in the PR.

durable-storage/Cargo.toml
victor-dumitrescu: The self-dependency is much nicer indeed, applied in latest version. Thanks!

victor-dumitrescu:
> PSA this is much easier to review locally with [histogram](https://adamj.eu/tech/2024/01/18/git-improve-diff-histogram/)
> 
> @victor-dumitrescu you've missed one of the check boxes in the PR.

I've left the documentation box unchecked on purpose since by moving the integration test helpers to a non-test module I "added" new undocumented public items. I don't think adding documentation for these would be particularly useful, but I can do it if reviewers disagree.

durable-storage/src/test_helpers.rs
```diff
@@ -0,0 +1,497 @@
+// SPDX-FileCopyrightText: 2026 Trilitech <contact@trili.tech>
```
emturner, line 1: (first commit) - does this module need to be added to `lib.rs`?

durable-storage/src/registry.rs
```diff
@@ -836,6 +836,19 @@ pub(super) mod tests {
             Err(Error::Operational(OperationalError::RegistryCommitMismatch))
         ));
     });
+
+    kv_test!(test_durable_storage_end_to_end, KV: BackgroundPersistentKeyValueStore,
+    [
+        generated in crate::test_helpers::operations_strategy(1usize..100)
+    ],
+    {
+        // Every test iteration expects an empty repo, so not setting it in a `setup` block
```
emturner, line 1449: alternatively: could every test iteration begin with `db.clear()`? / `resgistry.resize(0)`?

durable-storage/Cargo.toml
```diff
@@ -9,6 +9,7 @@ workspace = true
 
 [features]
 default = ["rocksdb"]
+test-utils = ["dep:proptest"]
```
emturner:
```suggestion
unstable-test-utils = ["dep:proptest"]
```

could you do this instead? Marks the feature flag as explicitly unstable for the purposes of semver etc

durable-storage/src/lib.rs
```diff
@@ -39,3 +39,6 @@ pub mod persistence_layer;
 pub mod registry;
 pub mod repo;
 pub mod storage;
+
+#[cfg(any(test, feature = "test-utils"))]
```
emturner: could you move the cfg into `test_helpers`? I think we're tending to go with that elsewhere too

durable-storage/src/storage/in_memory.rs
```diff
@@ -21,10 +21,10 @@ use crate::errors::OperationalError;
 /// Will never write to disk.
 #[derive(Debug, Default, Clone)]
 pub struct InMemoryRepo {
-    #[cfg(test)]
+    #[cfg(any(test, feature = "test-utils"))]
```
emturner: possible follow up: it might be nice to have a `build-test-utils` to emit a cfg helper to simplify this more easily

(see, for example in the KernelSDK - which allows devs to do `cfg(pvm_kind = "riscv")`

durable-storage/src/registry.rs
victor-dumitrescu: I think the issue here is that the repo preserves all registry commits. We do start each test with a fresh registry, but previous commits are still there, which results in failures when we expect the checkout of a commit which shouldn't exist to fail, but it doesn't.

durable-storage/Cargo.toml
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/d2b2b751db0f733b175232e8a46bca8791816294..5101cdbb693597967683fff0731fb89b986bb61e).

durable-storage/src/lib.rs
victor-dumitrescu: Done

durable-storage/src/test_helpers.rs
victor-dumitrescu: I'm now adding it to `lib.rs` in the first commit

durable-storage/src/storage/in_memory.rs
victor-dumitrescu: I created [RV-1002](https://linear.app/tezos/issue/RV-1002/build-configuration-for-tests-is-generated-via-build-script) for this.

kurtisc (approved):

emturner (approved):

PR #1053, opened 2026-05-07
Modal `Registry`

kurtisc:
Closes RV-966
Closes RV-967

# What
Implements `Verify` and `Prove` for the Registry

# Why
This finishes the basic milestone of implementing modality for durable storage

# How
 - Seals  `RegistryMode`
 - Implements `RegistryMode` for each mode

# Manually Testing

```
make all
```

# Regressions

# Tasks for the Author

### DIFF
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -570,7 +570,7 @@ impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
 mod traced_database;
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use std::collections::HashSet;
     use std::sync::Arc;
 
@@ -641,6 +641,27 @@ mod tests {
         let database = new_database::<KV>(handle, &repo);
 
         (runtime, keepalive, repo, database)
+    }
+
+    /// Wrap an existing Prove-mode [`MerkleLayer`] into a [`Database`].
+    pub(crate) fn from_prove_layer<KV: KeyValueStore>(
+        merkle: MerkleLayer<KV, Prove<'static>>,
+    ) -> Database<KV, Prove<'static>> {
+        Database {
+            inner: super::ProveImpl { merkle },
+        }
+    }
+
+    /// Generate a Merkle proof from `database` and produce a Verify-mode database that replays
+    /// the same data via that proof.
+    pub(crate) fn to_verify<KV: KeyValueStore>(
+        database: &Database<KV, Prove<'static>>,
+    ) -> Database<KV, Verify> {
+        Database {
+            inner: super::VerifyImpl {
+                merkle: database.inner.merkle.to_verify(),
+            },
+        }
     }
 
     kv_test!(test_database_commit_and_checkout, KV: BackgroundPersistentKeyValueStore,
@@ -2038,7 +2059,7 @@ mod tests {
 
         // Generate the proof and produce a Verify-mode database from it.
         let (prove_inner, prove_trace) = prove_db.into_parts();
-        let verify_ml: MerkleLayer<KV, Verify> = prove_inner.inner.merkle.into_verify();
+        let verify_ml: MerkleLayer<KV, Verify> = prove_inner.inner.merkle.to_verify();
         let verify_db: Database<KV, Verify> = Database {
             inner: super::VerifyImpl { merkle: verify_ml },
         };
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -24,6 +24,7 @@ use octez_riscv_data::mode::Mode;
 use octez_riscv_data::mode::Normal;
 use octez_riscv_data::mode::Prove;
 use octez_riscv_data::mode::Verify;
+use perfect_derive::perfect_derive;
 use tezos_smart_rollup_constants::core::MAX_FILE_CHUNK_SIZE;
 use tokio::runtime::Handle;
 
@@ -52,6 +53,7 @@ pub const MAX_VALUE_SIZE: usize =
 ///
 /// This structure unifies the key-value store and Merkle layer to allow for persistent storage
 /// alongside a representation which can provide a root hash.
+#[perfect_derive(Clone)]
 #[repr(transparent)]
 pub struct Database<KV, M: Mode> {
     inner: M::Select<DatabaseTemplate<KV>>,
@@ -370,7 +372,19 @@ struct NormalImpl<KV> {
     merkle: MerkleWorker<KV>,
 }
 
+impl<KV> Database<KV, Verify> {
+    /// An empty verify-mode database.
+    pub(crate) fn empty() -> Self {
+        Database {
+            inner: VerifyImpl {
+                merkle: <MerkleLayer<KV, Verify>>::empty(),
+            },
+        }
+    }
+}
+
 /// Verify-mode implementation for the [`Database`].
+#[perfect_derive(Clone)]
 struct VerifyImpl<KV> {
     merkle: MerkleLayer<KV, Verify>,
 }
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -372,6 +372,32 @@ struct NormalImpl<KV> {
     merkle: MerkleWorker<KV>,
 }
 
+impl<KV> Database<KV, Prove<'static>> {
+    /// An empty prove-mode database backed by the given persistence layer.
+    pub(crate) fn empty(persistence: Arc<KV>) -> Self
+    where
+        KV: KeyValueStore,
+    {
+        Database {
+            inner: ProveImpl {
+                merkle: <MerkleLayer<KV, Prove<'static>>>::empty(persistence),
+            },
+        }
+    }
+
+    /// Clone the database, attaching the new clone to a fresh persistence layer.
+    pub(crate) fn clone_with(&self, persistence: Arc<KV>) -> Self
+    where
+        KV: KeyValueStore,
+    {
+        Database {
+            inner: ProveImpl {
+                merkle: self.inner.merkle.clone_with(persistence),
+            },
+        }
+    }
+}
+
 impl<KV> Database<KV, Verify> {
     /// An empty verify-mode database.
     pub(crate) fn empty() -> Self {
--- a/durable-storage/src/merkle_layer.rs
+++ b/durable-storage/src/merkle_layer.rs
@@ -758,8 +758,8 @@ mod tests {
 
         /// Generate a Merkle proof from the current Prove-mode layer and produce a Verify-mode
         /// layer that replays the same data via that proof.
-        pub(crate) fn into_verify(self) -> MerkleLayer<KV, Verify> {
-            let proof = MerkleProof::from_foldable(&self);
+        pub(crate) fn to_verify(&self) -> MerkleLayer<KV, Verify> {
+            let proof = MerkleProof::from_foldable(self);
             MerkleLayer::from_proof(proof).expect("proof deserialization should succeed")
         }
     }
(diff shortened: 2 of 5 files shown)


durable-storage/src/registry.rs
```diff
@@ -195,8 +195,20 @@ impl<KV: KeyValueStore, M: Mode> Registry<KV, M> {
             _ => return Err(Error::from(InvalidArgumentError::RegistryResizeTooLarge)),
         }
```
NSant215, line 198: random questions about 1st commit message just for curiosity:
- why is it important to assert that only modes in the Registry module can implement RegistryMode? are there other places where we leak modes i.e. is this not a given? and same for RegistryTemplate privacy.

durable-storage/src/registry.rs
```diff
@@ -312,40 +324,41 @@ impl<KV: KeyValueStore> Modal for RegistryTemplate<KV> {
     type Verify = Infallible;
 }
 
+/// Sealing module for [`RegistryMode`].
+#[doc(hidden)]
+pub mod sealed {
```
```suggestion
mod sealed {
```

tried locally, doesn't need to be `pub`

(it pretty much has to be, otherwise it's not actually sealing anything)

durable-storage/src/registry.rs
```diff
@@ -312,40 +324,41 @@ impl<KV: KeyValueStore> Modal for RegistryTemplate<KV> {
...
+    /// Marker trait that prevents external types from implementing [`super::RegistryMode`].
+    pub trait Sealed {}
+}
+
 /// Modes that implement this support operations on [`Registry`]
-#[expect(
-    private_interfaces,
-    reason = "This method should not be used outside of this module"
-)]
```
emturner, line 319:
```suggestion
#[expect(
    private_interfaces,
    reason = "This method should not be used outside of this module"
)]
```
This should not be removed - the methods are still reachable outside of this module.

If you want to remove them, you likely want the approach outlined in [RV-986](https://linear.app/tezos/issue/RV-986/modal-components-leak-private-interface)

durable-storage/src/registry.rs
```diff
@@ -195,8 +195,20 @@ impl<KV: KeyValueStore, M: Mode> Registry<KV, M> {
...
+            None
+        };
+
         self.databases
-            .try_resize_with(new_size, || M::try_new_database(&self.inner))?;
+            .try_resize_with::<OperationalError>(new_size, || {
+                Ok(new_db
+                    .take()
+                    .expect("growth-by-1 should call this closure at most once"))
+            })?;
```
emturner: this feels the modal interface as is isn't quite helping right now? 

Given that you made the change to `Database` rather than `inner` - I suspect you'd want to move `try_resize` into the `RegistryMode` itself. (rather than `try_new_db`). (And then have `Registry::resize` still check at most one dB to be added/removed)

Or, just leave as is (given that we still need `expect(private-interface)` with the code in the current state).  With `M::Select` in the trait.

emturner (commented):
I'm leaning towards not doing the sealing and or changes to the `RegistryMode` trait in this MR - it needs more work, and we should probably look further at whether to change what we pass: `inner` or `this`

durable-storage/src/registry.rs
```diff
@@ -312,40 +324,41 @@ impl<KV: KeyValueStore> Modal for RegistryTemplate<KV> {
...
+    pub trait Sealed {}
+}
+
 /// Modes that implement this support operations on [`Registry`]
-#[expect(
-    private_interfaces,
-    reason = "This method should not be used outside of this module"
-)]
-pub trait RegistryMode: Mode {
+pub trait RegistryMode: Mode + sealed::Sealed {
```
NSant215: if this is public then can't anything also implement it?

durable-storage/src/merkle_layer.rs
```diff
@@ -96,6 +96,22 @@ impl<KV> MerkleLayer<KV, Normal> {
     }
 }
 
+impl<KV> MerkleLayer<KV, Prove<'static>> {
+    /// An empty prove-mode Merkle layer backed by the given persistence layer.
+    pub(crate) fn empty(persistence: Arc<KV>) -> Self
```
NSant215, line 101: what is this used for?

durable-storage/src/database.rs
```diff
@@ -372,6 +372,32 @@ struct NormalImpl<KV> {
     merkle: MerkleWorker<KV>,
 }
 
+impl<KV> Database<KV, Prove<'static>> {
+    /// An empty prove-mode database backed by the given persistence layer.
+    pub(crate) fn empty(persistence: Arc<KV>) -> Self
```
NSant215, line 377: what is this used for also?

durable-storage/src/registry.rs
```diff
@@ -894,6 +943,191 @@ pub(super) mod tests {
...
+            .copy_database(0, 1)
+            .expect("Copying should succeed");
+
+        registry
+            .database(1)
+            .expect("Database should exist.")
+            .assert_database_value(&key, b"src");
+    });
+
+    kv_test!(test_prove_database_clone_independence, KV: BackgroundKeyValueStore, {
```
NSant215, line 993: useful for multi-step proofs down the line - but not necessarily needed now? happy to keep but just validating understanding

durable-storage/src/registry.rs
```diff
@@ -894,6 +943,191 @@ pub(super) mod tests {
...
+        while registry.len() > 1 {
+            registry
+                .resize_tick(registry.len() - 1)
+                .expect("Shrinking the registry should succeed.");
+        }
+        assert_eq!(registry.len(), 1);
+
+        assert!(registry.resize_tick(5).is_err());
+    });
+
```
NSant215, line 1126: we probably need roundtrip tests at the registry level?

durable-storage/src/registry.rs
```diff
@@ -1119,6 +1123,78 @@ pub(super) mod tests {
         assert!(registry.resize_tick(5).is_err());
     });
 
+    kv_test!(test_roundtrip_prove_to_verify, KV: BackgroundKeyValueStore, {
+        // Reading through a `Registry<Prove>` populated from snapshots of Normal-mode source
+        // data records a proof; replaying the same reads through the resulting
+        // `Registry<Verify>` must yield the same values.
```
thomasathorne: I'd put this comment above the `kv_test!` call so it's clearer that it is describing the whole test.

I also tend to agree with Neelay's comment: this isn't enough to justify the name 'roundtrip', it seems fine for this PR but I'd maybe change the name and ensure there's follow up issues to test this component with a proptest over generated registry operations.

thomasathorne (approved):

durable-storage/src/registry.rs
kurtisc: Removed now

durable-storage/src/registry.rs
kurtisc: No longer applies, sealed traits removed

durable-storage/src/registry.rs
kurtisc: Sealed traits removed

durable-storage/src/registry.rs
kurtisc: Sealed traits removed

durable-storage/src/registry.rs
kurtisc: Sealed traits removed

durable-storage/src/merkle_layer.rs
kurtisc: [Here](https://github.com/tezos/riscv-pvm/pull/1053/changes#diff-8bf2f243fe3c2d0f52f4a51de2d5f17b7175a92445f2d9909e787eae97875afbR363)

durable-storage/src/database.rs
kurtisc: [Here](https://github.com/tezos/riscv-pvm/pull/1053/changes#diff-8bf2f243fe3c2d0f52f4a51de2d5f17b7175a92445f2d9909e787eae97875afbR363)

durable-storage/src/merkle_layer.rs
kurtisc: Inside the Database one below

durable-storage/src/registry.rs
kurtisc: Yes but something that should already be enforced going forward IMO

durable-storage/src/registry.rs
kurtisc: Moved and renamed [here](https://github.com/tezos/riscv-pvm/compare/5b4757d1421e63f856359fda706de0141918fc10..343584df677c1dbd96451dd5cc34efe35cf3e293)

durable-storage/src/registry.rs
kurtisc: There's a new test [here](https://github.com/tezos/riscv-pvm/pull/1053/changes/343584df677c1dbd96451dd5cc34efe35cf3e293) but I suppose generated instructions would be a good follow up?

durable-storage/src/registry.rs
```diff
@@ -836,6 +939,442 @@ pub(super) mod tests {
...
+
+        let len = prove_registry.len();
+        let mut verify_databases = Vec::with_capacity(len);
+
+        for i in 0..len {
+            let placeholder = <Database<KV, Prove<'static>>>::empty(Arc::new(
+                KV::new(&prove_registry.inner.repo)
+                    .expect("Persistence creation should succeed"),
+            ));
+            let prove_db = std::mem::replace(&mut prove_registry.databases[i], placeholder);
```
emturner: wold this work if you just generate a proof and then do `from_proof` instead?

emturner (approved):

durable-storage/src/registry.rs
```diff
@@ -1119,6 +1125,79 @@ pub(super) mod tests {
...
+        normal_b
+            .set(&key_b, b"bar")
+            .expect("Setting in Normal mode should succeed");
+
+        let prove_databases = vec![
+            from_prove_layer::<KV>(normal_a.start_proof()),
+            from_prove_layer::<KV>(normal_b.start_proof()),
+        ];
+
+        let mut prove_registry = Registry::<KV, Prove<'static>> {
```
NSant215: is it possible to compare the initial and final hashes of the registry in each mode in this test?

NSant215 (approved):

durable-storage/src/database.rs
```diff
@@ -370,7 +372,45 @@ struct NormalImpl<KV> {
...
+    {
+        Database {
+            inner: ProveImpl {
+                merkle: <MerkleLayer<KV, Prove<'static>>>::empty(persistence),
+            },
+        }
+    }
+
+    /// Clone the database, attaching the new clone to a fresh persistence layer.
+    pub(crate) fn try_clone_with(&self, persistence: Arc<KV>) -> Self
```
victor-dumitrescu: Since this isn't fallible, this should perhaps be called `clone_with`? Same goes for the `MerkleLayer` equivalent.

victor-dumitrescu (approved):

durable-storage/src/database.rs
kurtisc: [clone_with](https://github.com/tezos/riscv-pvm/pull/1053/changes/1784acb9d69edc9f465136ab34b15378daf866d7)

durable-storage/src/registry.rs
kurtisc: Yes, the helper was already doing it so I've used that [here](https://github.com/tezos/riscv-pvm/compare/679812cb388fa4234bad9ff9e36f5ae9797a5f57..1f099f93e37033b4a07f526e3cd56a93d3d9a84b)

durable-storage/src/registry.rs
kurtisc: [Here](https://github.com/tezos/riscv-pvm/compare/679812cb388fa4234bad9ff9e36f5ae9797a5f57..1f099f93e37033b4a07f526e3cd56a93d3d9a84b#diff-8bf2f243fe3c2d0f52f4a51de2d5f17b7175a92445f2d9909e787eae97875afbR1152) and onwards

NSant215 (approved):

PR #1059, opened 2026-05-13
feat(dursto): Support multi-step proofs for Single Database.

NSant215:
Relates to RV-985.
Part of RV-1004.

# What

Change the mechanism by which deleted nodes are added to the `deleted_nodes` list to a trait method on Resolver.
Expand current tests in Merkle Layer to including multi-step changes. 

# Why

Makes proof system more robust if it can support more complex changes. 

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/resolver.rs
+++ b/durable-storage/src/avl/resolver.rs
@@ -75,6 +75,9 @@ pub trait Resolver<Id, Value> {
 
     /// Resolve an identifier to a mutable value.
     fn resolve_mut<'a>(&mut self, id: &'a mut Id) -> Result<&'a mut Value, OperationalError>;
+
+    /// Add a node about to be deleted to the resolver's tracking.
+    fn track_deleted_node(&mut self, _id: &Id) {}
 }
 
 /// Trait for resolving data identifiers to bytes.
@@ -280,11 +283,10 @@ impl LazyNodeId {
     /// The returned [`ProveNodeId`] keeps the original lazy identifier for hash lookups and
     /// storage-backed resolution, while leaving its prove-mode cache empty until a
     /// [`ProveResolver`] materialises the node.
-    pub(crate) fn into_proof<R>(self, resolver: &ProveResolver<R>) -> ProveNodeId {
+    pub(crate) fn into_proof(self) -> ProveNodeId {
         ProveNodeId {
-            node: self.clone(),
-            inner: OnceLock::new(),
-            deleted_nodes: resolver.deleted_nodes.clone(),
+            lazy_id: Some(self.clone()),
+            cache: OnceLock::new(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


durable-storage/src/avl/tree.rs
```diff
@@ -119,18 +118,22 @@ impl<NodeId> Tree<NodeId> {
...
                 (None, Some(right)) => {
+                    resolver.track_deleted_node(node);
                     *node = right.clone();
                     Ok(true)
                 }
                 (Some(_), Some(_)) => {
+                    resolver.track_deleted_node(node);
                     let (new_node, shrank) = Node::replace_with_successor(node, resolver)?;
                     *node = new_node;
                     Ok(shrank)
```
kurtisc: Why does `track_deleted_node` need to be inside the match arms?

durable-storage/src/merkle_layer.rs
```diff
@@ -801,25 +801,81 @@ mod tests {
...
+                    count: usize,
+                ) -> Result<Vec<u8>, OperationalError> {
+                    let mut buf = vec![0u8; count];
+                    let n = self.get(key)?.map_or(0, |data| data.read(offset, &mut buf));
+                    buf.truncate(n);
+                    Ok(buf)
+                }
+            }
+        };
+    }
```
kurtisc: Do you need this macro because `get` isn't mode-agnostic? I know this is just test code, but maybe implies we should add `MerkleLayerMode::get`?

NSant215: yes that is the reason - I wasn't sure which route made more sense and happy either way, but yes since it would be just test code in either case it seemed that the macro was the easiest

durable-storage/src/avl/tree.rs
NSant215: good point! done [here](https://github.com/tezos/riscv-pvm/compare/c2b0edbca0fa1771448a5535fe2f11e41c4f9c73..9652787c93879f2582813e8f6e4f46a94a591e65)

durable-storage/src/merkle_layer.rs
emturner: I think making `MerkleLayer::get` mode generic would make sense, given there's an implementation for all three modes now (if possible)

emturner (commented):
overall looks good - same comment re `get`

durable-storage/src/merkle_layer.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/74a14799842a317c1ee4c48a5472a67034be1f98..5b4ec9fe5366377676bf575207b17a0526d94cec)

kurtisc: Can this be pulled into a standalone commit? Currently it's in a 'test' commit, but it changes non-test code.

kurtisc (approved):

durable-storage/src/merkle_layer.rs
NSant215: done now :)

kurtisc: The commit message title has a spelling error ;) please resolve when you fix

durable-storage/src/merkle_layer.rs
```suggestion
```

durable-storage/src/merkle_layer.rs
```diff
@@ -2257,23 +2299,52 @@ mod tests {
...
+            .zip(keys.iter())
+            .map(|(bytes, k)| (k.clone(), setup_data_len(bytes)))
+            .collect();
+
         let mut operations = Vec::new();
-        // TODO RV-985: Expand to more operations in each test.
-        // This is currently limited to one operation, as this is the minimum requirement,
-        // and proof semantics are not supported for more than one operation yet.
-        let ops_count = 1;
+        let ops_count = 10;
```
emturner: could you make this also part of the proptest?

emturner (commented):
before marking RV-1004 as done, would it be worth adding the database prove mode also to the integration tests for the database?

(even with a verify round trip too?)

Then we'll have more confidence it's all working correctly.

durable-storage/src/merkle_layer.rs
emturner: (both the filter and the TODO)

durable-storage/src/avl/resolver.rs
```diff
@@ -944,6 +925,23 @@ where
             id.inner.get_mut().expect("inner was just set"),
         ))
     }
+
+    fn track_deleted_node(&mut self, id: &ProveNodeId) {
+        // Nodes that were resolved during the step carry a prove-mode projection that needs
+        // to be available for the `MerkleLayer`-level fold after the step finishes. Non-resolved
+        // nodes are not relevant for the fold.
```
emturner: this tripped me up - could you possibly expand that non-resolved nodes are nodes e.g. inserted during the proof?

durable-storage/src/merkle_layer.rs
NSant215: done [diff](https://github.com/tezos/riscv-pvm/compare/2d6816084e74832d1bfda34d9c2d5a866403f51f..2c6e5ceaa85060b80339e26cf050a192e47e2ce1)

durable-storage/src/merkle_layer.rs
NSant215: like this? [diff](https://github.com/tezos/riscv-pvm/compare/2c6e5ceaa85060b80339e26cf050a192e47e2ce1..e64a6e3ec5cbad8d1717806a9021ad06f3565f4d)

durable-storage/src/avl/resolver.rs
NSant215: is this better? I realised the return should never be hit [diff](https://github.com/tezos/riscv-pvm/compare/8660e0b5cd7d11230e7cc1ad16eeb9b8be59676f..960ac5114aa310198b381388e7280585610fada7)

NSant215:
@emturner I will make a follow-up MR to extend multi-step proof proptests to the merkle layer.

durable-storage/src/avl/resolver.rs
NSant215: renames and changing the lazy id to optional [here](https://github.com/tezos/riscv-pvm/compare/960ac5114aa310198b381388e7280585610fada7..c6ff0c7a8c2fdaa52478897ec76ec85cb52f97c1) after discussion offline

emturner (approved):

PR #1060, opened 2026-05-13
test(data): regression test for largest valid `Bytes` proof

thomasathorne:
Closes RV-997.

# What

Add constants and a function to generate the operations that produce the longest valid proofs for a leaf in the durable storage. These are in `bytes/test_utils` as they will be useful for testing the durable storage more generally.

Add a unit test that checks these proofs are indeed as long as we believe they should be, and a proptest that will fail if it finds a worse proof.

# Why

We want to document the worst case valid proof, and test it for regression purposes.

# Manually Testing

```
make all
```

# Tasks for the Author


data/src/components/bytes/test_utils.rs
```diff
@@ -91,3 +91,51 @@ pub enum BytesOpResult {
...
     Void,
 }
+
+/// 64 MiB: the maximum size of a `Bytes` component in the durable storage
+pub const NDS_BYTES_LENGTH: usize = 1024 * 1024 * 64;
+
+/// The calculated theoretical maximum length of a proof for a single operation on a leaf in the
+/// durable storage. It is a read or write that crosses the worst possible boundary of two pages.
+///
+/// Page size is 4096 bytes, plus 8 for a `u64` representing the length of the page; there are two
```
emturner: you might be able to expose the page size from the bytes component here

data/src/components/bytes/test_utils.rs
```diff
@@ -91,3 +91,51 @@ pub enum BytesOpResult {
...
+/// such pages in the proof.
```
emturner:
```suggestion
/// such pages in the proof - because of the maximum read/write size?
```

data/src/components/bytes/test_utils.rs
```diff
@@ -91,3 +91,51 @@ pub enum BytesOpResult {
...
+/// other.
+///
+/// There is then one more layer with only two blinded nodes (64 bytes) and four tags (node, node,
+/// blind, blind).
+///
+/// Finally there are 8 bytes for the overall length of the `Bytes` component (serialised as a
+/// `u64`) and an extra three tags for the three nodes (length, tree, and root).
+///
+/// This all adds up to give 9487 bytes.
+pub const MAX_PROOF_LENGTH: usize = (4096 + 8) * 2 + (192 + 8) * 6 + (64 + 4) + 8 + 2 + 1;
```
emturner: it might be helpful to break these out into intermediate constants etc?

data/src/components/bytes/test_utils.rs
```diff
@@ -91,3 +91,51 @@ pub enum BytesOpResult {
...
+/// 64 MiB: the maximum size of a `Bytes` component in the durable storage
+pub const NDS_BYTES_LENGTH: usize = 1024 * 1024 * 64;
+
+/// The calculated theoretical maximum length of a proof for a single operation on a leaf in the
+/// durable storage. It is a read or write that crosses the worst possible boundary of two pages.
+///
+/// Page size is 4096 bytes, plus 8 for a `u64` representing the length of the page; there are two
+/// such pages in the proof.
+///
+/// There are then 6 identical layers of the tree, each with 6 blinded nodes (6 * 32 = 192) and 8
```
emturner: could you explain 6? (assume its the log)

data/src/components/bytes/test_utils.rs
thomasathorne: Used `PAGE_SIZE` [here](https://github.com/tezos/riscv-pvm/compare/70b6bfb330c8f08bf0741f5c935cf78b8fb23817..cb977d20a0d18db088d8ca3b89756191bd1cd737).

data/src/components/bytes/test_utils.rs
thomasathorne: See [here](https://github.com/tezos/riscv-pvm/compare/cb977d20a0d18db088d8ca3b89756191bd1cd737..6db8126a82c22e5d5e9a65a88b0738572a925812)

data/src/components/bytes/test_utils.rs
thomasathorne: Ok, done [here](https://github.com/tezos/riscv-pvm/compare/6db8126a82c22e5d5e9a65a88b0738572a925812..a9f8c22ec95517aaa65928c2d1a127c0fffb1eee)

data/src/components/bytes/test_utils.rs
thomasathorne: Yes, it's one less than the log (base 4) of the number of pages. This is laid out pretty clearly in one of the intermediate constants I introduced in the diff above. Let me know if you think I should add another comment explaining it though.

data/src/components/bytes/test_utils.rs
```diff
@@ -91,3 +94,80 @@ pub enum BytesOpResult {
     Len { len: usize },
     Void,
 }
+
+/// 64 MiB: the maximum size of a `Bytes` component in the durable storage
+pub const NDS_BYTES_LENGTH: usize = 1024 * 1024 * 64;
+
+/// The depth of the (data part of the) merkle tree for a maximal durable storage leaf. This should
```
kurtisc: Merkle should be capitalised ;)

data/src/components/bytes/test_utils.rs
```diff
@@ -91,3 +94,80 @@ pub enum BytesOpResult {
...
+
+/// Sanity check that the constant values calculated above have not changed.
+#[test]
+fn const_values() {
+    assert_eq!(NDS_BYTES_LENGTH, 67108864);
+    assert_eq!(NDS_BYTES_TREE_DEPTH, 7);
+    assert_eq!(MONOFURCATED_LAYER, 68);
+    assert_eq!(BIFURCATED_LAYER, 200);
+    assert_eq!(MAX_PROOF_LENGTH, 9487);
+}
```
kurtisc, line 185: Could you do this with goldenfile?

kurtisc (approved):

data/src/components/bytes/test_utils.rs
thomasathorne: It seems slightly odd to me to use goldenfile for what will be just a few characters in each case. 🤔

What's the argument for goldenfile over inline values? It's not a big deal, I will happily change it if people generally prefer that. 😁

data/src/components/bytes/test_utils.rs
thomasathorne: Ah... 🤦‍♂️

Done [here](https://github.com/tezos/riscv-pvm/compare/35b3cca593e3ba9db45e016d1d58409e329badff..d34d683eb4cb5e870e9011fbb25f908fbad4d5bf)

data/src/components/bytes/test_utils.rs
kurtisc: I think having a 'static' test feels odd. It depends on if you expect these ever to change. If you think they'd be updated (without the constant names being changed) in the future then the ergonomics of goldenfile are more straightforward.

You can also move the check to build time with [this trick](https://github.com/tezos/riscv-pvm/pull/62/changes#diff-c0656d3081a3cd2dcda31be427ed813c730aa6af7b274a0389534c85d609e3e7R1048) (but it will fail only with `cargo build`, not `cargo check`, because the assertion isn't checked until monomorphisation).

But feel free to resolve

thomasathorne: I've switched to using goldenfile [here](https://github.com/tezos/riscv-pvm/compare/d34d683eb4cb5e870e9011fbb25f908fbad4d5bf..7b088d3ad8088dc21321b8a63078e8ca5cfd25eb)

emturner (approved):

PR #1061, opened 2026-05-14
feat(durable): add `Registry::try_start_proof`

emturner:
- Closes TZX-133
- Part of TZX-113

# What

Expose `Registry::try_start_proof` for converting a normal mode registry into a provable one

# Why

Required as part of the integration into octez

# How

We add a new extended variant of `Provable`: `ProvableExt` - which the database, and the vector used by the registry, both implement. 

This is done this way, to avoid having to update all `Provable` instances in the codebase at all, they all work with `ProvableExt` out the box, when it's needed.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/data/src/mode.rs
+++ b/data/src/mode.rs
@@ -124,3 +124,24 @@ pub trait Provable<'normal> {
     /// Start proving.
     fn start_proof(&'normal self) -> Self::Prover;
 }
+
+/// Types that implement this can be switched to [`Prove`] mode,
+/// but this operation may fail.
+///
+/// Unlike `Provable`, `Prover` may have a lifetime potentially
+/// decoupled from `'normal` - such as `'static`.
+pub trait ProvableExt<'normal, 'prover, E> {
+    /// Variant of [`Self`] in [`Prove`] mode
+    type Prover: 'prover;
+
+    /// Attempt to start proving.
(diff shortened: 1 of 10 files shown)


data/src/mode.rs
```diff
@@ -124,3 +124,24 @@ pub trait Provable<'normal> {
     /// Start proving.
     fn start_proof(&'normal self) -> Self::Prover;
 }
+
+/// Types that implement this can be switched to [`Prove`] mode,
+/// but this operation may fail.
+///
+/// Unlike `Provable`, `Prover` may have a lifetime potentially
+/// decoupled from `'normal` - such as `'static`.
+pub trait ProvableExt<'normal, 'prover, E> {
```
kurtisc, line 133: What does Ext mean, just extension?

NSant215 (approved):

data/src/mode.rs
emturner: Yes, exactly.
I wasn't completely sure what to call it - but out of all the patterns it does, for me, feel closest to [extension traits](https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html)

Very open to suggestions though!

kurtisc (approved):

PR #1062, opened 2026-05-14
test(data): improve UX for proof benchmark, adding CLAP args

thomasathorne:
Closes RV-1003.
Closes RV-988.

# What

Add command line arguments to the benchmark for exploring proofs. Add a (short) run of the benchmark in CI to prevent it getting stale.

# Why

At some point we'll want to run different configurations of this benchmark in more or less long-running CI tasks. It's also useful to be able to set it running locally with a very large number of attempts, or just run one out of the seven different metrics, etc.

# Manually Testing

E.g.

```
cargo bench --bench proof --features='unstable-test-utils' -- --metric verifytime --attempts 2000
```

# Tasks for the Author

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,9 @@ jobs:
             make-target: test
             title: Linux Tests
           - runs-on: ubuntu-latest
+            make-target: bench
+            title: Linux Component Benches
+          - runs-on: ubuntu-latest
             make-target: codecov.json
             code-coverage: true
             title: Code Coverage
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,17 @@ rustfmt-check:
 
 test: jstz/test etherlink/test durable/test cargo-nextest-run cargo-test-doc
 
+bench:
+	@cargo bench --bench proof --features='unstable-test-utils' -- \
+		--metric length \
+		--metric verifytime \
+		--metric verifyallocs \
+		--metric verifybytesalloc \
+		--metric verifymaxalloc \
+		--metric verifymaxrollingallocs \
+		--metric verifymaxbytesalloc \
+		--attempts 200
+
 test-deps: dummy/build page-cache-tester/build dummy-no-std/build
 
 codecov.json: test-deps
@@ -106,4 +117,4 @@ docs/%:
 	@make -C docs ${@:docs/%=%}
 
 # Mark all non-pattern targets as phony to make sure they're always executed
-.PHONY: all build-deps build-deps-slim check check-nix check-format taplo-check-format rustfmt-check codecov.json audit build test test-deps clean cargo-nextest-run cargo-test-doc cargo-clean
+.PHONY: all build-deps build-deps-slim check check-nix check-format taplo-check-format rustfmt-check codecov.json audit build test test-deps clean cargo-nextest-run cargo-test-doc cargo-clean bench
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -53,6 +53,17 @@ rustfmt-check:
 
 test: jstz/test etherlink/test durable/test cargo-nextest-run cargo-test-doc
 
+bench:
+	@cargo bench --bench proof --features='unstable-test-utils' -- \
+	        --metric length \
+		--metric verifytime \
```
emturner:
```suggestion
	    --metric length \
		--metric verifytime \
```

data/Cargo.toml
```diff
@@ -23,10 +22,18 @@ range-collections.workspace = true
...
+features = ["from_str"]
+
 [dependencies.proptest]
 workspace = true
 optional = true
 
+[dependencies.clap]
+workspace = true
+optional = true
+
```
emturner:
```suggestion
[dev-dependencies.clap]
workspace = true
```
(can move it down into the `dev-dependencies` list below

I think? this should always be activated as part of the benchmarks

.github/workflows/ci.yml
```diff
@@ -41,7 +41,7 @@ jobs:
             make-target: check check-nix build audit
             title: Linux Checks
           - runs-on: ubuntu-latest
-            make-target: test
+            make-target: test bench
```
emturner: this might be worth adding into a separate job?  Reasoning, `bench` will build the whole thing with a different profile - so the time spent building would more or less be doubled I suspect.

In future, we'll want to look at stabilising these numbers on the dedicated bench machine, perhaps

Makefile
emturner: looking at the diff - there might be a mismatch of tabs vs spaces on the `--metric length` line?

.github/workflows/ci.yml
thomasathorne: I've split it out into a separate job [here](https://github.com/tezos/riscv-pvm/compare/dc82af181d8a6897f3212e51191bd4e8317e25f9..067dbd0bc8d55e7204e4dff3d81eee94175bd8f7)

Makefile
thomasathorne: Fixed [here](https://github.com/tezos/riscv-pvm/compare/067dbd0bc8d55e7204e4dff3d81eee94175bd8f7..8092a85115310b4436b6c4b43d15f5e62df8f270)

data/Cargo.toml
thomasathorne: This is a good point, not sure why I didn't put this in `dev-dependencies`. 🤔

Fixed it [here](https://github.com/tezos/riscv-pvm/compare/8092a85115310b4436b6c4b43d15f5e62df8f270..abff9f0151546b1f889f3179d4be6d4c13c83cda)

.github/workflows/ci.yml
```diff
@@ -43,6 +43,9 @@ jobs:
           - runs-on: ubuntu-latest
             make-target: test
             title: Linux Tests
+          - runs-on: ubuntu-latest
+            make-target: bench
+            title: Linux Bench
```
emturner: nit - could possibly adapt the benches to be clear this is component benchmarks? As opposed to the pvm throughput one
```suggestion
            title: Linux Component Benches
```

emturner (approved):

.github/workflows/ci.yml
thomasathorne: Ah, true that is a different job... fixed [here](https://github.com/tezos/riscv-pvm/compare/abff9f0151546b1f889f3179d4be6d4c13c83cda..a66d2495ced67d6570eb78dd389a353887157b2a)

victor-dumitrescu (approved):

PR #1065, opened 2026-05-18
test(dursto): add end to end traced PBT for Database

victor-dumitrescu:
Closes RV-978.

# What

Introduces an end-to-end property-based test for `Database` based on a strategy for `Database` operations and comparing traces across storage backends.

# Why

To gain confidence in the `Database` implementation and ensure consistent semantics irrespective of state backend.

# How

The addition of the `test_database_end_to_end` test is largely based on the already introduced `test_durable_storage_end_to_end` for the `Registry`. To this end, the database-specific parts of the registry operations strategy are split off in order to be used in the context of a single database, either `Database` directly (as part of a `Registry`) or `TracedDatabase` as part of the newly introduced test.

The existing `Registry` strategies and tests should be unchanged.

Note to reviewers: The diff for `durable-storage/src/test_helpers.rs` is not super helpful and obscures the fact that the implementation of `apply_database_operation` is largely based on the previous version of `run_operation`.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -587,6 +587,11 @@ impl<KV: BackgroundKeyValueStore, M: DatabaseMode> Database<KV, M> {
 mod traced_database;
 
 #[cfg(test)]
+pub(crate) use self::traced_database::Trace;
+#[cfg(test)]
+pub(crate) use self::traced_database::TracedDatabase;
+
+#[cfg(test)]
 pub(crate) mod tests {
     use std::collections::HashSet;
     use std::sync::Arc;
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -2229,4 +2229,19 @@ pub(crate) mod tests {
 
         database.into_trace()
     });
+
+    kv_test!(test_database_end_to_end, KV: BackgroundPersistentKeyValueStore,
+    [
+        generated in crate::test_helpers::database_operations_strategy(1usize..100)
+    ],
+    {
+        // Every test iteration expects an empty repo, so not setting it in a `setup` block.
+        // This is because a repo preserves database commits from previous test runs, resulting
+        // in test failures when checking out a commit which isn't expected to exist succeeds.
+        let (_keepalive, repo) = KV::setup_repo();
+
+        let (keys, values, ops) = generated;
+        let operations = crate::test_helpers::make_database_operations(keys, values, ops);
+        crate::test_helpers::run_database_operations::<KV>(&repo, operations)
+    });
 }
--- a/durable-storage/src/registry.rs
+++ b/durable-storage/src/registry.rs
@@ -75,6 +75,12 @@ impl<KV: BackgroundKeyValueStore> Registry<KV, Normal> {
             .map_err(|error| OperationalError::WorkerRuntimeCreationFailed { error })?;
         Ok(Arc::new(runtime))
     }
+
+    /// Get a [`Handle`] to the registry's runtime.
+    #[cfg(any(test, feature = "unstable-test-utils"))]
+    pub(crate) fn handle(&self) -> &tokio::runtime::Handle {
+        self.inner.runtime.handle()
+    }
 }
 
 impl<'normal, KV> ProvableExt<'normal, 'static, OperationalError> for Registry<KV, Normal>
@@ -1443,14 +1449,16 @@ pub(super) mod tests {
 
     kv_test!(test_durable_storage_end_to_end, KV: BackgroundPersistentKeyValueStore,
     [
-        generated in crate::test_helpers::operations_strategy(1usize..100)
+        generated in crate::test_helpers::registry_operations_strategy(1usize..100)
     ],
     {
-        // Every test iteration expects an empty repo, so not setting it in a `setup` block
+        // Every test iteration expects an empty repo, so not setting it in a `setup` block.
+        // This is because a repo preserves registry commits from previous test runs, resulting
+        // in test failures when checking out a commit which isn't expected to exist succeeds.
         let (_keepalive, repo) = KV::setup_repo();
 
         let (keys, values, ops) = generated;
-        let operations = crate::test_helpers::make_operations(keys, values, ops);
+        let operations = crate::test_helpers::make_registry_operations(keys, values, ops);
         crate::test_helpers::run_operations::<KV>(repo, operations)
     });
 }
--- a/durable-storage/src/test_helpers.rs
+++ b/durable-storage/src/test_helpers.rs
@@ -7,8 +7,8 @@
 
 //! Shared utilities for end to end durable storage property-based tests
 //!
-//! Used by the integration test in `tests/integration_test.rs` and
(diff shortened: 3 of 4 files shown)


durable-storage/src/registry.rs
```diff
@@ -75,6 +75,12 @@ impl<KV: BackgroundKeyValueStore> Registry<KV, Normal> {
             .map_err(|error| OperationalError::WorkerRuntimeCreationFailed { error })?;
         Ok(Arc::new(runtime))
     }
+
+    #[cfg(any(test, feature = "unstable-test-utils"))]
+    /// Get a [`Handle`] to the registry's runtime.
```
```suggestion
    /// Get a [`Handle`] to the registry's runtime.
    #[cfg(any(test, feature = "unstable-test-utils"))]
```
nit: docs before annots

durable-storage/src/test_helpers.rs
```diff
@@ -18,17 +18,27 @@ use octez_riscv_data::mode::Normal;
...
+use crate::database::TracedDatabase;
+use crate::errors::Error;
+use crate::errors::OperationalError;
 use crate::key::KEY_MAX_SIZE;
 use crate::key::Key;
 use crate::merkle_worker::BackgroundPersistentKeyValueStore;
 use crate::registry::Registry;
 use crate::repo::RegistryRepo;
 
+pub const VALUE_MAX_SIZE: usize = 10_000;
```
emturner, line 38: could you please add a doc string? 

(EDIT: sorry, I see this is just moved, but having one would still be useful I think)

emturner: e.g. `VALUE_MAX_SIZE >> MAX_FILE_CHUNK_SIZE` - so as is currently, more than expected of the `sets` would fail I think?

durable-storage/src/test_helpers.rs
```diff
@@ -236,164 +284,220 @@ fn update_value(value: &mut Bytes, offset: usize, bytes: Bytes) {
...
+    fn delete(&mut self, key: Key) -> Result<(), OperationalError> {
+        TracedDatabase::delete(self, key)
+    }
+    fn read(&self, key: &Key, offset: usize, output: &mut [u8]) -> Result<usize, Error> {
+        TracedDatabase::read(self, key, offset, output)
+    }
+    fn exists(&self, key: &Key) -> Result<bool, Error> {
+        TracedDatabase::exists(self, key)
+    }
+    fn value_length(&self, key: &Key) -> Result<usize, Error> {
```
emturner, line 388: nit - could you break up the trait def/impls just with newlines between the methods?

durable-storage/src/test_helpers.rs
```diff
@@ -117,6 +117,16 @@ fn make_database_operation(
     }
 }
 
+pub fn make_database_operations(
```
emturner, line 123: nit - doc?

durable-storage/src/test_helpers.rs
```diff
@@ -613,3 +651,38 @@ where
         }
     }
 }
+
+/// Run a sequence of [`DatabaseOperation`]s against a single [`TracedDatabase`],
+/// asserting against a golden in-memory model after each step, and return the recorded [`Trace`].
```
emturner: nit - could you possibly call `golden` something different? `golden` keeps making me think the model is read out a goldenfile 😅

durable-storage/src/registry.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/748eed08f0d73e6e091f7b736065921bed08520b..384f1c5f368e16c5f8341ed1a9abffac79013b2a).

durable-storage/src/test_helpers.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/748eed08f0d73e6e091f7b736065921bed08520b..384f1c5f368e16c5f8341ed1a9abffac79013b2a).

durable-storage/src/test_helpers.rs
victor-dumitrescu: Added [here](https://github.com/tezos/riscv-pvm/compare/748eed08f0d73e6e091f7b736065921bed08520b..384f1c5f368e16c5f8341ed1a9abffac79013b2a).

durable-storage/src/test_helpers.rs
victor-dumitrescu: Added doc [here](https://github.com/tezos/riscv-pvm/compare/748eed08f0d73e6e091f7b736065921bed08520b..384f1c5f368e16c5f8341ed1a9abffac79013b2a) for now.

You're right, with this value we'd expect ~80% of sets to fail. I think we should either adjust this value or the sampling distribution to get more valid sets.

durable-storage/src/test_helpers.rs
victor-dumitrescu: Fair, I renamed in a new commit [here](https://github.com/tezos/riscv-pvm/pull/1065/changes/174b55d10b21e9d603f365be5e378a64230dacba):
- `GoldenData` -> `DatabaseModel`
- `seen` field -> `data`
- `golden` references -> `model` / `registry_model`

durable-storage/src/test_helpers.rs
emturner: Would it be worth adding a follow up? 

Alternatively, I'd be happy to adjust it now too - I think the last one is roughly speaking `2.5 * PAGE_SIZE`, so would it make sense to instead be `1.5 page size`? that way, we do still cross page boundaries etc, but more of them should succeed.

But thinking about it, it might make more sense to either way discuss this properly too

kurtisc (approved):

durable-storage/src/database.rs
```diff
@@ -2224,4 +2229,19 @@ pub(crate) mod tests {
...
     });
+
+    kv_test!(test_database_end_to_end, KV: BackgroundPersistentKeyValueStore,
+    [
+        generated in crate::test_helpers::database_operations_strategy(1usize..100)
+    ],
+    {
+        // Every test iteration expects an empty repo, so not setting it in a `setup` block
+        // This is because a repo preserves database commits from previous test runs, resulting
+        // in test failures checking out a commit which isn't expected to exist succeeds.
```
thomasathorne:
```suggestion
        // in test failures when checking out a commit which isn't expected to exist succeeds.
```

victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/174b55d10b21e9d603f365be5e378a64230dacba..1310cfa548ac86f39f07d76076f73187c2a3247b).

durable-storage/src/test_helpers.rs
victor-dumitrescu: I've opened [TZX-139](https://linear.app/tezos/issue/TZX-139/review-sampling-distribution-for-values-in-end-to-end-property-based) for discussion.

emturner (approved):

PR #1068, opened 2026-05-21
test(dursto): commit + checkout steps do not affect `Database` semantics

victor-dumitrescu:
Closes RV-981.

# What

The `test_database_end_to_end` test checks that the same operations ran over `Database`s with different backends result in the same trace. This PR expands it to also check that inserting commit + checkout steps at different points in the execution of each run also results in the same trace. 

# Why

A step towards gaining confidence that a checked-out database behaves identically to the database which was committed.

# How

A `CommitCheckoutRoundtrip` variant is added to the `DatabaseOperation` test helper type. It is implemented by committing the database, then checking it out and replacing the original database with it. These steps are not produced in the proptest database operations strategy since its output is shared between the `kv_test!` runs. It is instead done using an RNG seeded differently for each backend. The seed is part of the proptest input though to allow reproducibility. 

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -2216,15 +2216,10 @@ pub(crate) mod tests {
         let mut database = new_database::<KV>(handle, &repo);
 
         // Force a per-backend divergence in the recorded trace
-        let value: &[u8] =
-            if std::any::type_name::<KV>().contains("InMemoryKeyValueStore") {
-                b"in-memory"
-            } else {
-                b"rocksdb"
-            };
+        let value = [KV::BACKEND as u8];
         let key = Key::new(b"divergence-probe").expect("Size less than KEY_MAX_SIZE");
         database
-            .set(key, Bytes::copy_from_slice(value))
+            .set(key, Bytes::copy_from_slice(&value))
             .expect("Setting should succeed");
 
         database.into_trace()
@@ -2232,7 +2227,7 @@ pub(crate) mod tests {
 
     kv_test!(test_database_end_to_end, KV: BackgroundPersistentKeyValueStore,
     [
-        generated in crate::test_helpers::database_operations_strategy(1usize..100)
+        generated in crate::test_helpers::database_operations_commit_checkout_strategy(1usize..100, 0.1)
     ],
     {
         // Every test iteration expects an empty repo, so not setting it in a `setup` block.
@@ -2240,7 +2235,14 @@ pub(crate) mod tests {
         // in test failures when checking out a commit which isn't expected to exist succeeds.
         let (_keepalive, repo) = KV::setup_repo();
 
-        let (keys, values, ops) = generated;
+        let (keys, values, ops_a, ops_b) = generated;
+
+        // Pick an operations vector so each backend exercises a different
+        // `CommitCheckoutRoundtrip` placement against the same base operations.
+        let ops = match KV::BACKEND {
+            crate::storage::Backend::Persistent => ops_a,
+            crate::storage::Backend::InMemory => ops_b,
+        };
         let operations = crate::test_helpers::make_database_operations(keys, values, ops);
         crate::test_helpers::run_database_operations::<KV>(&repo, operations)
     });
--- a/durable-storage/src/test_helpers.rs
+++ b/durable-storage/src/test_helpers.rs
@@ -49,6 +49,7 @@ pub enum DatabaseOperation {
     Hash,
     Commit,
     Checkout,
+    CommitCheckoutRoundtrip,
 }
 
 /// Operations on a [`Registry`]
@@ -73,6 +74,7 @@ pub enum DatabaseOperationView {
     Hash,
     Commit,
     Checkout,
+    CommitCheckoutRoundtrip,
 }
 
 #[derive(Debug, Clone)]
@@ -115,6 +117,9 @@ fn make_database_operation(
         DatabaseOperationView::Hash => DatabaseOperation::Hash,
         DatabaseOperationView::Commit => DatabaseOperation::Commit,
         DatabaseOperationView::Checkout => DatabaseOperation::Checkout,
+        DatabaseOperationView::CommitCheckoutRoundtrip => {
+            DatabaseOperation::CommitCheckoutRoundtrip
+        }
     }
 }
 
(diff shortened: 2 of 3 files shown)


durable-storage/src/test_helpers.rs
```diff
@@ -323,6 +324,15 @@ trait DatabaseOps<KV: BackgroundPersistentKeyValueStore>: Sized {
...
     fn checkout(handle: &Handle, repo: &KV::Repo, commit_id: CommitId) -> Result<Self, Error>;
+
+    /// Commit the database, then check out the resulting commit and replace
+    /// with the checked-out database. Tracing implementations must not record
+    /// these in the trace. Returns the resulting [`CommitId`].
+    fn commit_checkout_roundtrip(
+        &mut self,
+        handle: &Handle,
+        repo: &KV::Repo,
+    ) -> Result<CommitId, Error>;
```
emturner:
```suggestion
    ) -> Result<CommitId, OperationalError>;
```

does this need to be `Error`? it could be `OperationalError` only here I think

durable-storage/src/test_helpers.rs
```diff
@@ -400,6 +421,20 @@ impl<KV: BackgroundPersistentKeyValueStore> DatabaseOps<KV> for TracedDatabase<K
...
+        handle: &Handle,
+        repo: &KV::Repo,
+    ) -> Result<CommitId, Error> {
+        let commit_id = self.inner_mut().commit(repo)?;
+        let checked_out_db = Database::checkout(handle, repo, commit_id)?;
+
+        // Keep the trace but replace the inner database
+        *self.inner_mut() = checked_out_db;
+
+        Ok(commit_id)
```
emturner:
```suggestion
    self.inner_mut().commit_checkout_roundtrip(handle, repo)
```
would this work instead?

durable-storage/src/database.rs
```diff
@@ -2232,7 +2232,8 @@ pub(crate) mod tests {
 
     kv_test!(test_database_end_to_end, KV: BackgroundPersistentKeyValueStore,
     [
-        generated in crate::test_helpers::database_operations_strategy(1usize..100)
+        generated in crate::test_helpers::database_operations_strategy(1usize..100),
+        injection_seed in proptest::prelude::any::<u64>()
```
emturner: I'm slightly worried that this will lead to proptest really struggling with shrinking for this test.

Would it be possible instead, to have a strategy mapping that leverages `float_to_weight` and `Union::weighted` to do the mapping to/from checking out? that way it keeps it fully within the strategy generation, rather than manually seeding the hash etc

durable-storage/src/test_helpers.rs
```diff
@@ -690,14 +733,31 @@ where
 
 /// Run a sequence of [`DatabaseOperation`]s against a single [`TracedDatabase`],
 /// asserting against a reference in-memory model after each step, and return the recorded [`Trace`].
+///
+/// Before running, [`DatabaseOperation::CommitCheckoutRoundtrip`] operations
+/// are randomly but deterministically inserted at each step with probability
+/// `roundtrip_probability`. This is done here (rather than in `database_operations_strategy`)
+/// to check that test runs over the same initial `operations` are observationally equivalent
```
emturner: it would perhaps be possible to (in the `prop_map` call above, if you went that route) to return two vecs of operations in a tuple? the left one would be untouched, the right would be seeded with the commit-checkout roundtrips

alternatively, they could both be seeded but with those different placements?

emturner: but I get that it is doing something a little different than what it currently is - would be interested to hear your thoughts

durable-storage/src/database.rs
victor-dumitrescu: Yeah I thought about reproducibility but not about shrinking and you're right that I don't think shrinking on this input would work. I think the only way that could work would be to have the commit checkout steps be generated by the strategy. I've taken your suggestion to use `float_to_weight` and added a new strategy to generate 2 different operations vectors [here](https://github.com/tezos/riscv-pvm/compare/e85d10d4a0c666c9a4f6a02fa7ac42a789cffc0e..e3c57b61df86b53f4c4282494cc3191fa5a356b0).

durable-storage/src/test_helpers.rs
victor-dumitrescu: I've changed this to return 2 vectors with same base operations but different commit checkouts [here](https://github.com/tezos/riscv-pvm/compare/e85d10d4a0c666c9a4f6a02fa7ac42a789cffc0e..e3c57b61df86b53f4c4282494cc3191fa5a356b0).

durable-storage/src/test_helpers.rs
victor-dumitrescu: Good catch, done [here](https://github.com/tezos/riscv-pvm/compare/e85d10d4a0c666c9a4f6a02fa7ac42a789cffc0e..e3c57b61df86b53f4c4282494cc3191fa5a356b0)!

durable-storage/src/test_helpers.rs
victor-dumitrescu: I had it be `Error` initially because that's what `checkout` returns but you're right, done [here](https://github.com/tezos/riscv-pvm/compare/e3c57b61df86b53f4c4282494cc3191fa5a356b0..63cca20fa2a4448e50ac04e04096e50b98974d66).

durable-storage/src/test_helpers.rs
victor-dumitrescu: ~~One downside though is that the test is now roughly 3-4x slower than my initial version.~~ This actually seems to depend more on how many commit + checkouts we do rather than the strategy, I'll look more into it.

durable-storage/src/database.rs
```diff
@@ -2232,15 +2232,26 @@ pub(crate) mod tests {
...
         // This is because a repo preserves database commits from previous test runs, resulting
         // in test failures when checking out a commit which isn't expected to exist succeeds.
         let (_keepalive, repo) = KV::setup_repo();
 
-        let (keys, values, ops) = generated;
+        let (keys, values, ops_a, ops_b) = generated;
+
+        // Pick an operations vector so each backend exercises a different
+        // `CommitCheckoutRoundtrip` placement against the same base operations.
+        let backend_name = std::any::type_name::<KV>();
```
emturner: could `TestKeyValueStore` provide a helper for this? (`Backend::Persistent` vs `Backend::InMemory`)

emturner (approved):

durable-storage/src/database.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/63cca20fa2a4448e50ac04e04096e50b98974d66..8705272626dc94155868a638400db67916c159c4).

kurtisc (approved):

PR #1069, opened 2026-05-22
feat(durable): expose mode-generic foldable impl for Registry

emturner:
- Closes TZX-142
- Part of TZX-113

# What

Expose mode-generic `Foldable` for `Registry`, with `Prove` mode support for `HashFold` and `MerkleProofFold` for `Database`.

# Why

Required for `Foldable<HashFold>` and `Foldable<MerkleProofFold>` for `Prove` mode registry - needed to integrate `Prove` mode into octez.

Fix a bug in `test_verify_replays_prove_reads` that was taking the hash from the normal mode, instead of the intended prove mode, registry.

The wiring up of a `Proof` type, with `FromProof` for registry, will follow.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -16,9 +16,11 @@ use std::sync::Arc;
 
 use bytes::BufMut;
 use bytes::Bytes;
+use octez_riscv_data::foldable::Fold;
 use octez_riscv_data::foldable::Foldable;
 use octez_riscv_data::hash::Hash;
 use octez_riscv_data::hash::HashFold;
+use octez_riscv_data::merkle_proof::proof_tree::MerkleProofFold;
 use octez_riscv_data::mode::Modal;
 use octez_riscv_data::mode::Mode;
 use octez_riscv_data::mode::Normal;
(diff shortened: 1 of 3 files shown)


victor-dumitrescu (commented):
This is something that came up in an AI review, but there's an issue with the`Registry` mode-independent hash fold implementation, which is that in `Prove` mode it will record the access to the registry size, which then changes the proof:

```rust
    kv_test!(test_hashing_prove_registry_does_not_record_reads, KV: BackgroundKeyValueStore, {

        let key_a = Key::new(&[1]).expect("Size less than KEY_MAX_SIZE");
        let key_b = Key::new(&[2]).expect("Size less than KEY_MAX_SIZE");

        let mut registry = setup_size_2_registry::<KV>(repo);

        registry
            .database_mut(0)
            .expect("database at index 0 exists")
            .set(key_a, Bytes::copy_from_slice(b"foo"))
            .expect("Setting in Normal mode should succeed");

        registry
            .database_mut(1)
            .expect("database at index 1 exists")
            .set(key_b, Bytes::copy_from_slice(b"bar"))
            .expect("Setting in Normal mode should succeed");

        let prove_registry = registry
            .try_start_proof()
            .expect("Converting to prove mode should succeed");

        let proof_before_hash =
            octez_riscv_data::merkle_proof::proof_tree::MerkleProof::from_foldable(
                &prove_registry.databases,
            );
        let proof_before_hash_bytes =
            serialise(&proof_before_hash).expect("Serialising proof should succeed");

        let normal_root = Hash::from_foldable(&registry);
        let prove_root = Hash::from_foldable(&prove_registry);
        assert_eq!(prove_root, normal_root);

        let proof_after_hash =
            octez_riscv_data::merkle_proof::proof_tree::MerkleProof::from_foldable(
                &prove_registry.databases,
            );
        let proof_after_hash_bytes =
            serialise(&proof_after_hash).expect("Serialising proof should succeed");

        assert_eq!(
            proof_after_hash_bytes, proof_before_hash_bytes,
            "Proof should be unchanged by hashing"
        );
    });

```

emturner:
@victor-dumitrescu - thank you so much for the spot! That's a really good point.

I've traced the issue ultimately to the fold as you mentioned - for now, I think the easiest thing is to have the `fold` be over the vector fold itself; which indeed does the right thing (of not recording the size access in prove mode). Especially given that (quite rightly) the `unrecorded_*` methods are private to the vector module.

This does mean that the arity for the registry goes from `2 -> 4`, but I think that's okay; especially given we have the future follow up to move arity to `2` everywhere.

I've included your test, too - which now passes

victor-dumitrescu (approved):

NSant215 (approved):

PR #1075, opened 2026-05-29
test(dursto): end to end and regression tests for `Database` also test proofs

victor-dumitrescu:
Closes RV-982.

# What

Extends `test_database_end_to_end` to also produce a proof on every step which can be verified and is the same between the in-memory and the persistent backends, including in the presence of commit and checkout steps. Introduces `test_database_proof_regression` which runs one of the regression tests but also records the proof produced at every step.

# Why

To gain confidence that proofs can be produced and verified for every valid step.

# How

A `Proof` variant is added to the `TraceEntry` enum. The `TracedDatabase` machinery is then used to record produced proofs. 

# Manually Testing

```
make all
```

# Regressions

A new regression file `tests/expected/database_00.proof-trace` is checked-in, which is the trace corresponding to the existing `tests/inputs/database_00.input` with the addition of `Proof` entries. I've only added this for a single test in order to reduce bloat in the repo.

# Tasks for the Author


durable-storage/src/database.rs
```diff
@@ -444,6 +444,19 @@ impl<KV> Database<KV, Verify> {
             },
         }
     }
+
+    /// Construct a Verify-mode database by deserialising a [`MerkleProof`].
+    ///
+    /// [`MerkleProof`]: octez_riscv_data::merkle_proof::proof_tree::MerkleProof
+    #[cfg(test)]
+    pub(crate) fn from_proof(
```
emturner, line 452: could perhaps just implement `FromProof` directly? It's needed anyway

emturner: (feel free to resolve, can tackle this myself too)

durable-storage/src/test_helpers.rs
```diff
@@ -852,3 +864,132 @@ where
...
+        DatabaseOperation::ValueLength(key) => {
+            let _ = database.value_length(key);
+        }
+        DatabaseOperation::Hash => {
+            let _ = database.hash();
+        }
+        DatabaseOperation::Commit
+        | DatabaseOperation::Checkout
+        | DatabaseOperation::CommitCheckoutRoundtrip => {
+            return false;
```
emturner: should this still commit/checkout?

durable-storage/src/test_helpers.rs
```diff
@@ -852,3 +864,132 @@ where
...
 }
+
+/// Apply `operation` to a traced `database`, recording its [`TraceEntry`].
+///
+/// Returns `true` if the operation was a provable step.
+#[cfg(test)]
+fn apply_step<KV: BackgroundKeyValueStore, M: DatabaseMode>(
+    database: &mut TracedDatabase<KV, M>,
+    operation: &DatabaseOperation,
+) -> bool {
```
emturner: I think maybe `OperationalError` should still be surfaced? though it's fair to swallow `InvalidArgument` errors

durable-storage/src/test_helpers.rs
emturner: ah sorry, I see this is for `prove` instead - could the documentation be updated? could also specialise it onto `Prove` (e.g. have this in a module, and only expose `apply_prove_step` / `apply_verify_step` which are specialised on the relevant mode)

durable-storage/src/test_helpers.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/b623daa3ee47b22f4ecde08d56b5860111ea6e73..2a15028fb2d8912cee7a855f3073d45ed3b889c2).

durable-storage/src/test_helpers.rs
victor-dumitrescu: I'm using this function with both `Prove` and `Verify` Databases. I've renamed the function which calls it to `rove_and_verify_operation` [here](https://github.com/tezos/riscv-pvm/compare/2a15028fb2d8912cee7a855f3073d45ed3b889c2..0f063993fa9199c76a92b796b95ab79c88385df9) to make it clearer.

durable-storage/src/database.rs
victor-dumitrescu: This ended up being a bit more complicated than I anticipated and I think it deserves a review on its own, so I opened a follow-up PR https://github.com/tezos/riscv-pvm/pull/1077 which I'll take out of draft once this gets merged.

emturner (approved):

NSant215 (approved):

PR #1077, opened 2026-06-01
feat(dursto): implement FromProof for Database

victor-dumitrescu:
# What

Implements `FromProof` for `Database<KV, Verify>`.

# Why

Needed in order to expose proof verification

# How

This also required implementing `FromProof` for `MerkleLayer` and its inner `Tree<VerifyNodeId>`. The latter implementations are mostly based on the `MerkleLayer<KV, Verify>::from_proof` test helper. This helper, along with the equivalent `Database` one, are now removed.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/tree.rs
+++ b/durable-storage/src/avl/tree.rs
@@ -21,9 +21,12 @@ use octez_riscv_data::hash::HashFold;
 use octez_riscv_data::hash::PartialHash;
 use octez_riscv_data::hash::PartialHashFold;
 use octez_riscv_data::merkle_proof::Deserialiser;
+use octez_riscv_data::merkle_proof::DeserialiserError;
 use octez_riscv_data::merkle_proof::DeserialiserNode;
 use octez_riscv_data::merkle_proof::FromProof;
 use octez_riscv_data::merkle_proof::Partial;
+use octez_riscv_data::merkle_proof::ProofError;
+use octez_riscv_data::merkle_proof::SuspendedResult;
 use octez_riscv_data::mode::utils::not_found;
 use octez_riscv_data::serialisation::deserialise;
 use octez_riscv_data::serialisation::serialise;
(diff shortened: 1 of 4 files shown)


emturner (commented):
After investigating also, I suspect that this won't quite work, as the stream deserialisers don't allow capturing the owned proof.

I have a follow-up, to tackle this. The simplest way forward is to just deserialise twice: once to get the in-memory proof tree, the second time to allow the capturing to work.

We should investigate eventually, how to do capturing for stream deserialisers too.

emturner (approved):
Approving to unblock the octez integration plumbing (in combination with the produce proof MR)

durable-storage/src/avl/tree.rs
```diff
@@ -416,6 +419,22 @@ impl Foldable<PartialHashFold> for Tree<VerifyNodeId> {
     }
 }
 
+impl FromProof for Tree<VerifyNodeId> {
+    fn from_proof<Proof: Deserialiser>(proof: Proof) -> SuspendedResult<Proof, Self> {
+        let ctx = proof.into_node()?;
+        let (ctx, tree) = Tree::from_branches(ctx)?;
+        // The top-level fold for a Normal-mode `Tree<LazyNodeId>` always emits a node fold,
+        // so a well-formed proof's root deserialises as `Present`. `Blinded`/`Absent` here
+        // would mean the entire state has been hidden — an unusable verify layer.
```
kurtisc: em dash 😡
```suggestion
        // would mean the entire state has been hidden, so is an unusable verify layer.
```

kurtisc (approved):

durable-storage/src/avl/tree.rs
victor-dumitrescu: In my defence this piece of code was just moved, so I didn't rewrite the comment 😅 Reworded slightly [here](https://github.com/tezos/riscv-pvm/compare/42eb73cbe60c60378eb3b0456baf44f2dfa80134..67ce8f10d2b133ec229ed4be69431b6485c054f0) now.

PR #1083, opened 2026-06-05
feat(registry): implement FromProof/PartialHashFold for Registry

emturner:
- Closes TZX-135
- Closes TZX-134

# What

Implement `FromProof` for `Registry`

# Why

- Required to deserialise a proof, and to recover the `Registry<Verify>` instance.
- Required to get the final root hash from verify mode additionally.
- added e2e tests to demonstrate the current state, and capture the remaining issues with single-step that need to be addressed.

# How

mostly done by the verify mode for `Vector`.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/registry.rs
+++ b/durable-storage/src/registry.rs
@@ -506,13 +506,18 @@ pub(super) mod tests {
     use bytes::Bytes;
     use octez_riscv_data::components::vector::VectorMode;
     use octez_riscv_data::hash::Hash;
+    use octez_riscv_data::hash::PartialHash;
     use octez_riscv_data::merkle_proof::FromProof;
+    use octez_riscv_data::merkle_proof::proof::Proof;
+    use octez_riscv_data::merkle_proof::proof::deserialise_proof;
+    use octez_riscv_data::merkle_proof::proof::serialise_proof;
     use octez_riscv_data::merkle_proof::proof_tree::MerkleProof;
     use octez_riscv_data::merkle_proof::proof_tree::ProofPart;
     use octez_riscv_data::mode::Normal;
     use octez_riscv_data::mode::ProvableExt;
     use octez_riscv_data::mode::Prove;
     use octez_riscv_data::mode::Verify;
+    use octez_riscv_data::mode::utils::catch_not_found;
     use octez_riscv_data::serialisation::deserialise;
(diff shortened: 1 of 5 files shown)


victor-dumitrescu (approved):
LGTM. Do you think it would be useful to add an ignored test which attempts `proof_binary::deserialise::<Registry<KV, Verify>>(...)`? Could also leave for when TZX-161 gets resolved.

emturner:
> LGTM. Do you think it would be useful to add an ignored test which attempts `proof_binary::deserialise::<Registry<KV, Verify>>(...)`? Could also leave for when TZX-161 gets resolved.

I think so! I'm in the process of writing some e2e tests which demonstrate the currently missing gaps - which will include this indeed 🙏🏻 

I had intended them as a follow up - but actually it turns out it's preferable to do so now, as there's an additional change needed.

emturner:
@victor-dumitrescu this now includes implementation of `PartialHashFold` additionally, along with the tests.

Definitely worth a re-review, quite a lot more code has been added!

NSant215 (approved):

durable-storage/src/registry.rs
```diff
@@ -1537,6 +1541,378 @@ pub(super) mod tests {
...
+        let (stream_registry, verify_registry) = deserialise_proof_via_bytes::<KV>(&proof);
+
+        // The stream-pass registry is structurally correct: recorded reads replay fine.
+        stream_registry
+            .database(0)
+            .expect("Database should exist.")
+            .assert_database_value(&key_a, b"foo");
+
+        // ... but it cannot be hashed, because the owned proof was not captured.
+        let stream_hash_attempt = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
victor-dumitrescu: This could be narrowed down to `catch_not_found` I think

victor-dumitrescu (approved):

durable-storage/src/registry.rs
emturner: ah yes of course 🤦🏻‍♀️ . Thanks!
see [diff](https://github.com/tezos/riscv-pvm/compare/ae5fc29fce5e7355a1de56adcf34154b43d25d27..0f2ec91e9488a0095e8210e07ee8ad2f3afcf1c5)

PR #1084, opened 2026-06-08
test(dursto): add the machinery for a long-running `Database` test

victor-dumitrescu:
Relates to RV-984.

# What

Adds the components required for running long end to end tests over `Database`. Similar to the existing integration tests, the long test runs randomly sampled operations over multiple instances of `Database` and a reference model and checks for correctness and consistency. The main differences are:
- The reference model also informs the key strategy, in order to make sure that hot keys and recently deleted keys are also generated in addition to fresh keys.
- The test is applied over 3 `Database` instances: In addition to the 2 traced databases, it also tracks a "production" `Database<PersistenceLayer, Normal>`.
- The test is split into epochs, which are self-contained property-based tests. At the end of an epoch, one of the test cases is used to advance the state for the next epoch. In case of a test failure, everything needed to reproduce is saved to disk, so only the last epoch needs to be replayed, not the full test.

For now, a short version of this test is run as part of `make test`. 
- https://github.com/tezos/riscv-pvm/pull/1085 packages it as a separate binary and adds PR check which runs the test for 10 minutes
- https://github.com/tezos/riscv-pvm/pull/1090
- Next, it will also run as a nightly test.

# Why

To gain a higher degree of confidence in the correctness and consistency of the durable storage.

# How

Most of the functionality is added in `durable-storage/src/long_test`. In addtion, the in-memory repo is now serialisable with `rkyv` so that an in-memory database can be committed to disk and restored for failure replay.

# Manually Testing
```
make test
```

To see a printed log of the test:

```
cargo test database_long_test_restricted -- --no-capture
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/long_test/mod.rs
+++ b/durable-storage/src/long_test/mod.rs
@@ -1,0 +1,527 @@
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! Long-running property-based test for [`Database`].
+//!
+//! Each epoch advances and commits a shared base state, then runs a `proptest`
+//! test on that base. Every case applies a model-guided operation
+//! sequence to the reference model and three databases (in-memory traced,
+//! persistence traced, and production), cross-checking traces, root hashes, and
+//! proofs.
+//!
+//! On failure the committed base of the failing epoch and the shrunk operation
+//! sequence are written to `<out-dir>/failure/` so the failure can be replayed.
+//!
+//! [`Database`]: crate::database::Database
+
+pub mod model;
+pub mod run_case;
+pub mod strategy;
+
+use std::fs;
+use std::path::Path;
+use std::path::PathBuf;
+use std::time::Duration;
+use std::time::Instant;
+
+use anyhow::Context;
+use anyhow::Result;
+use anyhow::bail;
+use octez_riscv_data::hash::Hash;
+use proptest::strategy::Strategy;
+use proptest::strategy::ValueTree;
+use proptest::test_runner::Config as ProptestConfig;
+use proptest::test_runner::RngAlgorithm;
+use proptest::test_runner::TestError;
+use proptest::test_runner::TestRng;
+use proptest::test_runner::TestRunner;
+
+use self::model::LongTestModel;
+use self::run_case::Base;
+use self::run_case::advance_base;
+use self::run_case::initial_base;
+use self::run_case::run_case;
+use self::strategy::long_test_ops_strategy;
+use crate::commit::CommitId;
+use crate::persistence_layer::PersistenceLayer;
+use crate::repo::DirectoryManager;
+use crate::storage::PersistentKeyValueStore;
+use crate::storage::in_memory::InMemoryKeyValueStore;
+use crate::storage::in_memory::InMemoryRepo;
+use crate::test_helpers::DatabaseOperation;
+
+const IN_MEMORY_BASE: &str = "in-memory-base";
+const PERSISTENT_BASE: &str = "persistent-base";
+const META_FILE: &str = "meta.json";
+const REGRESSION_FILE: &str = "regression.json";
+const BASE_MODEL_FILE: &str = "base-model.json";
+
+/// Configuration for a long-running test invocation.
+pub struct LongTestConfig {
+    /// Maximum number of epochs to run. `None` runs until the time budget.
+    pub epochs: Option<u64>,
+    /// Maximum number of operations sampled per epoch.
+    pub ops_per_epoch: usize,
+    /// Number of test cases per epoch.
+    pub cases_per_epoch: u32,
+    /// Optional seed.
+    pub seed: Option<Hash>,
+    /// Time budget. The loop stops cleanly once exceeded.
+    pub time_budget: Option<Duration>,
+    /// If set, replay the failing epoch described by `<dir>/meta.json`.
+    pub replay: Option<PathBuf>,
+}
+
+/// Metadata persisted alongside a failure which enables replaying it.
+#[derive(serde::Serialize, serde::Deserialize)]
+struct FailureMeta {
+    /// Seed of the run.
+    seed: Hash,
+    /// Index of the failing epoch.
+    epoch: u64,
+    /// Operations sampled per epoch.
+    ops_per_epoch: usize,
+    /// Test cases per epoch.
+    cases_per_epoch: u32,
+    /// The commit identifying the failing epoch's starting state.
+    base_commit: CommitId,
+    /// Short description of the failure.
+    reason: String,
+    /// Git revision, if available from the environment.
+    git_sha: String,
+}
+
+/// Run the long-running test
+pub fn run_long_test(config: LongTestConfig) -> Result<()> {
+    // Replay reconstructs only the failing epoch; it is handled separately.
+    if let Some(replay_dir) = &config.replay {
+        return replay_failure(replay_dir);
+    }
+
+    let seed = config
+        .seed
+        .unwrap_or_else(|| rand::random::<[u8; 32]>().into());
+    let max_epochs = config.epochs;
+    let ops_per_epoch = config.ops_per_epoch;
+    let cases_per_epoch = config.cases_per_epoch;
+    let out_dir = tempfile::Builder::new()
+        .prefix("database_long_test-")
+        .tempdir()?
+        .keep();
+
+    eprintln!("test seed: {seed}");
+    eprintln!(
+        "out-dir: {} | ops/epoch: {ops_per_epoch} | cases/epoch: {cases_per_epoch}",
+        out_dir.display(),
+    );
+
+    let repo_dir = out_dir.join("repo");
+    fs::create_dir_all(&repo_dir)
+        .with_context(|| format!("creating repo dir {}", repo_dir.display()))?;
+    let persistent_repo =
+        DirectoryManager::new(&repo_dir).context("creating the directory manager")?;
+    let in_memory_repo = InMemoryRepo::default();
+
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(1)
+        .build()
+        .context("building the tokio runtime")?;
+    let handle = runtime.handle();
+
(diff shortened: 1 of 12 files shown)


durable-storage/src/storage/in_memory.rs
```diff
@@ -249,11 +318,70 @@ impl super::PersistentKeyValueStore for InMemoryKeyValueStore {
...
+            *store.blobs.read().expect("Lock should not be poisoned"),
+            *restored.blobs.read().expect("Lock should not be poisoned"),
+        );
+        assert_eq!(
+            *store.values.read().expect("Lock should not be poisoned"),
+            *restored.values.read().expect("Lock should not be poisoned"),
+        );
+    }
+}
+
```
emturner, line 384: sorry if I'm missing - are you actually using `fs_extra` anywhere? 

Can see  it's a 3+ year old crate, so we may not necessarily want to pull it in if we can avoid it

durable-storage/src/storage/in_memory.rs
```diff
@@ -223,7 +259,9 @@ impl super::PersistentKeyValueStore for InMemoryKeyValueStore {
             .values
             .read()
             .map_err(|_| OperationalError::LockPoisoned)?
-            .clone();
+            .iter()
+            .map(|(k, v)| (k.clone(), Bytes::copy_from_slice(v)))
+            .collect();
```
NSant215, line 264: why is this change needed? what was the code doing before that was different?

durable-storage/src/long_test/model.rs
```diff
@@ -0,0 +1,141 @@
...
+/// Snapshot of the model's key pools, used as an input for operation
+/// generation strategies
+#[derive(Clone, Debug, Default)]
+pub struct KeyPools {
+    /// All present keys, sorted.
+    pub existing: Vec<Key>,
+    /// Recently written or read keys.
+    pub hot: Vec<Key>,
+    /// Recently deleted keys.
+    pub deleted: Vec<Key>,
```
NSant215, line 31: what are these fields used for?

NSant215: is it just for including in the strategy for further operations?

durable-storage/src/storage/in_memory.rs
victor-dumitrescu: This type has changed:

```diff
struct InMemorySnapshot {
    blobs: HashMap<Bytes, Bytes>,
-    values: HashMap<Bytes, BytesMut>,
+    values: HashMap<Bytes, Bytes>,
```

to allow (de)serialisation (`BytesMut` can't be (de)serialised)

durable-storage/src/long_test/model.rs
victor-dumitrescu: Yes, the `KeyPools` extracted from the reference model are fed into the operations strategy for that epoch. For now, they only inform the key generation strategy specifically (see `pooled_key_strategy`).

durable-storage/src/long_test/mod.rs
```diff
@@ -0,0 +1,520 @@
...
+    eprintln!("test seed: {seed}");
+    eprintln!(
+        "out-dir: {} | ops/epoch: {ops_per_epoch} | cases/epoch: {cases_per_epoch}",
+        out_dir.display(),
+    );
+
+    let repo_dir = out_dir.join("repo");
+    fs::create_dir_all(&repo_dir)
+        .with_context(|| format!("creating repo dir {}", repo_dir.display()))?;
+    let persistant_repo =
```
NSant215:
```suggestion
    let persistent_repo =
```

durable-storage/src/long_test/run_case.rs
```diff
@@ -0,0 +1,291 @@
...
+pub struct Base {
+    /// The commit identifying the starting state (identical across backends).
+    pub commit: CommitId,
+    /// The reference model corresponding to `commit`.
+    pub model: LongTestModel,
+}
+
+/// Apply a single value operation to one database and assert its observable
+/// result matches `model` (the state *before* the operation is applied).
+fn apply_value_op<KV, D>(db: &mut D, op: &DatabaseOperation, model: &LongTestModel)
```
NSant215: I think this function has a lot of overlap with `test_helpers::apply-database_operation` - is there a way to refactor them together?

victor-dumitrescu: That's a good point, I've introduced a trait in a [new commit](https://github.com/tezos/riscv-pvm/pull/1084/changes/9fcd777bb753c52a7da0156d5b7f90f3e4a79d35) which both the existing integration test model and the new long tests model now implement. This allows reusing this function (now named `check_and_apply_value_operation`) with both of them.

durable-storage/src/long_test/mod.rs
victor-dumitrescu: Thanks, this always gets me for some reason :sweat_smile: (sorry, it's part of a larger [diff](https://github.com/tezos/riscv-pvm/compare/a47a293cf35d662467442eee79cc1ab1f6b7ad56..91b8c0b5ae7e803b3bff4c33576d10a31be1c2ce))

NSant215 (approved):

durable-storage/src/long_test/mod.rs
```diff
@@ -0,0 +1,520 @@
...
+        // Run the property test on this base.
+        let strategy = long_test_ops_strategy(&base.model.pools(), ops_per_epoch);
+        let result = runner.run(&strategy, |ops| {
+            run_case(handle, &in_memory_repo, &persistent_repo, &base, &ops);
+            Ok(())
+        });
+
+        match result {
+            Ok(()) => {
+                eprintln!("epoch {epoch} ok ({} keys)", base.model.data.len());
```
emturner:
```suggestion
                eprintln!("epoch {epoch} ok (db contains {} entries)", base.model.data.len());
```
maybe? initially I thought when running locally this was how many keys/ops were tried 😅

durable-storage/src/long_test/mod.rs
```diff
@@ -0,0 +1,520 @@
...
+        fs::File::create(failure_dir.join("meta.json")).context("creating failure meta.json")?;
+    serde_json::to_writer_pretty(meta_file, meta).context("writing failure meta.json")?;
+
+    let ops_file = fs::File::create(failure_dir.join("regression.json"))
+        .context("creating failure regression.json")?;
+    serde_json::to_writer_pretty(ops_file, ops).context("writing failure regression.json")?;
+
+    let model_file = fs::File::create(failure_dir.join("base-model.json"))
+        .context("creating failure base-model.json")?;
+    serde_json::to_writer_pretty(model_file, model).context("writing failure base-model.json")?;
```
emturner:
```suggestion    
fn write_failure_file(failure_dir: &Path, content: &impl serde::ser::Serialize, name: &str) -> Result<()> {
        let file = fs::File::create(failure_dir.join(name)).context(format!("creating failure {name}"))?;
        serde_json::to_writer_pretty(file, content).context(format!("writing failure {name}"))?;

        Ok(())
    }
    write_failure_file(&failure_dir, meta, "meta.json")?;
    write_failure_file(&failure_dir, &ops, "regression.json")?;
    write_failure_file(&failure_dir, model, "base_model.json")?;
```

nit:  i think it's the same operation three times?

victor-dumitrescu: Indeed, done [here](https://github.com/tezos/riscv-pvm/compare/e652f91afaea481594ab60a60586a9ce564a4711..e91b11e615fd9e78ef16c3a4e1d75f0e479ec3f8).

durable-storage/src/long_test/mod.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/e652f91afaea481594ab60a60586a9ce564a4711..e91b11e615fd9e78ef16c3a4e1d75f0e479ec3f8).

durable-storage/src/storage/in_memory.rs
victor-dumitrescu: I was using it for recursive directory copying in this PR and for getting a directory size in future PRs. Directory copying can actually be replaced with `commit_to_path` / `checkout_from_path` (done [here](https://github.com/tezos/riscv-pvm/compare/91b8c0b5ae7e803b3bff4c33576d10a31be1c2ce..e652f91afaea481594ab60a60586a9ce564a4711)), so I've removed the dependency.

durable-storage/src/long_test/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+    eprintln!(
+        "failure artifacts written to {}; replay with --replay {}",
+        failure_dir.display(),
+        failure_dir.display(),
+    );
+    Ok(())
+}
+
+fn read_failure_meta(dir: &Path) -> Result<FailureMeta> {
+    let file = fs::File::open(dir.join("meta.json"))
```
emturner: I think you could collapse these three with a shared helper too, in a similar manner?

durable-storage/src/long_test/model.rs
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+            DatabaseOperation::Read(key, _, _) => {
+                if self.data.contains_key(key) {
+                    self.touch_hot(key);
+                }
+            }
+            DatabaseOperation::Delete(key) => {
+                if self.data.remove(key).is_some() {
+                    self.mark_deleted(key);
```
emturner, line 111: should this remove the key from hot? (and vice versa, remove the key from deleted if it's written and so becomes hot?)

emturner (approved):

durable-storage/src/long_test/model.rs
victor-dumitrescu: Good point, done [here](https://github.com/tezos/riscv-pvm/compare/e91b11e615fd9e78ef16c3a4e1d75f0e479ec3f8..9152d8af004d598a5df7a0ad27dc9d932e6d299a).

durable-storage/src/long_test/mod.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/e91b11e615fd9e78ef16c3a4e1d75f0e479ec3f8..9152d8af004d598a5df7a0ad27dc9d932e6d299a).

PR #1089, opened 2026-06-09
fix(dursto): open databases read-only when checking out

victor-dumitrescu:
Relates to RV-984

# What

Opens RocksDB databases as read-only when checking out.

# Why

Stumbled upon this while trying to diagnose an issue with the long tests (https://github.com/tezos/riscv-pvm/pull/1084) where I noticed that the size of a commit is (surprisingly) correlated to the the number of proptest cases executed. This happens because every test case starts with a checkout, which, in read-write mode, will result in a new `LOG` file, with the previous log file stored as`LOG.old.XXXX`. This balloons the size of a commit from ~1.5 MiB for 10 test cases to ~22 MiB for 256 test cases. With this fix, the initial commit drops to 0.24 MiB.

# How

By using `open_cf_descriptors_read_only` instead of `open_cf_descriptors` for checking out.

An alternative fix would have been to limit the number of log files that are kept using [`set_keep_log_file_num`](https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#method.set_keep_log_file_num). We could fall back on it if there's a good reason to keep opening DBs in read-write mode.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/database.rs
+++ b/durable-storage/src/database.rs
@@ -1052,6 +1052,9 @@ pub(crate) mod tests {
         commit_db
             .delete_cf(blob_cf, commit_id.as_hash().as_ref())
             .expect("Deleting the root blob should succeed");
+        commit_db
+            .flush_cf(blob_cf)
+            .expect("Flushing the blob column family should succeed");
         drop(commit_db);
 
         assert!(matches!(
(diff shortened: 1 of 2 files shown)


emturner (approved):
I came across this yesterday too! But this time in the context of the file-handles investigation. 

The diff is identical to the one I hadn't pushed yet 😅 so LGTM

kurtisc (approved):

PR #1091, opened 2026-06-11
fix(dursto): Support database move and copy

NSant215:
Closes TZX-173. 

# What

Add support for move, copy and clearing a database for single step. 

# Why

These are operations we are required to support. 

# How

When overwriting a tree of a destination database, add required nodes to the deleted nodes set of ProveResolver. 

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/resolver.rs
+++ b/durable-storage/src/avl/resolver.rs
@@ -855,6 +855,46 @@ impl<R> ProveResolver<R> {
         self.deleted_nodes.deref().borrow()
     }
 
+    /// Stash the prove-mode projections of every initial-origin node in `tree` into the
+    /// deleted-nodes map.
+    ///
+    /// Replacing an entire working tree (database copy/move/clear) unlinks every node in it at
+    /// once. The `MerkleLayer`-level fold looks up the read flags of an accessed initial node
+    /// either in the deleted-nodes map or in the current working tree, so the outgoing tree's
+    /// resolved nodes must be stashed just like individually deleted ones.
+    ///
+    /// Unresolved nodes were never accessed — the fold blinds them without needing their flags —
+    /// and nothing beneath an unresolved node can have been materialised, so such subtrees are
+    /// skipped wholesale. Nodes created during the step carry no lazy identifier and are not part
+    /// of the initial tree, but their subtrees may be, so only the recursion continues for them.
+    pub(crate) fn track_unlinked_tree(&self, tree: &Tree<ProveNodeId>) {
+        let mut stack = vec![tree];
+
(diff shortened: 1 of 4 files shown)


durable-storage/src/avl/resolver.rs
```diff
@@ -855,6 +855,47 @@ impl<R> ProveResolver<R> {
...
+                self.deleted_nodes
+                    .borrow_mut()
+                    .insert(Hash::from_foldable(lazy_id), fields);
+            }
+
+            for child in [node.left_id(), node.right_id()] {
+                if let Some(subtree) = child.inner() {
+                    stack.push(subtree);
+                }
+            }
```
```suggestion
            [node.left_id(), node.right_id()]
                .filter_map(|child| child.inner())
                .for_each(|subtree| stack.push(subtree));
```
you could maybe do something like this, but no strong preference 😄

durable-storage/src/merkle_layer.rs
```diff
@@ -143,6 +143,24 @@ impl<KV> MerkleLayer<KV, Prove<'static>> {
...
+        self.inner.working_tree.clone()
+    }
+
+    /// Replace the working tree with `new`, returning the outgoing tree.
+    ///
+    /// The layer's initial tree and access tracking are deliberately kept: the proof must still
+    /// encode this layer's state at the start of the step, with everything read before the
+    /// replacement included. The outgoing tree's resolved nodes are stashed as unlinked so the
+    /// proof fold can still find their read flags.
+    pub(crate) fn replace_working_tree(&mut self, new: Tree<ProveNodeId>) -> Tree<ProveNodeId> {
```
emturner, line 161: could you link the TODO here too, relevant that this fails in multi-step (as the working tree will now resolve against the wrong initial tree)

emturner (approved):
discussed offline - happy to merge pretty much as is, but the approach might be shifted slightly with a view to tackling multi-step proofs

victor-dumitrescu (approved):

durable-storage/src/merkle_layer.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/7ec0442cc74073a6d500cded1161292c2e4705cf..3c074d906bf132c5c884764ab635a3704468ec57)

durable-storage/src/avl/resolver.rs
NSant215: done [here](https://github.com/tezos/riscv-pvm/compare/6dfbc6c7bea4ef48a320f9d7bc2ab3882bc741dc..7ec0442cc74073a6d500cded1161292c2e4705cf)

PR #1092, opened 2026-06-15
test(dursto): run Database long test nightly

victor-dumitrescu:
Closes RV-1009. Closes RV-984.

# What

Runs the `Database` long test as a nightly test for ~5h.

# Why

To increase confidence in `Database` semantics and proof validity. Nightly tests run on GitHub runners are capped at 6h, including job setup time.

# How

The long test binary can now take an optional `--out-dir` argument for the working directory instead of using a default temporary directory. The nightly test definition uses it so it can recover the failure artifact if needed.

Setup for the PR CI is pulled into a new `actions/setup/action.yml` configuration, which is now shared between PR and nightly jobs. 

# Manually Testing

Passing the output directory for the test binary can be tested with a short run:

```
cargo run --release --features rocksdb,unstable-test-utils --bin database_long_test -- test --epochs 10 --ops-per-epoch 100 --cases-per-epoch 20 --out-dir XXXX
```

I don't think there's any way to test the nightly CI configuration locally.

# Tasks for the Author

### DIFF
--- a/durable-storage/Makefile
+++ b/durable-storage/Makefile
@@ -20,8 +20,14 @@ reset-regressions:
 	@cargo run -p xtask -- gen-database-regression-inputs
 	@UPDATE_GOLDENFILES=1 cargo nextest run test_database_regression
 
-long-test:
+database-long-test:
 	@cargo run --release --features rocksdb,unstable-test-utils \
 		--bin database_long_test -- test --max-minutes 10 --keep-epochs 1
 
-.PHONY: all check test gen-regression-inputs long-test
+database-long-test-nightly:
+	@cargo run --release --features rocksdb,unstable-test-utils \
+		--bin database_long_test -- test \
+		--ops-per-epoch 2000 --cases-per-epoch 256 --keep-epochs 1 \
+		--max-minutes 300 $(if $(OUT_DIR),--out-dir "$(OUT_DIR)")
+
+.PHONY: all check test gen-regression-inputs database-long-test database-long-test-nightly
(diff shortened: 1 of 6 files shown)


durable-storage/Makefile
```diff
@@ -20,8 +20,14 @@ reset-regressions:
...
+database-long-test:
 	@cargo run --release --features rocksdb,unstable-test-utils \
 		--bin database_long_test -- test --max-minutes 10 --keep-epochs 1
 
-.PHONY: all check test gen-regression-inputs long-test
+database-long-test-nightly:
+	@cargo run --release --features rocksdb,unstable-test-utils \
+		--bin database_long_test -- test \
+		--ops-per-epoch 2000 --cases-per-epoch 256 --keep-epochs 1 \
+		--max-minutes 300 $(if $(OUT_DIR),--out-dir "$(OUT_DIR)")
```
jobjo, line 31: Non blocking. Consider making it configurable (with default to 300) so it can be passed from `nightly.yml`  and aligned with the `timeout-minutes`.

jobjo (approved):
LGTM

durable-storage/Makefile
victor-dumitrescu: Indeed, I think it would be good to make it (along with the other test options) configurable from the nightly job once this gets merged and I can confirm that the single run works well. This is because GitHub allows us to run up to 20 jobs in parallel. These multiple tests could be configured differently, for example some could rapidly add keys to explore a growing key space, while others might grow it slower but do more test cases per epoch.

Just to note, total job time (defined in the nightly job) and the test time out (defined in the Make target currently) will still need to be defined separately, because the former needs to be longer to account for setup time as well. If that's ok, I propose making tests fully configurable from the nightly job when we introduce more tests as a follow-up.

emturner (approved):

PR #1093, opened 2026-06-15
fix(durable): handle absent/blinded databases

emturner:
- Part of TZX-176

# What

Fixes two issues around blind/absent databases:
- registry proof deserialisation fails when the prove registry is not touched
- databases should only be present when actually touched by an operation

# Why

Required for e2e proof->verification flow to work. (Especially when such a proof is 'irmin-only': empty operation set on NDS *must* be supported)

Additionally, allowing databases to be blinded/omitted is required to avoid proof having a linear size correlation with the registry size.

# How

The verify mode merkle_layer now uses `VerifyTreeId` directly, rather than `Tree<VerifyNodeId>`. This allows the root tree itself to be absent/blinded.

# Manually Testing

```
make all
```

# Regressions

the database proof regressions are updated. There are now some which allow for the database to be fully blinded - whereas it was not the case before. (ie hash queries allow the db to be blinded, rather than contain all content.) This is mostly related to the 'empty tree'.

While this looks like a loss w.r.t. proof size, it's actually a win: this change allows the dbs to be blinded at the _registry_ level - which the regressions don't currently capture.

# Tasks for the Author


durable-storage/src/registry.rs
```diff
@@ -1765,6 +1766,128 @@ pub(super) mod tests {
...
+        for n in 1..=16 {
+            registry.resize_tick(n).expect("resize should succeed");
+        }
+        for i in 0..16 {
+            populate_database_with_key_value::<KV>(&mut registry, i, &[i as u8], b"val");
+        }
+
+        let root_hash = Hash::from_foldable(&registry);
+
+        let key5 = Key::new(&[5]).unwrap();
```
victor-dumitrescu: Might be good to have this test as a loop over all databases instead of just one?

durable-storage/src/merkle_layer.rs
```diff
@@ -642,7 +636,12 @@ impl<KV: KeyValueStore> Foldable<MerkleProofFold> for InitialTreeFold<'_, KV> {
     fn fold(&self, builder: MerkleProofFold) -> <MerkleProofFold as Fold>::Folded {
         let tree_hash = Hash::from_foldable(self.tree_id);
 
-        if !self.prove_impl.resolver.was_tree_accessed(&tree_hash) {
+        // If the database was not touched at all, then it is safely
+        // blindable.
+        //
+        // *NB* inserting into an empty tree works, as the deserialiser
+        // correctly recognises the special `EMPTY_TREE_HASH`.
```
victor-dumitrescu: I don't think there's a test for this specifically, might be good to have one if that's the case.

victor-dumitrescu (approved):

durable-storage/src/merkle_layer.rs
emturner: It's not covered at the root - but it is lower down the tree (which is then just the empty tree again?)

But I'll add an explicit root-based one

emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/7db020fdf6201774b050c392b521761fc740347d..e884e30693764580cdc79e89210b0128cdbfc1e7)

durable-storage/src/registry.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/7db020fdf6201774b050c392b521761fc740347d..e884e30693764580cdc79e89210b0128cdbfc1e7)

victor-dumitrescu (approved):

durable-storage/src/merkle_layer.rs
```diff
@@ -642,7 +636,12 @@ impl<KV: KeyValueStore> Foldable<MerkleProofFold> for InitialTreeFold<'_, KV> {
     fn fold(&self, builder: MerkleProofFold) -> <MerkleProofFold as Fold>::Folded {
         let tree_hash = Hash::from_foldable(self.tree_id);
 
-        if !self.prove_impl.resolver.was_tree_accessed(&tree_hash) {
```
NSant215, line 645: I'm a bit confused by this change? Before, this check would act on every tree and blind if not accessed. Now we always unblind trees as long as the root was accessed I guess? Not quite clear on the need for it.

emturner: great catch - thank you! have pushed in [diff](https://github.com/tezos/riscv-pvm/compare/e884e30693764580cdc79e89210b0128cdbfc1e7..73d854216fa7b6f8c30fd257983ab54d1843f2f2)

Essentially, we now do the whole database check 'once' at the root level in the `MerkleLayer` fold itself. Then on each subtree (other than the root tree) we do the check on the root hash of that subtree.

(This distinction is required, as the root tree is automatically resolved, and never tracked, by the prove resolver - which does simplify a few things around prove mode there IIRC (I have tried to not do this, but it does lead to some complications)).

The benefit is indeed, the prove regressions that are changed are explicitly only on the root/blindness issue (easier to see on the PR diff itself, rather than the change here). We now correctly blind subtrees further down, rather than only blind the contained node.

Thanks!

NSant215 (approved):

PR #1100, opened 2026-06-19
test(dursto): prove and verify every provable Registry step

victor-dumitrescu:
Closes TZX-179.

# What

Adds proof production and verification on every step of every `Registry` end to end test.

# Why

To gain confidence in `Registry` single step proof production and verification.

# How

The diff is quite long, but most of it is splitting `test_helpers/registry_test_helpers.rs` into 2 modules, one for `Database` helpers and one for `Registry` helpers. The names of equivalent declarations in either one has also been harmonised.

The last commit introduces the new test functionality: `run_operations` becomes `run_and_prove_registry_operations`. Its implementation and that of the extra helpers required are similar to their existing `Database` equivalents.

# Manually Testing

```
make all
```

# Tasks for the Author


durable-storage/src/test_helpers/registry.rs
```diff
@@ -0,0 +1,266 @@
+// SPDX-FileCopyrightText: 2026 Trilitech <contact@trili.tech>
```
emturner, line 1: I guess does the line it would be nice to avoid the global imports of the submodules? 

But at least for now it's completely okay 👍🏻

emturner: (sorry, this comment should be in `test_helpers.rs`

durable-storage/src/test_helpers/registry_test_helpers.rs
```diff
@@ -134,12 +155,139 @@ where
...
+    M: RegistryMode + DatabaseMode,
+{
+    match op {
+        Operation::Database(
+            DatabaseOperation::Commit
+            | DatabaseOperation::Checkout
+            | DatabaseOperation::CommitCheckoutRoundtrip,
+        ) => return false,
+        Operation::Database(db_op) => {
+            let index = get_registry_index(registry);
```
emturner: it would be nice to improve this perhaps, to avoid get_registry_index always returning the same thing?

Would mean that for copy/move below we could then try more situations / mixes

e.g. copy to self or error cases etc

Perhaps `Operation` could be generated with an index as needed?

durable-storage/src/test_helpers/registry_test_helpers.rs
```diff
@@ -134,12 +155,139 @@ where
...
+            let database = registry
+                .database_mut(index)
+                .expect("The index is in bounds");
+            return apply_database_step(database, db_op).expect("applying a step should succeed");
+        }
+        Operation::GrowRegistry => grow_registry(registry),
+        Operation::ShrinkRegistry => {
+            if registry.is_empty() {
+                grow_registry(registry);
```
emturner:
```suggestion
                return;
```
shinking an empty registry probably shouldn't grow it?

victor-dumitrescu: I think the original intention with the `Registry` end to end tests was to not have these steps be noops in case of an empty registry (it's the same behaviour on `main`, and for the other operations too either directly or indirectly through `get_registry_index_with_model` / `get_index` previously).

Although I agree this is not ideal, especially here when we're proving steps which turn out to be made of 2 operations. We could perhaps simplify all of these by making sure we have a much higher ratio of grows to shrinks in the operations strategy?

durable-storage/src/test_helpers/registry.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/bbb48792536891868ea50945b7af2d5aaa2a2c07..615b5dcf0e61f7f3c771f6c8ccba999e8a093c70).

emturner: sorry, could you possibly change to

`test_helpers::database::` and `test_helpers::registry` rather than have the second `test_helpers` after database/registry?

durable-storage/src/test_helpers/registry_test_helpers.rs
emturner: having more grows than shrinks would also work - though I worry with long-running tests we'd start to run into ulimit issues if it stays as it currently is?

emturner: it's nice to test pure no-ops too imo, as that would likely be the more frequent thing to happen (no-op proofs are part of the dual-proof system)

durable-storage/src/test_helpers/registry.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/dbe07d5baede3726944428e4108a59236c36e626..a417ec12c3d1766ca7b3dfffd1c4d7c66c041227).

durable-storage/src/test_helpers/registry_test_helpers.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/pull/1100/changes/a417ec12c3d1766ca7b3dfffd1c4d7c66c041227): the registry is now never shrunk to empty and any such shrink becomes a no-op. This means we also don't need to grow it on empty to apply the other operations any more.

durable-storage/src/test_helpers/registry_test_helpers.rs
victor-dumitrescu: Yes this is definitely an improvement we should do, I propose I do it in a follow-up as this PR is already quite big and this change will be somewhat substantial too.

emturner: sounds good!

emturner (approved):

jobjo (approved):

PR #1109, opened 2026-06-24
test(dursto): improve registry index selection in end to end tests

victor-dumitrescu:
Closes TZX-185.

# What

Currently in Registry end to end tests, operations which take indices of databases in the registry are always set to be the last index or indices (`len-1` and, if needed, `len-2`). Instead, indices are now sampled randomly as part of the `OperationView` -> `Operation` transformation.

# Why

In order to reduce this limitation to the breadth of the test space.

# How

A proptest `Index` is used for indices in the expanded `OperationView` variants. When generating operations, the values are sampled, bounded by the expected size of the registry at the point of applying that operation. Doing it this way instead of simply sampling a `usize` modulo the registry length allows shrinking in case of a test failure.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/test_helpers/registry.rs
+++ b/durable-storage/src/test_helpers/registry.rs
@@ -11,6 +11,7 @@
 //! [`super::database`].
 
 use std::collections::HashMap;
+use std::num::NonZeroUsize;
 
 use bytes::Bytes;
 use octez_riscv_data::hash::Hash;
@@ -23,6 +24,7 @@ use octez_riscv_data::mode::Normal;
 use octez_riscv_data::mode::ProvableExt;
 use octez_riscv_data::mode::Verify;
 use proptest::prelude::*;
+use proptest::sample::Index;
 
 use super::database::DatabaseModel;
 use super::database::DatabaseOperation;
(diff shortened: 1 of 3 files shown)


durable-storage/src/test_helpers/registry.rs
```diff
@@ -96,12 +118,15 @@ pub fn registry_operations_strategy(
...
+                        .prop_map(|(i, v)| OperationView::Database(i, v)),
                     3 => Just(OperationView::GrowRegistry),
                     2 => Just(OperationView::ShrinkRegistry),
-                    2 => Just(OperationView::CopyDatabase),
-                    1 => Just(OperationView::MoveDatabase),
-                    2 => Just(OperationView::ClearDatabase),
+                    2 => (any::<Index>(), any::<Index>())
+                        .prop_map(|(src, dst)| OperationView::CopyDatabase(src, dst)),
+                    1 => (any::<Index>(), any::<Index>())
+                        .prop_map(|(src, dst)| OperationView::MoveDatabase(src, dst)),
```
jobjo, line 128: I suppose we handle the case where `src = dst` for copy/move and don't need to guard against it at generation?

durable-storage/src/test_helpers/registry.rs
```diff
@@ -96,12 +118,15 @@ pub fn registry_operations_strategy(
...
+                    2 => any::<Index>().prop_map(OperationView::ClearDatabase),
```
jobjo: Nit. Not part of the scope for this MR but having frequencies add up to 100 would be a bit nicer. Feel free to close and ignore.

durable-storage/src/test_helpers/registry.rs
jobjo: I see these are no-ops in the implementation of registry but actual copies in the model:

```
            Operation::MoveDatabase(src, dst) => {
                registry
                    .move_database(src, dst)
                    .expect("Moving the database should be successful");

                registry_model[dst] = std::mem::take(&mut registry_model[src]);
            }
        }

```

Should the latter be a no-op too  for `src = dst`?

durable-storage/src/test_helpers/registry.rs
```diff
@@ -373,31 +367,19 @@ where
...
-
+            Operation::MoveDatabase(src, dst) => {
                 registry
                     .move_database(src, dst)
                     .expect("Moving the database should be successful");
 
-                let empty = Default::default();
-                let new_dst = std::mem::replace(&mut registry_model[src], empty);
-                registry_model[dst] = new_dst;
+                registry_model[dst] = std::mem::take(&mut registry_model[src]);
```
jobjo: See comment above: https://github.com/tezos/riscv-pvm/pull/1109/changes#r3474045158

victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/52b4e59915221963326b8045ff4e061d6ccb3e8a..aad4fb572344cebf08b5e691f64ab80122b7659a).

durable-storage/src/test_helpers/registry.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/52b4e59915221963326b8045ff4e061d6ccb3e8a..aad4fb572344cebf08b5e691f64ab80122b7659a).

durable-storage/src/test_helpers/registry.rs
victor-dumitrescu: That's a good point, the model doesn't need to be updated in case of a no-op. Done [here](https://github.com/tezos/riscv-pvm/compare/52b4e59915221963326b8045ff4e061d6ccb3e8a..aad4fb572344cebf08b5e691f64ab80122b7659a).

jobjo (approved):
LGTM

emturner (approved):

PR #1110, opened 2026-06-24
test(dursto): exercise commit-checkout roundtrips in Registry end to end test

victor-dumitrescu:
Closes TZX-177.

# What

Introduces commit-checkout roundtrips in the Registry end to end test at different points in the 2 separate runs and checks they are equivalent.

# Why

To gain more confidence in the commit and checkout semantics of the durable storage and bring up to par with the Database test.

# How

Since most of the functionality has already been implemented for Database, much of this PR is a refactoring which allows it to be reused. The main change is the introduction of an `OperationView` trait which is implemented for both `DatabaseOperationView` and `RegistryOperationView` and defines strategies for sampling operations.

The actual test functionality implemented on top simply adds a `CommitCheckoutRoundtrip` variant to `RegistryOperationView` and expands the `test_durable_storage_end_to_end` test.

Registry does not have a traced equivalent like Database does so to check equivalence we rely on the `kv_test!` hash check between the 2 runs.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/test_helpers/registry.rs
+++ b/durable-storage/src/test_helpers/registry.rs
@@ -31,10 +31,7 @@ use super::database::DatabaseOperation;
 use super::database::DatabaseOperationView;
 use super::database::apply_database_operation_with_model;
 use super::database::apply_database_step;
-use super::database::database_operation_view_strategy;
-use super::database::key_strategy;
 use super::database::make_database_operation;
-use super::database::value_strategy;
 use crate::commit::CommitId;
 use crate::database::DatabaseMode;
 use crate::key::Key;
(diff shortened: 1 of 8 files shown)


emturner (approved):

durable-storage/src/test_helpers/registry.rs
```diff
@@ -275,7 +278,10 @@ where
     )
 }
```
thomasathorne, line 280: Would be nice to have a doc-string, which would also help explain why the return type has changed. 😅

thomasathorne (approved):
Generally looks good, one clarification requested. 😁

durable-storage/src/test_helpers/registry.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/720d570d2fafc0178e18c830c4b2cf68768f3e8e..f737f578a0525b9c877a2bbc7e45c8cc8bf786aa).

PR #1114, opened 2026-06-29
fix(ci): don't abort if disk cleanup step fails on macOS runners

victor-dumitrescu:
# What

Splits the `Aggressive cleanup` CI step between macOS and Linux depending on which paths can be removed on each machine type. Makes an error in the macOS step not abort the job.

# Why

The `MacOS Checks` PR job sometimes fails the disk cleanup step because GitHub might run it on a read-only machine. See [this job](https://github.com/tezos/riscv-pvm/actions/runs/28368360545/job/84039742786?pr=1113) for example. This results in intermittent, unwanted CI failures.

# Manually Testing

No way to test manually, inspect the log of the `MacOS Checks` job on this PR. A pass would only indicate the change is effective if it comes after an unsuccessful disk cleanup (like [here](https://github.com/tezos/riscv-pvm/actions/runs/28370057707/job/84045396632?pr=1114)).

# Tasks for the Author

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -29,7 +29,8 @@ outputs:
 runs:
   using: composite
   steps:
-    - name: Aggressive cleanup
+    - name: Aggressive cleanup (Linux)
+      if: runner.os == 'Linux'
       shell: bash
       run: |
         # Remove Java (JDKs)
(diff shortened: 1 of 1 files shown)


jobjo (approved):
LGTM

thomasathorne (approved):
LGTM

PR #1121, opened 2026-07-03
test(dursto): add long-running Registry test

victor-dumitrescu:
Closes TZX-180.

# What

The Registry long test is structured similarly to the Database long test: it is based on running a number of independent "test epochs" (property-based test cases) on top of a base, which is advanced after each epoch. When a test case fails, the failure artifact can be persisted so that only the failing epoch needs to be replayed to reproduce it. 

Specifically in the case of the Registry long tests:
- The test is set up with a number of "permanent" databases which cannot be moved, cleared, or copied into. This is to guarantee that a certain number of databases will steadily grow during the course of the test.
- In addition, a `--keep-stable-size` mode adjusts the probabilities of sampling registry grow and shrink operations so that the size of the registry tends towards `2 * permanent` (as opposed to equal probabilities which yield an ever growing registry). This resembles more closely the way in which Etherlink/Tezos X is expected to use the NDS.
- The `--keep-epochs` option requires a slightly more complicated snapshot pruning approach compared to the Database tests.

For now, a 10-minute run of the Registry long test is added to PR CI. Next:
- https://github.com/tezos/riscv-pvm/pull/1122

# Why

To gain a higher degree of confidence in the correctness and consistency of the durable storage.

# How

Most of the large diff of this PR comes from the first 2 commits (notably from moves in the first commit):
- refactor(test): extract shared long-test harness: 
   - change directory structure, move shared helpers to `long_test/harness.rs` and database-specific ones to `long_test/database/*`
   - test `Base` becomes `Base<LongTestModel>` to enable sharing
   - slight refactor of the Database key generation strategy (`pooled_key_strategy`) for clarity
- test(dursto): add long-running Registry test:
   - adds the Registry long test in `long_test/registry/*`
   - adds `database` and `registry` subcommands to the long test binary

The other 4 commits should hopefully be more self-explanatory.

# Manually Testing

```
make all
```

# Tasks for the Author


durable-storage/src/bin/database_long_test.rs
```diff
@@ -31,36 +30,73 @@ struct Cli {
...
+    #[arg(long)]
+    keep_epochs: Option<NonZeroUsize>,
+
+    /// Directory for run state and failure artifacts (default: a fresh tempdir).
+    #[arg(long)]
+    out_dir: Option<PathBuf>,
+}
+
+#[derive(Debug, Subcommand)]
+enum SubjectCommand {
```
emturner: should this get renamed?

```suggestion
enum DatabaseCommand {
```

durable-storage/src/long_test/registry/mod.rs
```diff
@@ -0,0 +1,522 @@
...
+            if epoch >= max {
+                break;
+            }
+        }
+        if let Some(budget) = config.time_budget {
+            if start.elapsed() >= budget {
+                eprintln!("time budget reached after {epoch} epochs");
+                break;
+            }
+        }
```
emturner: would it be possible to abstract out some of this common logic between the two tests?

durable-storage/Makefile
```diff
@@ -40,4 +40,9 @@ run-registry-long-test:
 	@cargo run --release --features rocksdb,unstable-test-utils \
 		--bin database_long_test -- registry test --max-minutes 10 --keep-epochs 1
 
-.PHONY: all check test gen-regression-inputs run-database-long-test build-database-long-test reset-database-regressions reset-registry-regressions run-registry-long-test
+run-registry-long-test-stable:
+	@cargo run --release --features rocksdb,unstable-test-utils \
+		--bin database_long_test -- registry test --max-minutes 10  --keep-epochs 1 \
```
emturner:
```suggestion
		--bin database_long_test -- registry test --max-minutes 10 --keep-epochs 1 \
```

.github/workflows/nightly.yml
```diff
@@ -80,12 +80,12 @@ jobs:
       - name: Download binary
         uses: actions/download-artifact@v8
         with:
-          name: database_long_test
+          name: long_test
```
emturner, line 83: I'd argue `durable_long_test` might be a better name? In future we could (in theory) have `pvm` long tests too

durable-storage/src/bin/database_long_test.rs
thomasathorne: Sure, done [here](https://github.com/tezos/riscv-pvm/compare/e5450561c58fec0b1f55e13f51a6ecbf78d92f62..62c35e1c9e925970e656ada7e05a04d6a3332ee6)

durable-storage/Makefile
thomasathorne: Done [here](https://github.com/tezos/riscv-pvm/compare/62c35e1c9e925970e656ada7e05a04d6a3332ee6..01a61fd519b992310413ec6d86b6e866d112aae9)

.github/workflows/nightly.yml
thomasathorne: I've applied this suggestion to the other PR (https://github.com/tezos/riscv-pvm/pull/1122).

durable-storage/src/long_test/registry/mod.rs
thomasathorne: There's a lot of duplicated logic, I tried refactoring more of it but there's lots of small differences that make things tricky. In the end the most pragmatic thing seemed to be the commit I added [here](https://github.com/tezos/riscv-pvm/pull/1121/changes/0fbce0d0f0ef3e07b82e4ffc68456f6bd076510a) which just handles the epoch timing logic in a common method.

emturner (approved):

jobjo (approved):

PR #1129, opened 2026-07-09
refactor(data): thread `codec::Bincode` through all foldable traits/FromProof

emturner:
- Part of RV-1007

# What

Prepares the way for NDS to switch to a new encoding/decoding system in future (rather than the deprecated bincode)

# Why

To avoid having to change the PVM & the NDS in one go, we introduce the notion of `codec`, to allow NDS to select a different _leaf encoding_ system from the PVM, as the first step.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/data/src/components/atom.rs
+++ b/data/src/components/atom.rs
@@ -21,6 +21,9 @@ use bincode::error::EncodeError;
 use perfect_derive::perfect_derive;
 
 use crate::clone::CloneState;
+use crate::codec::LeafCodec;
+use crate::codec::LeafDecode;
+use crate::codec::LeafEncode;
 use crate::foldable::Fold;
 use crate::foldable::FoldLeaf;
 use crate::foldable::Foldable;
@@ -45,7 +48,6 @@ use crate::mode::Prove;
 use crate::mode::Verify;
 use crate::mode::utils::Source;
 use crate::mode::utils::not_found;
-use crate::serialisation::serialise;
 
 /// Single value state component
 ///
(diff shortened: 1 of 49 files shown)


data/src/components/atom.rs
```diff
@@ -173,15 +175,21 @@ impl<T: Default + 'static, M: AtomMode> Default for Atom<T, M> {
...
 }
 
-impl<T: Encode, F: FoldLeaf> Foldable<F> for Atom<T, Normal> {
+impl<T, F: FoldLeaf> Foldable<F> for Atom<T, Normal>
+where
+    Atom<T, Normal>: LeafEncode<F::Codec>,
+{
     fn fold(&self, builder: F) -> F::Folded {
         builder
             .fold_leaf(self)
```
victor-dumitrescu: I don't remember is there was a reason for not passing `&self.atom` directly here?

data/src/merkle_proof/proof_binary.rs
```diff
@@ -56,7 +62,10 @@ impl<'t> StreamInput<'t> {
 }
 
 /// Deserialiser for [`Deserialiser`] based on a stream of bytes.
-pub struct StreamDeserialiser<'t> {
+///
+/// Parameterised by the leaf [`LeafCodec`]; defaults to [`Bincode`]. Only leaf decoding depends on
+/// the codec — the tag/blind-hash structure is always bincode-encoded.
+pub struct StreamDeserialiser<'t, C = Bincode> {
```
victor-dumitrescu, line 68:
```suggestion
+pub struct StreamDeserialiser<'t, C: LeafCodec = Bincode> {
```

(could remove it from the impl block in turn)

victor-dumitrescu (approved):

data/src/components/atom.rs
emturner: don't think so, updated in [diff](https://github.com/tezos/riscv-pvm/compare/ec9c7b4b9eeea3f1f3e65b69891bc9002306d9f5..3321b7ab62dd382fecd76d91fea5ef07c52da6dd)

data/src/merkle_proof/proof_binary.rs
emturner: I've renamed this parameter to `LeafCodec` on the struct to make it more explicit.

I'd rather not explicitly over-constrain the struct with `C: LeafCodec` right now - as that trait bound isn't used by the struct at all - only the `impl` requires it. It's just that `Bincode` is a default but in theory could be swapped out for anything.

victor-dumitrescu: Makes sense, I assumed that while not super relevant at the moment with just `Bincode`, this argument will eventually need to be constrained by a trait such as `LeafCodec`.

Although I find the change from the latest diff

```diff
- pub struct StreamDeserialiser<'t, C = Bincode> {
+ pub struct StreamDeserialiser<'t, LeafCodec = Bincode> {
```

a bit confusing, since now it's using the `LeafCodec` name but is not actually constrained by the `LeafCodec` trait.

emturner: [reverted](https://github.com/tezos/riscv-pvm/compare/3321b7ab62dd382fecd76d91fea5ef07c52da6dd..2f509fc0e0b03210944f0b979336b365699752f5)

thomasathorne (approved):

PR #1131, opened 2026-07-09
feat(durable): restrict Key to maximum u8::MAX bytes

emturner:
Part of RV-1000 (will require an update to the `sdk` crates too).

# What

Restrict the maximum size of a key to 255 bytes, rather than 256.

# Why

This allows us to save 3 bytes on every encoded key. 

# How

Via a manual encode/decode impl.

# Manually Testing

```
make all
```

# Regressions

Registry/Database regression inputs needed resetting, as some operations contained keys that are now too long.

# Tasks for the Author


durable-storage/src/key.rs
```diff
@@ -42,18 +45,30 @@ impl AsRef<[u8]> for Key {
     }
 }
 
-// Manual implementation of Decode to ensure that we do not decode a key that is formed of
-// invalid bytes.
+// Manual implementation of Decode to allow the smaller prefix size to be used.
```
thomasathorne:
```suggestion
// Manual implementation of Encode to allow the smaller prefix size to be used.
```

thomasathorne (approved):

durable-storage/src/key.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/fabe41279537bf4bc3a356e8b18c216b142235c7..9c16746babddb28079007d07e7b9c74876bc1c9a)

durable-storage/src/key.rs
```diff
@@ -42,18 +45,30 @@ impl AsRef<[u8]> for Key {
     }
 }
 
-// Manual implementation of Decode to ensure that we do not decode a key that is formed of
-// invalid bytes.
+// Manual implementation of Encode to allow the smaller prefix size to be used.
+impl Encode for Key {
+    fn encode<E: Encoder>(&self, encoder: &mut E) -> Result<(), bincode::error::EncodeError> {
+        let bytes = self.0.as_slice();
+        let len = u8::try_from(bytes.len()).expect("KEY_MAX_SIZE is precisely u8::MAX");
```
jobjo, line 52: I suppose we really want this to panic rather than returning an Error for longer keys.

jobjo (approved):

durable-storage/src/key.rs
emturner: This module fully prevents any longer key from existing - so this panic can never happen

PR #1139, opened 2026-07-16
chore(durable): record decision w.r.t. versioning of databases

emturner:
Relates to RV-996

# What

Record an explanation of how we can re-use the byte for 'tree-is-empty-or-populated' for versioning in future.

# Why

Ensure we don't forget with an explicit TODO, while allowing us to de-prioritise the work to add explicit versioning.

### DIFF
--- a/durable-storage/src/avl/tree.rs
+++ b/durable-storage/src/avl/tree.rs
@@ -80,6 +80,8 @@ impl<NodeId: FromProof> Tree<NodeId> {
             }
             Partial::Blinded(hash) => Ok((ctx, Partial::Blinded(hash))),
             Partial::Present(()) => {
+                // TODO (RV-996): leverage <bool> as a union tag to allow future versioning
+                //                of the database/merkle layer in a backwards-compatible way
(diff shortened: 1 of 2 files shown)


durable-storage/src/avl/tree.rs
```diff
@@ -80,6 +80,8 @@ impl<NodeId: FromProof> Tree<NodeId> {
             }
             Partial::Blinded(hash) => Ok((ctx, Partial::Blinded(hash))),
             Partial::Present(()) => {
+                // TODO (RV-996): leverage <bool> as a union tag to allow future versioning
```
victor-dumitrescu, line 83: The Linear issue still mentions that Databases should include their version and that it is in progress, so should probably be updated to explain the proposed enum mechanism?

victor-dumitrescu (approved):

thomasathorne (approved):
Modulo Victor's comment on the issue RV-996 needing to be rewritten.

durable-storage/src/avl/tree.rs
emturner: done

PR #1142, opened 2026-07-17
feat(nds/test): inline proof into step in db regression; assert mode outcomes identical

emturner:
Follow up to #1141

# What

Inline proof into step in the database proof-trace. Additionally ensure that all modes return the same `StepOutcome`, including the model

# Why

The proof-trace regression was confusing as each operation would appear twice. Additionally, we need to ensure step outcomes are the same across modes, not just the resulting root hashes.

# Manually Testing

```
make all
```

# Regressions

database regression for proof-trace has been regenerated - proof is now inlined into the step

# Tasks for the Author


durable-storage/src/long_test/database/run_case.rs
```diff
@@ -74,7 +74,7 @@ fn apply_sequence(targets: &mut Targets, ops: &[DatabaseOperation], prove: bool)
     for op in ops {
         // Proofs are taken over the pre-operation state, so prove first.
         if prove {
-            prove_and_verify_database_operation(&targets.persistent_db, op);
+            prove_and_verify_database_operation(targets.persistent_db.inner(), op);
```
victor-dumitrescu: The `StepOutcome` which is now returned here could also be compared to that of the Normal mode application below. Otherwise I don't think the cross-mode outcome comparisons extend to the long tests at the moment.

durable-storage/src/test_helpers/database.rs
```diff
@@ -445,137 +445,124 @@ impl<KV: BackgroundPersistentKeyValueStore> DatabaseOps<KV> for TracedDatabase<K
...
-                    result.is_ok(),
-                    "Write should have succeeded but failed: {:?}",
-                    result.err()
-                );
-            } else {
-                assert!(result.is_err(), "Write should have failed but succeeded");
-            }
+            assert_eq!(
+                should_succeed,
+                result.is_ok(),
```
victor-dumitrescu, line 492: Not directly related to the PR, but I realised here we can tighten the check so that in case of success it also asserts the same number of bytes were written.

durable-storage/src/long_test/database/run_case.rs
emturner: done, [diff](https://github.com/tezos/riscv-pvm/compare/ade8fbe3881911671852fbeae5f11072e7faf190..db8f6418eb6b8338b79591f5bb281f039c4d4fec)

durable-storage/src/test_helpers/database.rs
emturner: done, [diff](https://github.com/tezos/riscv-pvm/compare/ade8fbe3881911671852fbeae5f11072e7faf190..db8f6418eb6b8338b79591f5bb281f039c4d4fec)

victor-dumitrescu (approved):

jobjo (approved):

PR #1144, opened 2026-07-20
feat(durable): add avl_storage bench

emturner:
- Relates to TZX-194

# What

Adds a new avl-storage bench, that targets the lazy resolver - cold from disk

# Why

Demonstrates subsequent optimisations have an impact

# How

Adds a bench

# Manually Testing

```
```

# Tasks for the Author

### DIFF
--- a/durable-storage/Cargo.toml
+++ b/durable-storage/Cargo.toml
@@ -87,6 +87,10 @@ name = "avl_tree"
 harness = false
 
 [[bench]]
+name = "avl_storage"
+harness = false
+
+[[bench]]
 name = "database"
 harness = false
 required-features = ["rocksdb"]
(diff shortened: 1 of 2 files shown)


victor-dumitrescu (commented):
Running this benchmark, I'm seeing pretty big variability between runs on my machine (+/- 10-15%). I think this comes down to how samples are generated, I imagine different keys and operations produce very different trees in which as the run progresses the number of nodes which need to be resolved varies a lot. Indeed, trying this with a deterministic RNG running with the same seed each time reduces the variability significantly. I guess whether this is relevant depends on how big an effect the subsequent optimisations are expected to have.

LGTM otherwise!

emturner:
@victor-dumitrescu could you try now? Have seeded but still seeing +/- 3% variability. But this is much smaller than the optimisation speedup, so will be very obvious either way

victor-dumitrescu:
> @victor-dumitrescu could you try now? Have seeded but still seeing +/- 3% variability. But this is much smaller than the optimisation speedup, so will be very obvious either way

Same, seeing < +/- 2% now.

victor-dumitrescu (approved):

jobjo (approved):

PR #1149, opened 2026-07-22
feat(dursto): add upper-bound proof size model and check in tests

victor-dumitrescu:
Part of TZX-188.

# What

Introduces a proof size model, which, given the reference model for a Database or Registry and an operation on that Database or Registry, models the maximum expected size for the proof of that operation. 

An extra check is introduced in both regular and long tests: every proof is smaller than both its modelled bound and 16 KiB.

# Why

To introduce a meaningful check that produced proofs have the expected size and to model what we expect to be the least well-behaved proofs.

# How

The model makes the least favourable assumptions for every operation, for both Registry and Database: maximum size for keys and stored values, longest unblinded paths in the tree, no path sharing, maximum number of rotations where applicable. The goal is for the model to be a realistic, strict upper bound on proof size, against which we can compare actual proofs during testing.

The tightness of the bound was evaluated on a run of each of the Database (blue) and Registry (red) long tests. Every point can correspond to multiple samples.

image

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/bin/long_test.rs
+++ b/durable-storage/src/bin/long_test.rs
@@ -73,6 +73,11 @@ struct TestArgs {
     /// Directory for run state and failure artifacts (default: a fresh tempdir).
     #[arg(long)]
     out_dir: Option<PathBuf>,
+
+    /// Fail the test when a proof size exceeds the maximum, instead of printing
+    /// a warning.
+    #[arg(long)]
+    fail_on_warning: bool,
 }
 
 #[derive(Debug, Subcommand)]
@@ -131,6 +136,7 @@ fn config(args: TestArgs) -> Result<LongTestConfig> {
         epochs: args.epochs,
         keep_epochs: args.keep_epochs,
         out_dir: args.out_dir,
+        fail_on_warning: args.fail_on_warning,
     })
 }
 
--- a/durable-storage/src/test_helpers/proof_size.rs
+++ b/durable-storage/src/test_helpers/proof_size.rs
@@ -1,0 +1,387 @@
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! Model for upper bound on Regsitry and Database proof size
+//!
+//! Every Database proof is checked against the worst case model for the current
+//! number of keys in tree shape and key sizes: the deepest AVL shape (a
+//! Fibonacci-minimal tree), every key at [`KEY_MAX_SIZE`], and the most
+//! expensive rebalancing per operation type (one rotation for an insert,
+//! `floor((depth - 1) / 2)` for a delete). Operation value sizes are real:
+//! the lengths and the byte ranges an operation touches are taken from
+//! the [`DatabaseReferenceModel`].
+//!
+//! The bounds mirror the proof encoding: pre-order tree with a one-byte tag per
+//! node or leaf, a `Blind` leaf carrying a hash and a `Read` leaf carrying raw
+//! bytes (see the `Encode` impl in `octez_riscv_data::merkle_proof::proof_tree`).
+
+use std::collections::BTreeSet;
+use std::ops::Range;
+
+use octez_riscv_data::components::bytes::NODE_ARITY as BYTES_NODE_ARITY;
+use octez_riscv_data::components::bytes::PAGE_SIZE;
+use octez_riscv_data::components::vector::NODE_ARITY as VECTOR_NODE_ARITY;
+use octez_riscv_data::foldable::seq_tree::tree_depth;
+use octez_riscv_data::hash::Hash;
+
+use crate::key::KEY_MAX_SIZE;
+use crate::test_helpers::database::DatabaseOperation;
+use crate::test_helpers::database::DatabaseReferenceModel;
+use crate::test_helpers::registry::RegistryOperation;
+
+/// Serialised size of a proof tree tag. Tags are written as one byte each by the
+/// `Encode` impl of `octez_riscv_data::merkle_proof::proof_tree::MerkleProof`.
+const TAG_BYTES: usize = 1;
+
+/// A `Blind` leaf: tag plus [`Hash::DIGEST_SIZE`] bytes.
+const BLIND_LEAF: usize = TAG_BYTES + Hash::DIGEST_SIZE;
+
+/// An accessed AVL tree wrapper: proof node tag plus a `Read` leaf holding the
+/// one-byte present flag (see `fold_resolved_tree` in `crate::merkle_layer`).
+const TREE_WRAP: usize = TAG_BYTES + TAG_BYTES + size_of::<bool>();
+
+/// A `Read` leaf holding a `u64` length (fixed-int bincode encoding).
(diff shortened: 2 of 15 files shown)


durable-storage/src/test_helpers/proof_size.rs
```diff
@@ -0,0 +1,386 @@
...
+
+/// A `Read` leaf holding a `u64` length (fixed-int bincode encoding).
+const LEN_LEAF: usize = TAG_BYTES + size_of::<u64>();
+
+/// A `Read` leaf holding a node's `i64` balance factor (fixed-int bincode
+/// encoding).
+const BALANCE_FACTOR_LEAF: usize = TAG_BYTES + size_of::<i64>();
+
+/// A `Read` leaf holding a node's [`Key`](crate::key::Key): a one-byte length
+/// prefix plus the key bytes (see the `Encode` impl in `crate::key`).
```
emturner:
```suggestion
/// A `Read` leaf holding a node's [`Key`]: a one-byte length
/// prefix plus the key bytes (see the `Encode` impl in `crate::key`).
///
/// [`Key`]: crate::key::Key
```

(mega nit) - avoiding the inline link makes it easier to read when just perusing the docs in the editor? feel free to resolve though

durable-storage/src/test_helpers/proof_size.rs
```diff
@@ -0,0 +1,386 @@
...
+fn key_leaf(key_len: usize) -> usize {
+    TAG_BYTES + 1 + key_len
```
emturner, line 57: nit - `assert(key_len < KEY_MAX_SIZE)` perhaps?

durable-storage/src/test_helpers/proof_size.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+/// An accessed node on the search path: tree wrapper, node tag, balance factor
+/// and key leaves, blinded data and one blinded sibling subtree. The balance
+/// factor is shorter than a hash so it is always included directly; the key
+/// may instead be blinded at [`BLIND_LEAF`], hence the `max`. The subtree the
+/// path continues into is charged by its own level; the terminal node's second
+/// child is charged via [`terminal_blind`].
+fn avl_path_node(key_len: usize) -> usize {
+    TREE_WRAP + TAG_BYTES + BALANCE_FACTOR_LEAF + key_leaf(key_len).max(BLIND_LEAF) + 2 * BLIND_LEAF
```
emturner: why would the key be blinded? Currently, we don't compare by hashed key - if the key is short you'd still expect it be included in full I think?

durable-storage/src/test_helpers/proof_size.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+/// An accessed node off the search path (touched by a rebalancing rotation):
+/// like [`avl_path_node`], but both children may be blinded.
+fn avl_extra_node() -> usize {
+    avl_path_node(KEY_MAX_SIZE) + BLIND_LEAF
```
emturner, line 74: here though, the key you'd expect to be blinded I think? so not sure this is quite right - would I think imply the max bound is currently higher than it could be?

durable-storage/src/test_helpers/proof_size.rs
```diff
@@ -0,0 +1,386 @@
...
+/// step.
+pub(crate) fn database_operation_proof_size_bound(
+    model: &impl DatabaseReferenceModel,
+    op: &DatabaseOperation,
+) -> Option<usize> {
+    let key = match op {
+        DatabaseOperation::Commit
+        | DatabaseOperation::Checkout
+        | DatabaseOperation::CommitCheckoutRoundtrip => return None,
+        DatabaseOperation::Hash => None,
```
emturner:
```suggestion
DatabaseOperation::Hash => return Some(BLIND_LEAF),
```

Then you don't need the `if let` below?

durable-storage/src/bin/long_test.rs
```diff
@@ -73,6 +73,11 @@ struct TestArgs {
     /// Directory for run state and failure artifacts (default: a fresh tempdir).
     #[arg(long)]
     out_dir: Option<PathBuf>,
+
+    /// Fail the test when a proof size exceeds the maximum, instead of printing
+    /// a warning.
+    #[arg(long)]
+    fail_on_warning: bool,
```
emturner, line 80: could this maybe be `fail_on_proof_bound_exceeded`? I know it's more verbose, but `warning` is v generic

durable-storage/src/test_helpers/proof_size.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/f3d7dc669c6564e7058d0cc6fa5b7bfdf725ae3a..508c0104ead1f4f5f12e3414ca296961129d567b).

durable-storage/src/test_helpers/proof_size.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/f3d7dc669c6564e7058d0cc6fa5b7bfdf725ae3a..508c0104ead1f4f5f12e3414ca296961129d567b).

durable-storage/src/test_helpers/proof_size.rs
victor-dumitrescu: Indeed, fixed [here](https://github.com/tezos/riscv-pvm/compare/f3d7dc669c6564e7058d0cc6fa5b7bfdf725ae3a..508c0104ead1f4f5f12e3414ca296961129d567b).

durable-storage/src/test_helpers/proof_size.rs
victor-dumitrescu: Looking at it again given the comment above, I think keys are never actually blinded in present nodes?

durable-storage/src/test_helpers/proof_size.rs
victor-dumitrescu: Done [here](https://github.com/tezos/riscv-pvm/compare/f3d7dc669c6564e7058d0cc6fa5b7bfdf725ae3a..508c0104ead1f4f5f12e3414ca296961129d567b).

durable-storage/src/bin/long_test.rs
victor-dumitrescu: I made it generic because I thought other warnings beside proof size might be raised in the future for things like verification time or allocations for example. Although maybe we'll treat those differently, for example recording the proof which raised the warning rather than potentially failing the whole test. Let me know what you think.

durable-storage/src/test_helpers/proof_size.rs
emturner: My understanding is that this is for a node 'off the search path'? ie we only look at the balance factor of the node?

emturner: No, you're right - it is read but shouldn't be (it's actually a proof-generation artifact). 

I'll take a look

emturner (approved):

jobjo (approved):

PR #1150, opened 2026-07-24
feat(nds): reduce resolution lookups and overall disk size for databases

emturner:
- closes TZX-194

# What

 Replace the `tree_hash -> tree_content` and `node_hash -> node_content` mappings in the merkle column families, with just `tree_hash -> node_content` directly.

# Why

This saves a lookup whenever resolving a node. Additionally it saves disk space as an intermediate mapping is skipped.

Tree's have two variants: 
- `Tree(None)` (the empty tree)
- `Tree(Some(Node))` (populated trees)

# How

The empty tree variant is never persisted, as the hash is well known. 

Therefore all persisted trees are populated. All trees persisted have content `<true><node hash>`. Therefore, we can skip this intermediate mapping as we can recalculate the tree hash from just the child-node content.

This does require deleting `shallow_options` (at least for tree's - as they now must always persist the node). We _could_ keep the option for nodes, but as this option was test only (and only used in a couple places) I felt it made sense to just remove it entirely.

# Manually Testing

```
make all

# demonstrate the speedup - run this on `main` first.
# My machine - see a >60% improvement
```

# Regressions

None. We currently do not have regression tests for snapshots/databases.

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -869,7 +869,8 @@ where
         options: &StoreOptions,
     ) -> Result<(), OperationalError> {
         // The stored representation is more compact. We don't include the `data` field, as that
-        // should be written to the KV store separately.
+        // should be written to the KV store separately. `left`/`right` are the hashes of the
+        // child *trees* (the same values folded into this node's hash).
         let repr = StoredNode {
             balance_factor: self.balance_factor.read(),
             key: self.key.deref().clone(),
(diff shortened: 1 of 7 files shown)


victor-dumitrescu (approved):
LGTM. I can observe the benchmarking boost locally.

I ran an AI review which was confused by the existence of

```rust
impl From<Hash> for LazyNodeId {
    fn from(hash: Hash) -> LazyNodeId {
        LazyNodeId(hash.into())
    }
}
```

since it can result in resolving a node changing its hash (it thought it needs to be updated to reflect the changes in this PR). But this is only used in tests, so should probably just be cfg-gated?

PR #1157, opened 2026-07-28
feat(nds): allow Registry to checkout in read-only mode

emturner:
- Part of TZX-201

# What

Split registry/database functionality over `ReadableKeyValueStore` vs `WriteableKeyValueStore`.

# Why

This enables an explicit path to checking out a registry in read-only mode, without the need for an intermediate mutable checkpoint if it's not required. This use-case is directly aimed at the Rollup Node RPC's.

# How

Split `KeyValueStore` into readable vs writeable functionality. By leveraging constraints, we can ensure that any database/registry backed by `ReadOnlyKeyValueStore` never tries to modify this or otherwise perform any mutable operations.

# Manually Testing

```
make all
```

# Tasks for the Author


durable-storage/src/persistence_layer.rs
```diff
@@ -206,6 +209,71 @@ pub(crate) fn rocksdb_checkpoint_options() -> rocksdb::Options {
...
+        None => Err(InvalidArgumentError::KeyNotFound)?,
+    }
+}
+
+/// Retrieve the value associated with a key from `db`.
+fn get_from<'db>(db: &'db rocksdb::DB, key: &[u8]) -> Result<rocksdb::DBPinnableSlice<'db>, Error> {
+    let value = db
+        .get_pinned(key)
+        .map_err(|error| OperationalError::GetFailed {
+            column: "default".to_owned(),
```
victor-dumitrescu: Just noticed we hardcode `"default"` in a few places, perhaps it should be its own constant instead?

durable-storage/src/persistence_layer.rs
```diff
@@ -790,6 +878,140 @@ mod tests {
...
+            .expect("Commit directory should be readable")
+            .map(|entry| {
+                let entry = entry.expect("Directory entry should be readable");
+                let len = entry.metadata().expect("Metadata should be readable").len();
+                (entry.file_name(), len)
+            })
+            .filter(|(name, _len)| !name.to_string_lossy().starts_with("LOG"))
+            .collect();
+
+        contents.sort();
```
victor-dumitrescu: Why do we need to sort?

durable-storage/src/persistence_layer.rs
victor-dumitrescu: Another thing that could happen is the same data being committed unmodified again, in which case deleting and rewriting is a bit wasteful + now I think it could result in a situation where a read-only checkout (briefly) points to missing data?

durable-storage/src/persistence_layer.rs
```diff
@@ -206,6 +209,71 @@ pub(crate) fn rocksdb_checkpoint_options() -> rocksdb::Options {
...
+
+    match value {
+        Some(value) => Ok(value),
+        None => Err(InvalidArgumentError::KeyNotFound)?,
+    }
+}
+
+/// Open the database committed at `commit_path` in read-only mode.
+///
+/// A read-only instance does not acquire the directory's lock, replay its write-ahead log or
```
victor-dumitrescu: Given that the `LOG*` files are still being written to, that's not strictly speaking true, right? Could that result in any issues?

durable-storage/src/registry.rs
```diff
@@ -66,26 +69,22 @@ impl<KV: BackgroundReadableKeyValueStore> Registry<KV, Normal> {
     ///
     /// The registry owns a Tokio [`Runtime`] and a register state repository.
     pub fn new(repo: KV::Repo) -> Result<Self, OperationalError> {
```
victor-dumitrescu: This function is now infallible.

durable-storage/src/registry.rs
```diff
@@ -210,6 +227,83 @@ where
...
+        KV::Writeable: BackgroundPersistentKeyValueStore,
+    {
+        let NormalImpl { repo, runtime } = &self.inner;
+        let handle = runtime.handle()?;
+
+        let databases = self
+            .databases
+            .iter()
+            .map(|db| db.to_writeable(handle, repo))
+            .collect::<Result<_, _>>()?;
```
victor-dumitrescu, line 308: `TODO RV-946` would be applicable here too

emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/7912834ee09506362c50f259a2119c21ace1a5ab..7a7f2b22786e06bfa72a4f2a663b2deec4e7c132)

durable-storage/src/registry.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/7912834ee09506362c50f259a2119c21ace1a5ab..7a7f2b22786e06bfa72a4f2a663b2deec4e7c132)

durable-storage/src/persistence_layer.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/7912834ee09506362c50f259a2119c21ace1a5ab..7a7f2b22786e06bfa72a4f2a663b2deec4e7c132)

durable-storage/src/persistence_layer.rs
emturner: Really good spot. The comment here was actually pretty misleading overall.

Broken out this change into it's own PR [riscv#1176](https://github.com/tezos/riscv-pvm/pull/1176), which this now targets.

plus [diff](https://github.com/tezos/riscv-pvm/compare/7a7f2b22786e06bfa72a4f2a663b2deec4e7c132..578f50fdabb4eac8339ec8bb9086ba8fc329031c) - but most of this is just the commit that's broken out into the other PR

durable-storage/src/persistence_layer.rs
emturner: ah, turned out that commits themselves never contain the INFO LOG entries - so this no longer filters them

[diff](https://github.com/tezos/riscv-pvm/compare/f86a20af05fda26c86b0ac39782ac9bfd3745773..e6588ca2093a1cf563094a2811314e3e43d846be)

emturner:
Additionally, [diff](https://github.com/tezos/riscv-pvm/compare/f86a20af05fda26c86b0ac39782ac9bfd3745773..e6588ca2093a1cf563094a2811314e3e43d846be) fixes an issue with lazy runtime contention

victor-dumitrescu (approved):

PR #1158, opened 2026-07-28
refactor(nds): split KeyValueStore into writeable and readable

emturner:
- part of TZX-201

# What

Split `KeyValueStore` into `ReadableKeyValueStore` and `WriteableKeyValueStore`

Constrain every path that only reads the store - or only mutates the in-memory Merkle tree - from `KeyValueStore` to `ReadableKeyValueStore`

# Why

No functional changes here - but subsequently unblocks adding a full `checkout_readonly` method, whilst ensuring that such a registry _cannot_ mutate the underlying store. This makes it safe to open a checkout directly, rather than going through an intermediate copy: and ultimately allowing the rollup node's RPC flow to be more efficient - as it only requires read-access to the store

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/merkle_layer.rs
+++ b/durable-storage/src/merkle_layer.rs
@@ -15,6 +15,12 @@
 //!
 //! [`MerkleLayer::clone_with`] enables forking snapshots. Clones share the underlying tree
 //! cheaply via an `Arc` and diverge upon mutation, using copy-on-write (CoW) semantics.
+//!
+//! *NB* perhaps counterintuitively, all operations (other than `commit`) only require the underlying
+//! KV store be _readable_, even for operations that mutate the AVL tree. This occurs because all these
+//! operations, including mutable ones, only ever load data from the store. The reason `commit` requires
+//! the store to be _writeable_, is precisely because this is the only place where the (possibly updated)
+//! in-memory store is persisted back to the underlying store.
 
 use std::convert::Infallible;
 use std::marker::PhantomData;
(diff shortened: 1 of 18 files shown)


durable-storage/src/merkle_layer.rs
```diff
@@ -5,7 +5,7 @@
 
 //! A Merkleised key-value store layer.
 //!
-//! [`MerkleLayer`] wraps a [`KeyValueStore`] (KV) and duplicates all stored data in an AVL
+//! [`MerkleLayer`] wraps a [`ReadableKeyValueStore`] (KV) and duplicates all stored data in an AVL
```
thomasathorne: This is unintuitive for me, seems like the `MerkleLayer` (at least occasionally) needs to write to the key value store too? It's also not strictly true as the definition of `MerkleLayer` doesn't specify any constraints on `KV` and there are places (at least in the tests below) where a merkle layer is wrapping a writeable key value store.

Basically I think there could me more clarity over exactly when and why the merkle layer needs each of the two traits.

thomasathorne (approved):
Generally looks good, I'm a little confused about how the `MerkleLayer` uses the new traits and think this could be better explained in comments.

durable-storage/src/merkle_layer.rs
emturner: that's fair, I'll add some comments.

It essentially comes down to anything that only does 'loading' from the database (which is all operations normally) only requires that the database be readable, and doesn't need to be writeable.

The _only_ operation that merkle layer does that requires the store to be writeable is `commit`

emturner: Pushed in [diff](https://github.com/tezos/riscv-pvm/compare/3655e8ac72bb361376d167ca12d96332e6eb5757..ebad6442e8400ad923069c14251f7898bbe4e379) (apologies - mixed the diff up with the rebase onto the page-size reduction)

thomasathorne: Looks good, much clearer now thanks! 🥳

PR #1159, opened 2026-07-28
feat(data): reduce bytes node_arity to 2

emturner:
- closes TZX-141

# What

Follow up to #1156 

Reduces the arity of `Bytes` merkleisation from `4 -> 2`.

# Why

Proof-size is improved in the worst case for bytes. Further alignement with future `blake3` switch.

# How

Update constant + regression tests. *NB* there is now only one location where the worst-case scenario occurs, rather than 3, due to the reduction in arity.

# Manually Testing

```
make all
```

# Regressions

Root hashes of anything using `bytes` has changed (where `bytes` was larger than `2048` in length).

# Tasks for the Author


emturner:
interesting discovery: we have no regression tests covering a `bytes.len() > 2048` - hence why db/registry regressions are unchanged

data/src/components/bytes/test_utils.rs
```diff
@@ -128,21 +128,16 @@ pub const MONOFURCATED_LAYER: usize = (NODE_ARITY - 2) * Hash::DIGEST_SIZE + NOD
...
 
-/// There are three offsets at which the worst possible boundary between two pages occurs---at the
-/// 1st, 2nd and 3rd quartile in the `Bytes` component.
-pub const MAX_PROOF_OFFSETS: [usize; 3] = [
-    NDS_BYTES_LENGTH / 4 - 1,
-    (NDS_BYTES_LENGTH / 4) * 2 - 1,
-    (NDS_BYTES_LENGTH / 4) * 3 - 1,
-];
+/// There is one offset at which the worst possible boundary between two pages occurs---half way in the `Bytes` component.
+pub const MAX_PROOF_OFFSETS: [usize; NODE_ARITY - 1] = [(NDS_BYTES_LENGTH / 2) - 1];
```
thomasathorne: Seems a bit silly having an array when there is just one? Would just be a few more lines changed to switch to a bare `usize`.

thomasathorne (approved):

data/src/components/bytes/test_utils.rs
emturner: no that's fair, I was more just trying to demonstrate why it's changed.

emturner: it's now an array of 2!

data/tests/goldenfiles/bytes_test_utils_constants
```diff
@@ -1,5 +1,5 @@
 NDS_BYTES_LENGTH = 67108864
-NDS_BYTES_TREE_DEPTH = 8
-MONOFURCATED_LAYER = 68
-BIFURCATED_LAYER = 200
-MAX_PROOF_LENGTH = 4543
+NDS_BYTES_TREE_DEPTH = 16
+MONOFURCATED_LAYER = 2
```
thomasathorne, line 3: Could add `TRIFURCATED_LAYER` (or `ASYMMETRIC_BIFURCATED_LAYER` or whatever else it gets renamed to) here?

thomasathorne (approved):

data/tests/goldenfiles/bytes_test_utils_constants
emturner: done in [diff](https://github.com/tezos/riscv-pvm/compare/81395ce729727ce67ee6f3bfc1a1f55e656ce240..dc4c522cd80d781f58867583f4cc9edfa4966621) - also captured in the previous commit

PR #1161, opened 2026-07-30
test(nds): single-key counterfeit DBs

thomasathorne:
# What

Introduces a function `single_key_counterfeit` that produces a normal mode database which is mostly fake, it only includes just enough real structure to support a single node operation (eg. read or write) on a key placed a specific depth.

Adds a property test to show that (if key sizes are all identical---we use 32 bytes) the size of proofs for a read operation is precisely linear in the depth, with 113 extra bytes for each extra layer the AVL tree. The test ranges from depth 0 to depth 500 trees, demonstrating that the stack overflow is unlikely to be a problem---it's only at depth greater than 500 that this starts to happen, at which point the proof size (for the database component alone) is more than 50 kB.

# Why

Begin testing proof verification on extremely deep trees using counterfeits.

Begin work towards establishing a precise model for minimum and maximum proof size as a function of AVL tree depth.

Demonstrate that stack overflow isn't something we need to worry about (at least for the database component and depth of AVL tree).

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/test_helpers/counterfeit.rs
+++ b/durable-storage/src/test_helpers/counterfeit.rs
@@ -1,0 +1,280 @@
+// SPDX-FileCopyrightText: 2026 Trilitech <contact@trili.tech>
+//
+// SPDX-License-Identifier: MIT
+
+//! Utilities for proof safety testing that allow generation of proofs for NDS instances that
+//! cannot be created in the normal way, because they are too large, or may even be invalid. We
+//! call such databases 'counterfeits'.
+
+#![cfg(test)]
+
+use std::sync::Arc;
+
+use bytes::Bytes;
+use octez_riscv_data::hash::Hash;
+use octez_riscv_data::mode::Normal;
+use tokio::runtime::Handle;
+
+use crate::avl::node::Node;
+use crate::avl::resolver::LazyDataId;
+use crate::avl::resolver::LazyNodeId;
+use crate::avl::resolver::LazyResolver;
+use crate::avl::resolver::LazyTreeId;
+use crate::avl::tree::Tree;
+use crate::database::Database;
+use crate::key::Key;
+use crate::merkle_layer::MerkleLayer;
+use crate::merkle_worker::BackgroundWriteableKeyValueStore;
+use crate::merkle_worker::MerkleWorker;
+
+/// To build a valid-seeming counterfeit we need the sequence of nodes (built from the deepest
+/// layer upwards) to have `Key`s that appear in a sensible order.
+///
+/// This utility function can be given any slice of nodes where a node is represented by its `Key`
+/// and some other data. It will sort the slice around a given 'pivot' `key` so that the keys are
+/// monotonically (in both directions) further from the pivot. This slice can then be passed to
+/// `single_key_counterfeit` and it will produce a counterfeit with keys that respect the tree
+/// shape.
+///
+/// For every index in the slice, the reordered node will be below or above the pivot key depending
+/// on the node at that index before sorting. This means that randomly generating sequences of keys
+/// will randomly generate different tree shapes, while ensuring a uniform distribution of keys.
+///
+/// This approach has some nice properties: it means that if the randomly generated pivot key is
+/// very small, the resulting tree shape will be left-leaning in a natural way. (In particular, the
+/// probability of generating a completely left or right-leaning tree is not negligible. If we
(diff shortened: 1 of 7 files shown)


durable-storage/src/test_helpers/counterfeit.rs
```diff
@@ -0,0 +1,276 @@
...
+use crate::merkle_layer::MerkleLayer;
+use crate::merkle_worker::BackgroundWriteableKeyValueStore;
+use crate::merkle_worker::MerkleWorker;
+
+/// To build a valid-seeming counterfeit we need the sequence of nodes (built from the deepest
+/// layer upwards) to have `Key`s that appear in a sensible order.
+///
+/// This utility function can be given any slice of nodes where a node is represented by its `Key`
+/// and some other data. It will sort the slice around a given 'pivot' `key` so that the keys are
+/// monotically (in both directions) further from the pivot. This slice can then be passed to
```
jobjo:
```suggestion
/// monotonically (in both directions) further from the pivot. This slice can then be passed to
```

thomasathorne: Good spot! 🤦‍♂️ Changed it [here](https://github.com/tezos/riscv-pvm/compare/0a8fcc9ea5a87357f3ea2336a44bbae30716e809..92161ecb3c4ea30ab3bf1af0c54e2b9d48fe01d3).

durable-storage/src/test_helpers/counterfeit.rs
```diff
@@ -0,0 +1,276 @@
...
+                } else {
+                    assert!(*k <= bottom_bound);
+                    bottom_bound = k.clone();
+                }
+            }
+
+            assert_same_set(nodes, sorted_nodes);
+        }
+
+        #[test]
```
jobjo, line 206: Consider a comment to explain the test.

```suggestion
        /// Builds counterfeit databases with the target key at arbitrary depths,
        /// then checks that read proofs verify and grow by a fixed number of
        /// bytes per layer.
        #[test]
```

durable-storage/src/test_helpers/counterfeit.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    /// The length of the part of the proof which is present regardless of the depth of the tree.
+    const ZERO_DEPTH_PROOF_LENGTH: usize = 188;
+
+    /// The extra length of proof added for each extra layer of depth.
+    const PROOF_LENGTH_PER_LAYER: usize = 113;
+
+    proptest! {
+        #[test]
+        fn test_sort_nodes(nodes in collection::vec(node_strategy(), 0..500), key in key_strategy()) {
```
jobjo, line 180: Nit but is the `test_` prefix superfluous here?

jobjo (approved):
LGTM. Comments/suggestions are optional.

durable-storage/src/test_helpers/counterfeit.rs
thomasathorne: I think the `test_` prefix is standard for all of our tests. Perhaps a little superfluous but it helps disambiguate, for example the function `sort_nodes` is in the module above, so if you did `use super::sort_nodes` that would then collide with this test. 🤷‍♂️

If anything my mistake was in _not_ putting the prefix on the test below, which I have fixed [here](https://github.com/tezos/riscv-pvm/compare/92161ecb3c4ea30ab3bf1af0c54e2b9d48fe01d3..f1ef79e3faa03da7f27bda986bf72212f718af12).

durable-storage/src/test_helpers/counterfeit.rs
thomasathorne: Added a docstring [here](https://github.com/tezos/riscv-pvm/compare/f1ef79e3faa03da7f27bda986bf72212f718af12..8e66418ec7ea0e385e9fe863d8f51b8867196ea5).

thomasathorne:
Thanks for the review! I will leave off marking this as 'merge when ready' because there's one specific part I think should get another pair of eyes: basically, I was a little surprised when I found that the stack overflow didn't occur for proof verification at depths between 250 and 500.

I assume it is because the main thread has a larger stack size by default (according to the rust `std::thread` docs this isn't determined by rust though), or some other stack configuration difference. Anyway, just to sanity check this one point it might be worth having two reviews. Happy for either Emma or Victor to shepherd the PR to being merged while I'm off.

emturner (approved):

PR #1176, opened 2026-08-18
feat(nds): prevent database commit re-publishing

emturner:
- Related to TZX-201
- Broken out from [riscv#1157#discussion_r3803953610](https://github.com/tezos/riscv-pvm/pull/1157#discussion_r3803953610) (Read-ony checkout PR - NB a couple of references to read-only checkouts are included here as a result).

# What

Don't re-write perfectly good commits on disk that already existed.

# Why

Previously, a commit that already existed would be fully overwritten in a 'last-write-wins' scenario. This is both unecessary - as the use (repo) stores databases by their root hash, so a second write is unecessary. Before, this would remove perfectly good inodes, and fully replace the directory, possibly reducing sharing in certain contexts.

This change also makes commit publishing fully atomic - meaning data races (two databases with the same root hash) is no longer a concern.

It makes the subsequent read-only checkout mechanism much safer, as we never accidentally remove a commit from under a read-only checkout, only to write the same things back.

# How

Don't overwrite a commit if it's already there. Distinguish between committing on the rocks DB level, and 'publishing' (ie renaming the commit into the correct place). This rename is atomic - and therefore a rename that failed because the destination exists, is safely due to a different DB in the registry having the same root hash - and thus publishing first.

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/persistence_layer.rs
+++ b/durable-storage/src/persistence_layer.rs
@@ -206,6 +206,50 @@ pub(crate) fn rocksdb_checkpoint_options() -> rocksdb::Options {
     options
 }
 
+/// Whether `commit_path` holds a published commit.
+///
+/// A commit is published by renaming a complete checkpoint into place, so the `CURRENT` file that
+/// every RocksDB directory holds is enough to tell a published commit from a directory left behind
+/// by an interrupted removal.
+fn is_published_commit(commit_path: &Path) -> bool {
+    commit_path.join("CURRENT").exists()
+}
+
+/// The outcome of an attempt to publish a staged checkpoint under a commit path.
+enum Publish {
+    /// The commit is published - either by this attempt, or by a concurrent one, which published
+    /// the same state.
+    Done,
+    /// Something that is not a published commit occupies the commit path, holding the error the
+    /// attempt failed with.
+    Occupied(std::io::Error),
+}
+
+/// Move the checkpoint staged at `staged` into place under `commit_path`.
+///
+/// The rename is the only authority on whether the commit is published: it is a single step, so a
+/// concurrent publisher of the same commit either lost to it or won it outright. Nothing observed
+/// beforehand can be relied on to still hold by the time it runs.
+fn publish_staged_commit(staged: &Path, commit_path: &Path) -> Result<Publish, OperationalError> {
+    match std::fs::rename(staged, commit_path) {
+        Ok(()) => Ok(Publish::Done),
+
+        // Lost a race to publish this commit. The winner published the state staged here - the
+        // commit id is a hash of it - so the commit is made either way, and the staged copy is
+        // discarded along with its staging directory.
+        Err(_) if is_published_commit(commit_path) => Ok(Publish::Done),
+
+        // A directory in the way is the one failure the caller can repair.
+        Err(error) if commit_path.is_dir() => Ok(Publish::Occupied(error)),
+
+        Err(error) => Err(OperationalError::CommitPublishFailed {
+            staged: staged.to_owned(),
+            commit: commit_path.to_owned(),
+            error,
+        }),
+    }
+}
+
 /// Persistence layer for durable solution used by the RISC-V PVM.
 ///
 /// Invariants:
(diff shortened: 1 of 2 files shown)


durable-storage/src/persistence_layer.rs
```diff
@@ -428,22 +437,55 @@ impl PersistentKeyValueStore for PersistenceLayer {
...
 
-        // If the path already exists, we overwrite the existing commit. This is highly unlikely to
-        // happen anyway if the commits are a hash of the content.
-        if Path::exists(&checkpoint_path) {
-            std::fs::remove_dir_all(&checkpoint_path).map_err(|error| {
+        // A directory holding no `CURRENT` file is not a published commit. A checkpoint is built
+        // under a temporary name and renamed into place, so an interrupted commit cannot leave one
+        // behind - only an interrupted removal by an older version of this function can, and
+        // replacing it is the repair.
+        if Path::exists(&commit_path) {
```
victor-dumitrescu: In theory a concurrent commit might have been finalised between a false`is_published_commit` check above and a true path check here, in which case it would be deemed as incomplete and removed.

durable-storage/src/persistence_layer.rs
```diff
@@ -428,22 +437,55 @@ impl PersistentKeyValueStore for PersistenceLayer {
...
+            log::warn!("Replacing incomplete commit: {}", id.hex_encode());
+
+            std::fs::remove_dir_all(&commit_path).map_err(|error| {
```
victor-dumitrescu: If `remove_dir_all` gets interrupted, it would leave a partial commit behind and it might not have deleted `CURRENT` yet. Should we make sure `CURRENT` is deleted first or maybe use the same kind of (un-)staging mechanism as for writing a commit, where a commit dir is first renamed?

durable-storage/src/persistence_layer.rs
```diff
@@ -206,6 +206,15 @@ pub(crate) fn rocksdb_checkpoint_options() -> rocksdb::Options {
     options
 }
 
+/// Whether `commit_path` holds a published commit.
+///
+/// A commit is published by renaming a complete checkpoint into place, so the `CURRENT` file that
+/// every RocksDB directory holds is enough to tell a published commit from a directory left behind
+/// by an interrupted removal.
+fn is_published_commit(commit_path: &Path) -> bool {
```
victor-dumitrescu, line 214: If the definition of a valid commit is now "it contains the `CURRENT` file", should this check be done when checking out too?

victor-dumitrescu (commented):
LGTM modulo a few comments.

Perhaps it would be good to add a test specifically for the case where the same state is committed twice, checking not only the commit directory contents but that the files themselves have not changed on disk between the commits?

durable-storage/src/persistence_layer.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/e830503a42e666a722d49bbe0c57e56d1055b07f..b1d6c3b1a3a191acbb1ccf33eae3a81770063fda)

durable-storage/src/persistence_layer.rs
emturner: good point - done. 

Additionally, after renaming (destaging) we check to make sure it hadn't become published in the meantime - restoring it if so done [diff](https://github.com/tezos/riscv-pvm/compare/e830503a42e666a722d49bbe0c57e56d1055b07f..b1d6c3b1a3a191acbb1ccf33eae3a81770063fda)

durable-storage/src/persistence_layer.rs
emturner: good spot, done [diff](https://github.com/tezos/riscv-pvm/compare/e830503a42e666a722d49bbe0c57e56d1055b07f..b1d6c3b1a3a191acbb1ccf33eae3a81770063fda)

emturner:
> LGTM modulo a few comments.
> 
> Perhaps it would be good to add a test specifically for the case where the same state is committed twice, checking not only the commit directory contents but that the files themselves have not changed on disk between the commits?

ah sorry, that test partially existed, but under the wrong name, fixed (and expanded) in [diff](https://github.com/tezos/riscv-pvm/compare/e830503a42e666a722d49bbe0c57e56d1055b07f..b1d6c3b1a3a191acbb1ccf33eae3a81770063fda)

victor-dumitrescu (approved):

PR #1178, opened 2026-08-21
test(nds): add metrics for disk-space usage

emturner:
- Relates to TZX-209

# What

`gc_space`, a harness that measures what a commit costs on disk and how much of that is still
needed: live against dead Merkle node data, what the retained history pins, and how much of each
commit the previous one already holds. `--simulate-dir-gc` removes every commit no retained root
reaches and reports what that frees.

Also fixes the long tests' per-epoch repository size, which summed file lengths and so counted a
hard-linked SST once per link, growing with the retention window rather than with the data.

Measurement only, gated behind `rocksdb` + `unstable-test-utils` — which is why the patch-coverage
check fails by construction.

There'll be follow ups to start running this in CI, and tracking in datadog

# Why

Of the bytes a commit occupies, how many are still needed to serve it? The two column families
answer differently. Values are keyed by your key, so an overwrite becomes an obsolete version that
compaction reclaims. Merkle node bodies are keyed by content hash, so every version is a distinct
live key that nothing ever deletes. The dead figure for the Merkle side is what justifies
collecting at all, and `--simulate-dir-gc` shows why deleting directories cannot recover it: those
dead bodies sit in files the retained commits still reference.

# How

Keys and values are derived from their indices, so a run at ten million keys holds no key list, and
the base state is recorded in `--repo-dir` and reused when the shape matches. Live bodies come from
walking the stored representation from the committed root; the totals they are subtracted from come
from a full column-family scan, since an estimate cannot be subtracted from an exact figure.
Occupancy counts blocks per unique inode, because checkpoints hard-link their SSTs — that walk is
what the long tests now share.

# Manually Testing

```
make -C durable-storage check

cargo run --release --features unstable-test-utils --bin gc_space -- \
  --databases 2 --keys 500 --commits 3 --modified-keys 20 --simulate-dir-gc
```

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -921,6 +921,48 @@ where
     }
 }
 
+/// Visit every node body reachable from the tree stored under `tree_hash`.
+///
+/// `visit` receives the blob key each body is stored under and the length of the stored bytes,
+/// once per body reached. Nothing is materialised in memory: a body is stored under the hash of
+/// the *tree* holding it, and its `left`/`right` fields are the hashes of the child trees, so the
+/// stored representation can be walked directly. Empty trees are never stored, so a child hash
+/// equal to [`Tree::empty_hash`] terminates that branch.
+///
+/// Shared subtrees are visited once per reference, since content-addressed children may be
+/// reachable by more than one path. `visit` should deduplicate by the visited hash if required.
+#[cfg(rocksdb_test_utils)]
+pub(crate) fn walk_stored_tree(
+    store: &impl ReadableKeyValueStore,
+    tree_hash: Hash,
+    mut visit: impl FnMut(Hash, usize),
+) -> Result<(), OperationalError> {
+    let empty = Tree::<Hash>::empty_hash();
+    let mut pending = vec![tree_hash];
+
+    while let Some(hash) = pending.pop() {
+        if hash == empty {
+            continue;
+        }
+
+        let bytes = store
+            .blob_get(hash)
+            .map_err(|error| OperationalError::CommitDataMissing {
+                root: hash,
+                source: Box::new(error),
+            })?;
+        let bytes = bytes.as_ref();
+
+        visit(hash, bytes.len());
+
+        let StoredNode { left, right, .. } = deserialise(bytes)?;
+        pending.push(left);
+        pending.push(right);
+    }
+
+    Ok(())
+}
+
 impl<TreeId: Loadable, DataId: DataLoadable> Loadable for Node<TreeId, DataId, Normal> {
     fn load(id: Hash, store: &impl ReadableKeyValueStore) -> Result<Self, OperationalError> {
         let StoredNode {
--- a/durable-storage/src/bin/gc_space.rs
+++ b/durable-storage/src/bin/gc_space.rs
@@ -1,0 +1,108 @@
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! Binary for the durable storage space accounting harness.
+//!
+//! Reports how much of a commit's Merkle node data is still reachable and how much is dead. See
+//! [`octez_riscv_durable_storage::gc_space`] for what is measured and why.
+//!
+//! Point `--repo-dir` at a volume with room to spare: the base state is recorded there and reused
+//! by later runs of the same shape, which matters because prepopulating dominates a large run. Each
+//! run resets the repository to that base state first, so reusing a directory does not fold earlier
+//! runs' commits into what is reported.
+
+use std::path::PathBuf;
+
+use anyhow::Result;
+use clap::Parser;
+use octez_riscv_durable_storage::gc_space::SpaceConfig;
+
+#[derive(Debug, Parser)]
+#[command(version, long_about = None)]
+struct Cli {
+    /// Number of databases in the registry.
+    #[arg(long, default_value_t = 1)]
+    databases: usize,
+
+    /// Keys written to each database before the measured commits begin.
+    #[arg(long, default_value_t = 100_000)]
+    keys: usize,
+
+    /// Length of each key, in bytes.
+    ///
+    /// The Etherlink trace has a median path of 33 bytes and a p90 of 99, with storage slots
+    /// (`/evm/world_state/eth_accounts/<address>/storage/<slot>`) around 130. Keys are stored inline
+    /// in each Merkle node, so this drives node size directly.
+    #[arg(long, default_value_t = 130)]
+    key_size: usize,
+
+    /// Length of an ordinary value, in bytes. The trace's median write is 32.
+    #[arg(long, default_value_t = 32)]
+    value_size: usize,
+
+    /// Fraction of keys holding a large value, reproducing the trace's contract-code tail.
+    #[arg(long, default_value_t = 0.0)]
+    large_value_fraction: f64,
+
+    /// Length of a large value, in bytes. The trace's largest write is 131220.
+    #[arg(long, default_value_t = 131_220)]
+    large_value_size: usize,
+
+    /// Number of commits to make and measure.
+    #[arg(long, default_value_t = 20)]
+    commits: usize,
+
+    /// Keys modified before each commit, spread across the registry's databases.
+    #[arg(long, default_value_t = 1_000)]
+    modified_keys: usize,
+
+    /// Measure every this many commits. Raise it at large scales, where scanning a column family
+    /// costs more than the commits do.
+    #[arg(long, default_value_t = 1)]
+    sample_every: usize,
+
+    /// Seed for choosing which keys each commit modifies.
+    #[arg(long, default_value_t = 0)]
+    seed: u64,
+
+    /// Where the repository lives. A fresh temporary directory when absent, which is removed on
+    /// exit along with the base state.
+    #[arg(long)]
+    repo_dir: Option<PathBuf>,
+}
+
+fn main() -> Result<()> {
+    let cli = Cli::parse();
+
+    if cli.databases == 0 {
+        anyhow::bail!("--databases must be at least 1");
+    }
+
+    if cli.keys == 0 {
+        anyhow::bail!("--keys must be at least 1");
+    }
+
+    if cli.sample_every == 0 {
+        anyhow::bail!("--sample-every must be at least 1");
+    }
+
+    if !(0.0..=1.0).contains(&cli.large_value_fraction) {
+        anyhow::bail!("--large-value-fraction must be between 0 and 1");
+    }
+
+    SpaceConfig {
+        databases: cli.databases,
+        keys_per_database: cli.keys,
+        key_size: cli.key_size,
+        value_size: cli.value_size,
+        large_value_fraction: cli.large_value_fraction,
+        large_value_size: cli.large_value_size,
+        commits: cli.commits,
+        modified_keys: cli.modified_keys,
+        sample_every: cli.sample_every,
+        seed: cli.seed,
+        repo_dir: cli.repo_dir,
+    }
+    .run()
+}
--- a/durable-storage/src/bin/gc_space.rs
+++ b/durable-storage/src/bin/gc_space.rs
@@ -70,6 +70,12 @@ struct Cli {
     /// exit along with the base state.
     #[arg(long)]
     repo_dir: Option<PathBuf>,
+
+    /// After the last commit, delete every commit not reachable from it and measure again.
+    ///
+    /// Shows what collecting at directory granularity reclaims, and what dead node data survives it.
+    #[arg(long)]
+    simulate_dir_gc: bool,
 }
 
 fn main() -> Result<()> {
@@ -103,6 +109,7 @@ fn main() -> Result<()> {
         sample_every: cli.sample_every,
         seed: cli.seed,
         repo_dir: cli.repo_dir,
+        simulate_dir_gc: cli.simulate_dir_gc,
     }
     .run()
 }
--- a/durable-storage/src/gc_space/measure.rs
+++ b/durable-storage/src/gc_space/measure.rs
@@ -176,10 +176,31 @@ fn measure_blob(
 }
 
 /// Occupancy of a directory tree, counting each inode's blocks once.
-fn disk_usage(root: &Path) -> Result<DiskUsage> {
+pub(super) fn disk_usage(root: &Path) -> Result<DiskUsage> {
     let mut usage = DiskUsage::default();
     let mut seen: HashSet<(u64, u64)> = HashSet::new();
     accumulate_disk_usage(root, &mut usage, &mut seen)?;
+
+    Ok(usage)
+}
+
+/// Occupancy of the committed history: the database commit directories and the registry manifests,
+/// leaving out the working databases.
+///
+/// Note this is what the history *references*, not what deleting it would free: commits hard-link
+/// the working database's files, so blocks counted here can be held by the working database too.
+/// Use [`SpaceConfig::simulate_dir_gc`] for what deletion actually reclaims.
+pub(super) fn commits_disk_usage(repo_path: &Path) -> Result<DiskUsage> {
+    let mut usage = DiskUsage::default();
+    let mut seen: HashSet<(u64, u64)> = HashSet::new();
+
+    // One `seen` set across both roots, so an inode linked from each is still counted once.
+    for root in [
+        repo_path.join("databases").join("commits"),
+        repo_path.join("registries"),
+    ] {
+        accumulate_disk_usage(&root, &mut usage, &mut seen)?;
+    }
 
     Ok(usage)
 }
--- a/durable-storage/src/gc_space/measure.rs
+++ b/durable-storage/src/gc_space/measure.rs
@@ -1,0 +1,308 @@
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! One measurement of a committed registry.
+//!
+//! [`measure`] produces a [`Sample`]. Everything else here is one of its figures: what the
+//! column families hold, what the directory tree occupies, what the history pins, and how the
+//! files are spread over LSM levels.
+
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::fs;
+use std::os::unix::fs::MetadataExt;
+use std::path::Path;
+use std::time::Duration;
+
+use anyhow::Context;
+use anyhow::Result;
+use octez_riscv_data::hash::Hash;
+
+use super::sample::BlobBreakdown;
+use super::sample::DiskUsage;
+use super::sample::FileSet;
+use super::sample::LevelSummary;
+use super::sample::PinnedBytes;
+use super::sample::Sample;
+use super::sample::Sharing;
+use super::scenario::Reg;
+use crate::avl::node::walk_stored_tree;
+use crate::commit::CommitId;
+use crate::persistence_layer::ReadOnlyPersistenceLayer;
+use crate::persistence_layer::measurement::CfTotals;
+use crate::persistence_layer::measurement::SstOwner;
+use crate::repo::DirectoryManager;
+use crate::storage::ReadOnlyKeyValueStore;
+
+/// Measure every database of a registry commit, plus the repository as a whole.
+pub(super) fn measure(
+    repo: &DirectoryManager,
+    repo_path: &Path,
+    registry_commit: &CommitId,
+    commit_index: usize,
+    commit_time: Duration,
+    previous: Option<&FileSet>,
+) -> Result<(Sample, FileSet)> {
+    let database_commits = Reg::database_commits(repo, registry_commit)
+        .context("reading the registry manifest being measured")?;
+
+    let mut blob = BlobBreakdown::default();
+    let mut value_stored_bytes = 0;
+    let mut files: FileSet = HashMap::new();
+
+    for (db_index, database_commit) in database_commits.iter().enumerate() {
+        let committed = ReadOnlyPersistenceLayer::checkout_read_only(repo, database_commit)
+            .with_context(|| {
+                format!(
+                    "opening database commit {} read-only",
+                    database_commit.hex_encode()
+                )
+            })?;
+
+        blob.add(&measure_blob(&committed, database_commit)?);
+
+        let values = committed
+            .value_totals()
+            .map_err(|error| anyhow::anyhow!("scanning the value column family: {error}"))?;
+        value_stored_bytes += values.stored_bytes();
+
+        for file in committed
+            .sst_files()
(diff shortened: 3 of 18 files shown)


durable-storage/src/persistence_layer.rs
```diff
@@ -181,6 +183,7 @@ fn add_creation_options(options: &mut rocksdb::Options) {
 fn rocksdb_default_options() -> rocksdb::Options {
     let mut options = rocksdb::Options::default();
     set_memtable_to_hash_link_list(&mut options);
+
```
emturner, line 223: remove this whitespace change

durable-storage/src/persistence_layer.rs
```diff
@@ -453,6 +456,180 @@ impl ReadOnlyKeyValueStore for ReadOnlyPersistenceLayer {
     }
 }
 
+/// One SST file of a committed database.
```
emturner: please add small doc for what an SST file actually is

durable-storage/src/persistence_layer.rs
```diff
@@ -453,6 +456,180 @@ impl ReadOnlyKeyValueStore for ReadOnlyPersistenceLayer {
...
+
+#[cfg(rocksdb_test_utils)]
+impl CfTotals {
+    /// Total bytes stored, keys and values together.
+    pub fn stored_bytes(&self) -> u64 {
+        self.key_bytes + self.value_bytes
+    }
+}
+
+/// Count the entries and bytes in `cf` by iterating all of it.
```
emturner:
```suggestion
/// Count the entries and bytes in the column family `cf` by iterating all of it.
```

durable-storage/src/persistence_layer.rs
```diff
@@ -453,6 +456,180 @@ impl ReadOnlyKeyValueStore for ReadOnlyPersistenceLayer {
...
+    /// hard-links the files live when it was taken, so a file present in two checkpoints of the
+    /// same database is one file on disk, and a file present in only the later one was written by
+    /// compaction or a flush in between.
+    pub fn sst_files(&self) -> Result<Vec<SstFile>, Box<dyn std::error::Error>> {
+        let files = self
+            .db_instance
+            .live_files()?
+            .into_iter()
+            .map(|file| SstFile {
+                owner: if file.column_family_name == BLOB_CF {
```
emturner: match on the family name maybe? (any other column name would be an error)

durable-storage/src/gc_space/mod.rs
```diff
@@ -0,0 +1,1141 @@
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
```
emturner: don't use `mod.rs` files - instead use the 

```
mod_name.rs
mod_name/<submodles.rs>
```

pattern

durable-storage/src/bin/gc_space.rs
```diff
@@ -69,6 +69,12 @@ struct Cli {
     /// exit along with the base state.
```
emturner, line 70: could any of this be hooked up into the long running tests?

durable-storage/src/persistence_layer.rs
```diff
@@ -453,6 +455,190 @@ impl ReadOnlyKeyValueStore for ReadOnlyPersistenceLayer {
...
+        BLOB_CF => Ok(SstOwner::Blob),
+        KV_CF => Ok(SstOwner::Value),
+        other => Err(format!("unexpected column family {other} in a database instance").into()),
+    }
+}
+
+/// Measurement of a committed database, for space accounting.
+///
+/// Only available to the harnesses: these are whole-column-family scans, far too expensive to sit
+/// on any path that matters.
```
emturner:
```suggestion
/// on any path where performance matters.
```

durable-storage/src/persistence_layer.rs
```diff
@@ -453,6 +455,190 @@ impl ReadOnlyKeyValueStore for ReadOnlyPersistenceLayer {
     }
 }
```
emturner, line 497: suggestion: make a `pub mod` submodule of the persistence layer for the purpose of all these helpers? 

(ie keep them separated from production code)

durable-storage/src/avl/node.rs
```diff
@@ -921,6 +921,51 @@ where
     }
 }
 
+/// Visit every node body reachable from the tree stored under `tree_hash`.
```
emturner, line 924: nit - please use `feat(nds)` instead of `dursto` in commits going forward

durable-storage/src/avl/node.rs
```diff
@@ -921,6 +921,51 @@ where
...
+/// stored representation can be walked directly. Empty trees are never stored, so a child hash
+/// equal to [`Tree::empty_hash`] terminates that branch.
+///
+/// Shared subtrees are visited once per reference, since content-addressed children may be
+/// reachable by more than one path. Deduplicate by the visited hash where that matters.
+///
+/// The walk is iterative over an explicit stack, so it is bounded by the tree's height rather
+/// than by the call stack.
+#[cfg(rocksdb_test_utils)]
+pub(crate) fn walk_stored_tree(
```
emturner, line 935: should this be in the tree module instead?

durable-storage/src/avl/node.rs
```diff
@@ -921,6 +921,51 @@ where
...
+/// once per body reached. Nothing is materialised in memory: a body is stored under the hash of
+/// the *tree* holding it, and its `left`/`right` fields are the hashes of the child trees, so the
+/// stored representation can be walked directly. Empty trees are never stored, so a child hash
+/// equal to [`Tree::empty_hash`] terminates that branch.
+///
+/// Shared subtrees are visited once per reference, since content-addressed children may be
+/// reachable by more than one path. Deduplicate by the visited hash where that matters.
+///
+/// The walk is iterative over an explicit stack, so it is bounded by the tree's height rather
+/// than by the call stack.
```
```suggestion
```

remove - implementation detail

durable-storage/src/avl/node.rs
```diff
@@ -921,6 +921,51 @@ where
...
+/// Visit every node body reachable from the tree stored under `tree_hash`.
+///
+/// `visit` receives the blob key each body is stored under and the length of the stored bytes,
+/// once per body reached. Nothing is materialised in memory: a body is stored under the hash of
+/// the *tree* holding it, and its `left`/`right` fields are the hashes of the child trees, so the
+/// stored representation can be walked directly. Empty trees are never stored, so a child hash
+/// equal to [`Tree::empty_hash`] terminates that branch.
+///
+/// Shared subtrees are visited once per reference, since content-addressed children may be
+/// reachable by more than one path. Deduplicate by the visited hash where that matters.
```
```suggestion
/// reachable by more than one path. `visit` should deduplicate by the visited hash if required.
```

durable-storage/src/bin/gc_space.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    if cli.sample_every == 0 {
+        anyhow::bail!("--sample-every must be at least 1");
+    }
+
+    if !(0.0..=1.0).contains(&cli.large_value_fraction) {
+        anyhow::bail!("--large-value-fraction must be between 0 and 1");
+    }
+
+    run(SpaceConfig {
```
emturner: could `run` be an `impl` on SpaceConfig instead? or is that not worth doing?

durable-storage/src/gc_space/report.rs
```diff
@@ -0,0 +1,178 @@
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! Turning samples into the printed output.
```
emturner: avoid gerunds in doc comments please! 

`Helpers for converting samples into printed output.`

durable-storage/src/gc_space/report.rs
```diff
@@ -0,0 +1,178 @@
...
+        "dead %",
+        "value MiB",
+        "disk MiB",
+        "commit ms",
+        "new MiB",
+        "shared %"
+    );
+}
+
+pub(super) fn report(sample: &Sample) {
```
emturner: feels like this would be worth possibly having as a `impl Display` for Sample instead

durable-storage/src/gc_space/report.rs
```diff
@@ -0,0 +1,178 @@
...
+    );
+    eprintln!(
+        "  repository occupies {:.1} MiB across {} commit directories ({:.1} MiB apparent, so \
+         hard links save {:.1} MiB)",
+        mib(last.disk.unique_bytes),
+        last.commit_dirs,
+        mib(last.disk.apparent_bytes),
+        mib(last
+            .disk
+            .apparent_bytes
```
emturner: this whole module could be a submodule of `scenario`. And then could be called by passing in/giving it to `eprintln`/`stderr()` later/higher up the satck

durable-storage/src/gc_space/measure.rs
```diff
@@ -0,0 +1,327 @@
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! Taking one measurement of a committed registry.
+//!
+//! [`measure`] produces a [`Sample`]. Everything else here is one of its figures: what the
+//! column families hold, what the directory tree occupies, what the history pins, and how the
+//! files are spread over LSM levels.
```
emturner: apply other feedback here too

durable-storage/src/gc_space/sample.rs
```diff
@@ -0,0 +1,214 @@
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! What one measurement of the repository records.
+//!
+//! [`Sample`] is the whole of it; the types it is built from each answer a different question
+//! about the same commit, so they are grouped here rather than beside the code that fills them
+//! in.
```
emturner, line 9: ditto

durable-storage/src/gc_space/scenario.rs
```diff
@@ -0,0 +1,377 @@
...
+            value_size: self.value_size,
+            large_value_permille: (self.large_value_fraction * 1000.0) as u32,
+            large_value_size: self.large_value_size,
+        }
+    }
+
+    /// Whether the key at `key_index` holds a large value.
+    ///
+    /// Decided from the index alone, so it is stable across commits and needs no bookkeeping.
+    fn is_large(&self, db_index: usize, key_index: usize) -> bool {
```
emturner:
```suggestion
fn is_large_key
```

durable-storage/src/gc_space/scenario.rs
```diff
@@ -0,0 +1,377 @@
...
+        if self.large_value_fraction <= 0.0 {
+            return false;
+        }
+
+        let permille = (self.large_value_fraction * 1000.0) as u64;
```
emturner:
```suggestion
        let per_mille = (self.large_value_fraction * 1000.0) as u64;
```

durable-storage/src/persistence_layer.rs
emturner: Applied.

durable-storage/src/persistence_layer.rs
emturner: Done — the helpers now live in `durable-storage/src/persistence_layer/measurement.rs`. The module declaration carries the `#[cfg(rocksdb_test_utils)]`, so the eight per-item attributes are gone. `merkle_store.rs` needed its `crate::persistence_layer::CfTotals` path updated to match.

durable-storage/src/avl/node.rs
emturner: Done, and applied retroactively while I was at it: all 16 `dursto`-scoped commits in this stack now say `nds`. Two notes — there are ~20 more `dursto` commits on unrelated branches that I have left alone, and one non-commit reference survives at `.github/workflows/nightly.yml:114` (a CI artifact name). Happy to rename that too if you want the term gone entirely.

durable-storage/src/avl/node.rs
emturner: I tried this and backed it out. `StoredNode` and its `left`/`right` fields are private to `node.rs`, so moving the walk needs them widened to `pub(super)` — which costs more than the `Tree::empty_hash` accessor it saves. The middle version (walk in `tree.rs`, a narrow `stored_child_trees` helper left behind in `node.rs`) comes out at the same item count, just relocated. The move also conflicts 7 commits of the GC stack on production AVL files.

So it stays in `node.rs` for now. If you would still rather have it under `Tree`, it is cheapest as its own commit above the stack. Your other two comments on this function are applied.

durable-storage/src/avl/node.rs
emturner: Removed.

durable-storage/src/avl/node.rs
emturner: Applied verbatim.

durable-storage/src/bin/gc_space.rs
emturner: Done — `run` is now `pub fn run(self)` inside `impl SpaceConfig`, which puts it in `scenario.rs` beside the config it consumes. That leaves `gc_space.rs` as module doc, `mod` declarations and re-exports, and the binary builds a `SpaceConfig` and calls `.run()`.

durable-storage/src/gc_space/report.rs
emturner: Applied verbatim, and the same treatment given to the other four module summaries: `One measurement of a committed registry.`, `Collection at directory granularity, and what it frees.`, `The shape of the scenario, and the state to measure it on.`, `What one measurement of the repository records.`

durable-storage/src/gc_space/report.rs
emturner: Left as `report` for now — tell me if you disagree and I will change it.

With the writer threaded through, the call site is already `report(&mut out, &sample)?`, so `Display` would not buy much at the point of use. What puts me off is that it would fix the table-row format onto `Sample` while `report_header` keeps the matching column widths separately; I would rather a row and its header stay next to each other in one module, so a column change is one edit.

durable-storage/src/gc_space/report.rs
emturner: The writer half is done. `report_header`, `report`, `summarise` and `report_prune` all take `out: &mut impl Write` and return `io::Result<()>`; `run` takes the handle once (`let mut out = io::stderr().lock();`) and lends it down. I threaded it through `base_registry`, `recorded_base` and `prepopulate` too, since those print progress, so there is no `eprintln!` left anywhere in the module. Output is unchanged and still on stderr — I checked stdout comes out empty.

I left the module where it is rather than nesting it under `scenario`: printing samples feels like a peer of measuring them rather than part of building the state. Say the word and I will move it.

durable-storage/src/gc_space/measure.rs
emturner: Done.

durable-storage/src/gc_space/sample.rs
emturner: Done.

durable-storage/src/gc_space/scenario.rs
emturner: Applied.

durable-storage/src/gc_space/scenario.rs
emturner: Applied, and the `BaseShape` field renamed to `large_value_per_mille` to match. No `serde(rename)` to preserve the old spelling on disk — nothing has landed, so there is no recorded base state to stay compatible with.

durable-storage/src/gc_space/prune.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Database commit directories removed.
+    pub databases_removed: u64,
+
+    /// Registry manifests removed.
+    pub registries_removed: u64,
+}
+
+impl PruneOutcome {
+    /// Blocks returned to the filesystem.
```
victor-dumitrescu: /// Bytes returned to the filesystem, counted as allocated blocks.

durable-storage/src/bin/gc_space.rs
```diff
@@ -0,0 +1,113 @@
+// SPDX-FileCopyrightText: 2026 Nomadic Labs <contact@nomadic-labs.com>
+//
+// SPDX-License-Identifier: MIT
+
+//! Binary for the durable storage space accounting harness.
+//!
+//! Reports how much of a commit's Merkle node data is still reachable and how much is dead. See
+//! [`octez_riscv_durable_storage::gc_space`] for what is measured and why.
+//!
+//! Point `--repo-dir` at a volume with room to spare: the base state is recorded there and reused
```
victor-dumitrescu, line 10: If the base state is used from `--repo-dir` directly, won't commits from previous runs affect the metrics for successive runs?

durable-storage/src/gc_space/sample.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub blob: BlobBreakdown,
+
+    /// Value column family totals, summed over the registry's databases.
+    pub value_entries: u64,
+
+    /// Bytes stored in the value column families, summed over the registry's databases.
+    pub value_stored_bytes: u64,
+
+    /// Size of the SST files backing the measured commits, as RocksDB reports it.
+    pub sst_bytes: u64,
```
victor-dumitrescu: Some of these fields are never read (maybe in a future PR?). I was surprised this wasn't a warning but it's because `Sample` is re-exported as `pub use` in `gc_space`, which I'm not sure it needs to?

durable-storage/src/gc_space/report.rs
```diff
@@ -0,0 +1,243 @@
...
+        out,
+        "  repository occupies {:.1} MiB across {} commit directories ({:.1} MiB apparent, so \
+         hard links save {:.1} MiB)",
+        mib(last.disk.unique_bytes),
+        last.commit_dirs,
+        mib(last.disk.apparent_bytes),
+        mib(last
+            .disk
+            .apparent_bytes
+            .saturating_sub(last.disk.unique_bytes)),
```
victor-dumitrescu: Does it make sense to subtract these if one is apparent bytes and one is block bytes?

durable-storage/src/gc_space/measure.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    Ok(pinned)
+}
+
+/// Count the database commit directories in the repository.
+fn count_commit_dirs(repo_path: &Path) -> Result<u64> {
+    let path = repo_path.join("databases").join("commits");
+
+    let entries = fs::read_dir(&path).with_context(|| format!("reading {}", path.display()))?;
```
victor-dumitrescu, line 314: Should filter on `is_dir`?

victor-dumitrescu (commented):
LGTM, and thanks for the long tests commit size report fix!

durable-storage/src/gc_space/prune.rs
emturner: Applied verbatim, thanks. It returns a byte count derived from `unique_bytes`, so "Blocks returned" had the unit wrong. [diff](https://github.com/tezos/riscv-pvm/compare/e29c34e3..f73d5715#diff-8ba571f2cd6c2b7bce9628f36e92209820d8895a9d6c9a4d65bb1c396ff974e7)

durable-storage/src/bin/gc_space.rs
emturner: Yes, they do — good catch, this was real.

Confirmed by running the same shape twice against one `--repo-dir`: sample 0's `disk` was 1.4 MiB on the first run and **6.5 MiB** on the second, and the summary's `history pins … growing` collapsed from 1.3 MiB per commit to **0.0**, because the pins were already on disk.

Fixed with `reset_to_base`, called from `run` just after `base_registry`: it removes every database commit directory and registry manifest the base state does not reach. That also cleans up after a shape-mismatched recorded base, which until now was ignored but left on disk.

Chasing this turned up a second, quieter version of the same problem: the run that *built* the base measured the database it had just prepopulated, while every later run measured a checkout of that commit — and the two have different LSM shapes. First run against a fresh directory reported 6.6 MiB occupancy and 81.4% Merkle share; every run after it, 4.7 MiB and 73.1%. `base_registry` now checks the base out even on the run that builds it, so all runs take the same path. Three consecutive runs on a fresh directory are byte-identical apart from commit timings. [diff](https://github.com/tezos/riscv-pvm/compare/e29c34e3..f73d5715)

durable-storage/src/gc_space/report.rs
emturner: It doesn't, no — thanks for catching it. `apparent_bytes` is `Σ len()` counted per link; `unique_bytes` is `Σ blocks * 512` counted per inode. Subtracting them folds together two corrections that point in opposite directions: sharing takes bytes away, and rounding each file up to whole blocks puts them back.

On an SST-dominated repo the error is modest — 6.21 MiB reported against 6.33 actual on a small run — but it has a floor. On a 6-epoch registry long-test repo the apparent sum was 1.47 MiB against 1.57 MiB block-deduped, so `saturating_sub` would have printed "hard links save 0.0 MiB" while sharing was really saving around half a megabyte.

`DiskUsage` now also carries `linked_bytes`, which is the same blocks as `unique_bytes` counted once per link, and `shared_bytes()` is the difference between the two — sharing on its own, with no rounding folded in. The struct doc made the same claim and is corrected too. [diff](https://github.com/tezos/riscv-pvm/compare/e29c34e3..f73d5715)

durable-storage/src/gc_space/measure.rs
emturner: Added. It's hardening rather than a live fix — `databases/commits` only ever receives `DirectoryManager::database_commit_dir`, so every entry there is a commit directory today — but it now matches `accumulate_disk_usage`, which already skipped non-directories.

The same assumption was load-bearing more sharply in `prune_unreachable`: there a non-directory entry fails the name check and then gets `remove_dir_all`'d, which errors outright rather than just miscounting. That loop got the filter too. [diff](https://github.com/tezos/riscv-pvm/compare/e29c34e3..f73d5715)

durable-storage/src/gc_space/sample.rs
emturner: Right on both counts, and the answer to "maybe in a future PR?" is yes — so I moved them rather than annotating them.

Those five are consumed by the commit that adds the JSON output, so they now live there instead of here. Worth it beyond tidiness: `disk_commits` was costing a full directory walk over `databases/commits` and `registries` on every sample, computed and thrown away.

The re-exports are gone too — all seven were dead crate-wide (`BlobBreakdown`, `DiskUsage`, `LevelSummary`, `PinnedBytes`, `Sharing`, `PruneOutcome` and `Sample` itself). Nothing outside needs to name them: a caller reaches them through `Sample`'s fields, and Rust is fine with public fields whose types aren't publicly nameable. `BlobBreakdown::dead_entries` was genuinely unused and is deleted. [diff](https://github.com/tezos/riscv-pvm/compare/e29c34e3..03481091)

victor-dumitrescu (approved):

PR #1180, opened 2026-08-24
fix(data): bound NDS proof verification by proof size, not claimed length

emturner:
Closes TZX-208

# What

- `descend_tree` no longer descends into absent or blinded subtrees.
- Verify-mode folding of `Bytes` and `Vector` skips subtrees that hold no state, via `PrunedSeqAsTree` and `PartialHashFold::skip_unchanged_subtree`.
- Removes the unused `Many` helper.

# Why

`sequence_as_tree_from_proof` recovers a sequence length from a leaf in the proof — a `Bytes` value length, or a registry `Vector` size — and uses it as the number of leaves to visit. Nothing bounds what that figure may claim, and hashing is no help: a proof carrying an astronomical length can still hash correctly, so there is no later check that would catch it. A small proof could therefore ask the verifier to do a large and effectively unbounded amount of work.

The same length is used again on the way out, since the verify-mode `PartialHashFold` implementations build an `IndexableSeqAsTree` over it and walk every node. Bounding parsing alone would not have been enough.

`Many` had no callers, and was the only consumer of `descend_tree` that assumed one visit per index, so pruning would have broken it.

# How

No limit is placed on the claimed length, so proofs describing legitimately large values keep working. Both halves are bounded structurally instead.

Parsing stops at an absent or blinded node, making work proportional to the nodes actually present — bounded by the proof's byte length, since a present node costs at least a tag byte per level. This changes neither which proofs are accepted nor the hashes derived from them: both deserialisers already yield absent branches beneath such a node without consuming input, and `OwnedProofTree::node_from_children` discards their children outright.

Hashing folds a subtree with nothing beneath it straight to the hash the reference proof already carries. That shortcut is narrow by design, since acceptance is pinned by the committed state hash: it applies only to the two cases provably identical to a full descent — no reference proof, and a blinded subtree. A present node is still descended, so one carrying fewer children than the state expects keeps reporting `InvalidProof`. An empty sequence is left alone too, folding to the hash of an empty node.

For anyone splitting these commits: the second is a prerequisite for the third having any effect on `Vector`, whose fold predicate asks whether any item in a range is defined — and before pruning, every item below the claimed length was.

# Manually Testing

```
make all
```

# Regressions

Adds `data/proptest-regressions/components/vector/tests.txt`: the zero-length cases proptest found while this was written, kept so they stay covered.

# Tasks for the Author

### DIFF
--- a/data/src/components/bytes/tests.rs
+++ b/data/src/components/bytes/tests.rs
@@ -12,13 +12,17 @@ use tezos_smart_rollup_constants::core::MAX_FILE_CHUNK_SIZE;
 
 use super::test_utils::BytesOp;
 use super::test_utils::MAX_PROOF_OFFSETS;
+use crate::codec::Bincode;
+use crate::codec::LeafEncode;
 use crate::components::bytes::Bytes;
+use crate::components::bytes::NODE_ARITY;
 use crate::components::bytes::PAGE_SIZE;
 use crate::components::bytes::test_utils::BytesMutOp;
 use crate::components::bytes::test_utils::MAX_PROOF_LENGTH;
 use crate::components::bytes::test_utils::NDS_BYTES_LENGTH;
 use crate::foldable::Foldable;
 use crate::foldable::Unfoldable;
+use crate::foldable::seq_tree::tree_depth;
 use crate::foldable::tests::TestFolder;
 use crate::hash::Hash;
 use crate::hash::PartialHash;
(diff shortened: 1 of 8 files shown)


data/src/merkle_proof/tests.rs
```diff
@@ -564,3 +565,62 @@ fn round_trip_descend_tree_indexable_seq_as_tree() {
...
     });
 }
+
+/// A proof that blinds a subtree must not cost the verifier anything proportional to the number of
+/// leaves that subtree claims to stand in for.
+///
+/// The leaf count handed to [`descend_tree`] comes from a length leaf carried by the proof itself,
+/// and is only reconciled with the committed state hash once parsing has finished. A proof is
+/// therefore free to claim a sequence of `usize::MAX` items. If the descent walked the implied tree
+/// rather than the nodes actually present, this test would not terminate.
```
emturner: I mean, it would terminate - just in a very, very, very long time?

Secondly - worth noting that reconciling against a state hash doesn't help at all: because an attacker can force a proof that indeed contains such astronomically large values that hash correctly, by playing a refutation game against itself - so this comment is slightly misleading?

data/src/merkle_proof.rs
```diff
@@ -276,6 +276,24 @@ where
 {
     let mut ctx = proof.into_node()?;
 
+    // An absent or blinded node carries no information about its descendants, so descending into
+    // one costs time proportional to `total_leaves` - a figure the proof itself supplies - rather
+    // than to the size of the proof. Stop here instead.
+    //
+    // This is a correctness requirement rather than an optimisation. `total_leaves` derives from
+    // a length leaf that the proof is free to set, and it is only ever checked against the
+    // committed state hash *after* parsing, so a constant-size proof claiming a sequence of 2^64
```
emturner: again, the committed state hash has no bearing here

data/src/components/bytes/tests.rs
```diff
@@ -633,3 +635,56 @@ fn test_bytes_largest_valid_proof_nds() {
         assert!(proof.len() <= MAX_PROOF_LENGTH);
     });
 }
+
+/// Folding a verify-mode value must cost what the proof contains, not what it claims to describe.
+///
+/// The length reaching the fold is recovered from the proof and is only reconciled with the
+/// committed state hash afterwards, so a proof is free to assert a value of any size. Nothing has
```
emturner: ditto - committed state hash does not help

data/src/components/vector/tests.rs
```diff
@@ -571,3 +574,33 @@ fn fold_unfold_bytes() {
         prop_assert_eq!(vector, unfolded);
     });
 }
+
+/// A registry proof may claim any number of databases: the count is recovered from the proof and
+/// only reconciled with the committed state hash once verification is over. Both parsing such a
```
emturner: ditto

data/src/foldable/seq_tree.rs
```diff
@@ -122,6 +122,136 @@ where
     }
 }
 
+/// Driver for folding a sequence as a tree in verify mode, skipping subtrees that hold no state.
+///
+/// [`IndexableSeqAsTree`] visits every leaf, which is what you want when the length is known to be
+/// sound. In verify mode it is not: the length is recovered from the proof and only reconciled with
+/// the committed state hash afterwards, so folding leaf by leaf lets a proof choose how much work
```
emturner: ditto

data/src/foldable/seq_tree.rs
```diff
@@ -122,6 +122,136 @@ where
...
+            .arity
+            .checked_pow(self.current_depth)
+            .unwrap_or(usize::MAX);
+        let end = self.current_start.saturating_add(width).min(self.total_len);
+
+        self.current_start..end
+    }
+}
+
+impl<'a, C, L, G, P> Foldable<PartialHashFold<C>> for PrunedSeqAsTree<'a, L, G, P>
```
emturner: please give more explicit Generic arg names

data/src/hash.rs
```diff
@@ -398,6 +398,38 @@ impl<C: LeafCodec> PartialHashFold<C> {
...
+    ///
+    /// Nothing read or written beneath a subtree means every leaf under it would fold to
+    /// [`PartialHash::Previous`], and the node would collapse back to whatever the reference proof
+    /// already says about it. Answering straight away avoids a descent whose cost is set by the
+    /// sequence length rather than by the proof - and in verify mode that length comes from the
+    /// proof itself.
+    ///
+    /// Returns `Err(self)` where the shortcut would not match a full descent, leaving the caller
+    /// to descend as it otherwise would.
+    pub fn skip_unchanged_subtree(self) -> Result<PartialHash, Self> {
```
emturner, line 411: does this work in the case that the verifiier computation sets a value at an index in the vector, but in a large range of undefined? (such that partial hash folding requires additional blinded nodes?)

data/src/merkle_proof/tests.rs
emturner: Both fair, thank you.

On termination: you're right, it's finite. Reworded to say it would visit on the order of 2^64 nodes - finite, but not in any time anyone will wait for.

On the state hash: agreed, and this is the more important correction. I'd written it as though hashing were a late backstop, when a proof carrying an astronomically large length can hash perfectly well - so there is no later check to fall back on and the bound has to be structural. I'd made that same claim in five places; all reworded, along with the commit message and the PR description.

data/src/merkle_proof.rs
emturner: Agreed, reworded. It now says nothing bounds what the proof may claim here and that hashing is no help, since such a proof can still hash correctly - rather than implying a late check that would eventually catch it.

data/src/components/bytes/tests.rs
emturner: Reworded here too - same correction as above.

data/src/components/vector/tests.rs
emturner: Reworded here too.

data/src/foldable/seq_tree.rs
emturner: Reworded here too. This one now says the length is recovered from the proof and nothing bounds what it may claim, so folding leaf by leaf would let a proof decide how much work the verifier does.

data/src/foldable/seq_tree.rs
emturner: Done: `L` -> `Item`, `G` -> `Generator`, `P` -> `HasState`, and `C` -> `Codec` on the impl. `L` and `G` were carried over unthinkingly from `IndexableSeqAsTree`.

data/src/hash.rs
emturner: Good question - it holds, but checking it properly exposed a gap in my testing rather than in the code.

Writing to index `i` makes `is_any_defined` true for every range containing `i`, so the whole path from root to `i` is descended. Only its siblings can take the shortcut, and only when their reference proof is blinded or absent:

- blinded: returns `Present(h)`, which is what a full descent recovers from the node's own `node_hash` anyway;
- absent: returns `Previous`, and the parent then sees `Present` (the path) alongside `Previous` (the sibling), so it reports `InvalidProof` - the correct rejection when the proof lacked a sibling hash needed to re-hash the path.

That is the reason the shortcut declines for a present node instead of answering from its root hash.

I had asserted this rather than tested it, so I've added a differential proptest, `pruned_fold_agrees_with_full_descent`. It folds the same state through `PrunedSeqAsTree` and `IndexableSeqAsTree` over random arities, lengths, defined-index sets and proof shapes, and asserts the two agree.

Worth reporting how that went, because my first attempt at it was worthless. I checked it by mutating `skip_unchanged_subtree` to answer for a present node from its root hash - and the test still passed. For a present node carrying a full complement of children the shortcut genuinely does agree: each child folds to `Present(child.root_hash())`, and re-hashing those reproduces the parent's root hash. The divergence only appears when the proof node carries *fewer* children than the state expects, which my generator never built. With truncation added to the generator the mutation is caught, and the real implementation passes.

data/src/components/bytes/tests.rs
```diff
@@ -633,3 +637,136 @@ fn test_bytes_largest_valid_proof_nds() {
...
+    );
+}
+
+/// A write into a subtree the proof only blinded must be rejected, not folded away.
+///
+/// This is the case the `has_state` predicate exists to catch. The proof carries a single blind for
+/// the whole page sequence, so there is no way to re-hash it once a page underneath has changed:
+/// the written page hashes to something, its neighbours defer to a proof that says nothing about
+/// them, and mixing the two is what `InvalidProof` reports. Were the predicate to under-report,
+/// the fold would answer from the blind and quietly accept a state the proof cannot support.
```
emturner: this should succeed, though - if the entire region that was blinded has now been written to!

e.g. a blinded page - can be fully overwritten and still hash correctly

data/src/hash.rs
```diff
@@ -398,6 +398,38 @@ impl<C: LeafCodec> PartialHashFold<C> {
...
 
+    /// Fold a subtree that holds no state of its own, without descending into it.
+    ///
+    /// Nothing read or written beneath a subtree means every leaf under it would fold to
+    /// [`PartialHash::Previous`], and the node would collapse back to whatever the reference proof
+    /// already says about it. Answering straight away avoids a descent whose cost is set by the
+    /// sequence length rather than by the proof - and in verify mode that length comes from the
+    /// proof itself.
+    ///
+    /// Returns `Err(self)` where the shortcut would not match a full descent, leaving the caller
```
emturner, line 409: is this exploitable in any way?

data/src/components/bytes/tests.rs
emturner: You're right, and the implementation already behaves that way - it was my test that was wrong. Measured on a 4-page blinded region:

- overwrite all 4 pages: `Present(b10f8cfd...)`
- overwrite 1 of 4: `InvalidProof`

So blinding is not what makes it fail; the write being *partial* is, because the untouched pages still need previous contents that nothing can supply. My test asserted the rejection and its name and doc generalised that to any write under a blind, which is simply not true.

Fixed by splitting the two:

- renamed to `verify_fold_rejects_a_partial_write_into_a_blinded_subtree`, with the doc now saying explicitly that full overwrite succeeds and pointing at the test below;
- added `verify_fold_accepts_a_full_overwrite_of_a_blinded_subtree`.

The new one asserts against `Hash::from_foldable(&Bytes::<Normal>::from(contents))` rather than just checking it isn't rejected - so it pins that a fully overwritten blinded region re-hashes to exactly what a concrete value of those contents hashes to, and the blind is never consulted.

data/src/hash.rs
emturner: Not as far as I can establish, and here is the reasoning so you can check it rather than take my word.

The shortcut only fires when `has_state` reports nothing defined under the subtree, and only for two proof shapes - absent, and blinded. Both are provably what a full descent produces, and `pruned_fold_agrees_with_full_descent` checks that differentially over random arities, lengths, defined-index sets and proof shapes, including proofs no honest prover would emit (blinded over written state, and nodes carrying fewer children than the state expects).

The exploitable direction would be `has_state` **under**-reporting: calling a populated range empty would fold real state away and let a forged state hash to the committed value. Two things bound that:

- `PartialVec::is_any_defined` is exact - a binary search to the first entry that could overlap, then a scan with early exit.
- All the range arithmetic saturates *outward*. `covered_range` uses `checked_pow(..).unwrap_or(usize::MAX)` and `saturating_add`, so on overflow the range only widens; the `Bytes` predicate maps page indices to byte offsets with `saturating_mul` and clamps the end to `length`. A wider range can only make the predicate report *more* state, which means descending more. Over-reporting costs time; it is never wrong.

Worth noting the one extreme does exercise that: at `length` near `usize::MAX` the page count reaches `2^54` and `pages.end * PAGE_SIZE` saturates - whereupon `.min(length)` clamps it to `length`, which is the correct end. The start cannot overflow, since `pages.start < 2^54`.

One limitation I would rather flag than have you find. The differential proptest builds proofs whose depth matches the state's, so it does not exercise `DepthAdjustedSeqAsTree` re-scoping the reference proof after a value has been resized. That path is untouched by this PR and is covered end to end by the existing `Bytes` and `Vector` round-trip and cross-mode proptests, whose operation sets include resizes - but it is the corner where I would most want a second opinion, since it is the one place the proof and the state tree are deliberately not the same shape.

emturner: It does overlap, and I had only reasoned about it rather than tested it - so I have tested it now. Two additions.

**A differential proptest across the adjustment**, `pruned_fold_agrees_with_full_descent_across_a_depth_adjustment`. Lengths are drawn at `arity^k` and one either side, the reference proof is built at the *original* length, and both drivers are folded inside the same `DepthAdjustedSeqAsTree`. That covers narrowing (shrink) and dummy-layer padding (grow), against random defined-index sets, blinded subtrees and truncated nodes. Confirmed it discriminates: widening the shortcut to answer for a present node from its root hash makes it fail.

**A concrete large-scale case**, `verify_fold_stays_bounded_across_a_depth_adjusting_resize`. A value claiming a tebibyte grows by one byte, taking the page count from `2^30` to `2^30 + 1` and the depth from 30 to 31, so the proof gets a dummy layer. It asserts the depth really does change (otherwise the test would be vacuous), then that the fold hashes. Only the page holding the new byte is descended into; the rest is skipped from the blind. Completes instantly rather than walking a billion pages.

On why this holds, now that it is checked: the shortcut's equivalence to a full descent is a property of two things only - the shape of the reference proof at the builder's current position, and whether anything is defined in the range. The depth adjustment rewrites the proof *before* handing it over, so it can change which shape the shortcut sees, but not whether that shape is answered correctly. And it cannot touch `has_state` at all, since the predicate consults the state's `PartialVec` and never the proof.

Bounded work, separately: both adjustment loops run `|original_depth - current_depth|` times, which is at most the tree depth - about 32 for a `usize` length at arity 4 - so re-scoping cannot reintroduce work proportional to a claimed length.

data/src/foldable/seq_tree.rs
```diff
@@ -3,8 +3,7 @@
 // SPDX-License-Identifier: MIT
 
 //! Helpers for sequences that need to fold and unfold like trees
-
```
victor-dumitrescu, line 6: Nit, empty line deleted here.

data/proptest-regressions/foldable/seq_tree.txt
victor-dumitrescu, line 1: Are these really worth including? I don't think they cover anything more than the added tests themselves?

victor-dumitrescu (approved):

data/src/foldable/seq_tree.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/f9072624d10dac812923aac83044f69caa05534a..eba97cbe463a41150e415b71bfdc2da180b89822)

data/proptest-regressions/foldable/seq_tree.txt
emturner: no that's fair - removed in [diff](https://github.com/tezos/riscv-pvm/compare/f9072624d10dac812923aac83044f69caa05534a..eba97cbe463a41150e415b71bfdc2da180b89822)

PR #1186, opened 2026-08-25
feat(nds): add 'lifecycle' benchmarks

emturner:
Relates to TZX-209

# What

Add benchmarks covering the full lifecycle of how NDS is intended to be used

Six commits: the prepopulation harness (with checking a registry out), then copy, move, clear, commit and hash — thirteen benchmarks in the `Registry lifecycle` group.

# Why

Allows us to understand the performance implications, on a more 'real-world' basis.

While it may look tangential to the GC work - a lot of the work that goes on there _does_ have performance implications too. Therefore, it's worth having these benchmarks available to run, to demonstrate any speedup/slowdown that may occur as a result of implementing GC

# Manually Testing

```
make all
```

The defaults — 4 databases of 10M keys, 60s per measurement — are the reference-machine scale, and will fill a laptop's disk. Override them for a smoke run; this covers all thirteen benchmarks in about 80 seconds and ~260MB of scratch:

```
REGISTRY_DATABASE_COUNT=2 PREPOPULATED_NODE_KEYS_COUNT=20000 COMMIT_MODIFIED_KEYS_COUNT=500 \
  cargo bench -p octez-riscv-durable-storage --bench database_lifecycle -- \
  --sample-size 10 --measurement-time 1
```

The prepopulated registry lives in a temp dir, so set `TMPDIR` to a path on a real volume if `/tmp` is a tmpfs. To keep it between runs instead — worth it as the scale grows — set `LIFECYCLE_REGISTRY_DIR` to a path on a large filesystem; a directory built with a different config is rebuilt rather than misused.

# Tasks for the Author

### DIFF
--- a/durable-storage/benches/database_lifecycle.rs
+++ b/durable-storage/benches/database_lifecycle.rs
@@ -1,0 +1,294 @@
+// SPDX-FileCopyrightText: 2026 TriliTech <contact@trili.tech>
+//
+// SPDX-License-Identifier: MIT
+
+//! Focused lifecycle benchmarks for the durable-storage [`Registry`].
+//!
+//! The new durable storage is a [`Registry`] owning a vector of
+//! [`Database`]s, so these benchmarks operate on the registry — not a bare
+//! database — to capture the registry-level costs the PVM actually pays, in
+//! particular the registry hash taken on every commit and re-derived on every
+//! checkout.
+//!
+//! Unlike the ERC-20 workload benchmarks, these replay no store-access trace.
+//! Each isolates one whole-database/registry lifecycle operation on a large,
+//! prepopulated registry (`REGISTRY_DATABASE_COUNT` databases of
+//! `PREPOPULATED_NODE_KEYS_COUNT` keys each):
+//!
+//! - **Checkout** — `Registry::checkout`: restore the whole registry from a
+//!   committed snapshot.
+//!
+//! All scenarios use [`BatchSize::PerIteration`], so only one large registry is
+//! live at a time and the teardown drop of the value *returned* from the routine
+//! is untimed.
+//!
+//! The sample size and measurement time default to small, stable values;
+//! `--sample-size` and `--measurement-time` override them.
+//!
+//! Populating a large registry is slow. Set `LIFECYCLE_REGISTRY_DIR` to a path
+//! on a **large filesystem** to persist the prepopulated registry there and
+//! reuse it across runs. The manifest saved alongside it records the config it
+//! was built with, so a mismatched directory is rebuilt rather than misused.
+//! Unset → ephemeral temp dir, populate every run.
+//!
+//! [`Database`]: octez_riscv_durable_storage::database::Database
+
+mod random;
+
+use std::path::Path;
+use std::path::PathBuf;
+use std::time::Duration;
+
+use bytes::Bytes;
+use criterion::BatchSize;
+use criterion::Criterion;
+use criterion::criterion_group;
+use criterion::criterion_main;
+use octez_riscv_data::mode::Normal;
+use octez_riscv_durable_storage::commit::CommitId;
+use octez_riscv_durable_storage::database::DirectoryManager;
+use octez_riscv_durable_storage::key::Key;
+use octez_riscv_durable_storage::persistence_layer::PersistenceLayer;
+use octez_riscv_durable_storage::registry::Registry;
+use octez_riscv_test_utils::TestableTmpdir;
+use rand::rng;
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::random::generate_keys;
+use crate::random::generate_random_bytes_in_range;
+
+/// A Normal-mode registry backed by the persistent (RocksDB) key-value store.
+type Reg = Registry<PersistenceLayer, Normal>;
+
+/// Number of keys to prepopulate *each* database with, controlling how "large"
+/// the databases under test are. Override with the `PREPOPULATED_NODE_KEYS_COUNT`
+/// environment variable (shared with the ERC-20 database benchmarks).
+const PREPOPULATED_NODE_KEYS_COUNT: usize = 10_000_000;
+
+/// Number of databases held in the registry. Override with the
+/// `REGISTRY_DATABASE_COUNT` environment variable.
+const REGISTRY_DATABASE_COUNT: usize = 4;
+
+/// Number of keys retained per database for the modification scenarios. Keeping
+/// *every* key (up to `PREPOPULATED_NODE_KEYS_COUNT` per database, each up to
+/// [`KEY_MAX_SIZE`] bytes) would hold gigabytes of keys in memory for the whole
+/// run; a bounded random sample is enough to pick keys to modify.
+///
+/// [`KEY_MAX_SIZE`]: octez_riscv_durable_storage::key::KEY_MAX_SIZE
+const KEY_SAMPLE_PER_DB: usize = 100_000;
+
+/// File name (within `LIFECYCLE_REGISTRY_DIR`) of the manifest describing a
+/// prepopulated registry that can be reused across runs.
+const PREPOPULATED_MANIFEST_FILE: &str = "prepopulated_lifecycle.json";
+
+fn env_usize(name: &str, default: usize) -> usize {
+    std::env::var(name)
+        .ok()
+        .map(|v| {
+            v.parse()
+                .unwrap_or_else(|_| panic!("{name} must be a number"))
+        })
+        .unwrap_or(default)
+}
+
+/// Manifest persisted alongside a prepopulated registry so a later run can skip
+/// the (slow) populate phase.
+#[derive(Serialize, Deserialize)]
+struct PrepopulatedManifest {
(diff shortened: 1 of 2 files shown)


durable-storage/benches/database_lifecycle.rs
```diff
@@ -0,0 +1,464 @@
+// SPDX-FileCopyrightText: 2026 TriliTech <contact@trili.tech>
```
emturner, line 1: Never record claude-session in a commit. 

the title `restore` in the commit title is misleading, as these benches never landed before

durable-storage/benches/database_lifecycle.rs
```diff
@@ -0,0 +1,464 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+//! Focused lifecycle benchmarks for the durable-storage [`Registry`].
+//!
+//! The PVM's durable storage is a [`Registry`] owning a vector of
+//! [`Database`](octez_riscv_durable_storage::database::Database)s, so these
```
emturner: when writing doc comments, make links work by moving the link to the bottom of the doc comment:

```
//! ... [`Database`] ...
//!
...
//!
//! [`Database`]: octez_riscv_durable_storage::database::Database
```

This makes it easier to read the doc comments inline in the code

durable-storage/benches/database_lifecycle.rs
```diff
@@ -0,0 +1,464 @@
...
+//! benchmarks operate on the registry — not a bare database — to capture the
+//! registry-level costs the PVM actually pays, in particular the registry hash
+//! (`Hash::from_foldable`) taken on every commit and re-derived-and-verified on
```
emturner: no need to mention the mechanism of hashing here

durable-storage/benches/database_lifecycle.rs
```diff
@@ -0,0 +1,464 @@
...
+        .map(|v| {
+            v.parse()
+                .unwrap_or_else(|_| panic!("{name} must be a number"))
+        })
+        .unwrap_or(default)
+}
+
+/// File name (within `LIFECYCLE_REGISTRY_DIR`) of the manifest describing a
+/// prepopulated registry that can be reused across runs.
+const PREPOPULATED_MANIFEST_FILE: &str = "prepopulated_lifecycle.bin";
```
emturner: please (in general) group constants together towards the top of the file

durable-storage/benches/database_lifecycle.rs
```diff
@@ -0,0 +1,464 @@
...
+
+/// Manifest persisted alongside a prepopulated registry so a later run can skip
+/// the (slow) populate phase: the snapshot commit to check out, the config it
+/// was built with (to detect a stale directory), and the per-database key sample
+/// the modification scenarios pick from.
```
emturner:
```suggestion
/// the (slow) populate phase.
```

this is a good example of being tautologous/overly verbose: you're directly recounting the fields!

Please use serialise/json here rather than bincode - there's no need for this to not be human readable

durable-storage/benches/database_lifecycle.rs
```diff
@@ -0,0 +1,464 @@
...
+///
+/// Returns the (checked-out, lazily-loaded) registry, the keys of each database
+/// — so the commit benchmark can modify existing entries — and the commit id of
+/// the populated registry, used by the checkout benchmark.
+///
+/// The registry is built one database at a time: after populating each database
+/// we commit and check the registry back out, which flushes the just-built
+/// database to disk and reloads every database lazily. This keeps only a single
+/// fully-materialised database tree in memory at a time — populating all of them
+/// at once exhausts memory (four 10M-key trees do not fit in ~16 GiB).
```
emturner: may exhaust memory perhaps? no need to mention a specific amount of RAM imho

durable-storage/benches/database_lifecycle.rs
```diff
@@ -0,0 +1,464 @@
...
+            BatchSize::PerIteration,
+        )
+    });
+
+    group.finish();
+}
+
+criterion_group! {
+    name = benches;
+    // Set the defaults on the `Criterion` instance rather than on the group, so
```
emturner: ditto - reduce verbosesness everywhere (take this as a general comment)

durable-storage/benches/database_lifecycle.rs
```diff
@@ -264,6 +265,106 @@ fn populate_registry(repo: &DirectoryManager) -> (Reg, Vec<Vec<Key>>, CommitId)
...
+        let value = Bytes::from(generate_random_bytes_in_range(rng, 1..32));
+        registry
+            .database_mut(db_index)
+            .expect("The database index should be valid")
+            .set(key, value)
+            .expect("Modifying a key should succeed");
+    }
+}
+
+fn clone_modified_spread(
```
emturner, line 276: you could maybe also break out the first commit into multiple ones? Per operation type?

durable-storage/benches/database_lifecycle.rs
emturner: Both gone. That commit is now `feat(bench): add a registry lifecycle benchmark harness` — no `Claude-Session` trailer, and the paragraph about the abandoned revision it was recovered from has gone with it. `Co-Authored-By` stays.

durable-storage/benches/database_lifecycle.rs
emturner: Done — `[Database]` and `[KEY_MAX_SIZE]` are both link definitions at the end of their doc comments now.

durable-storage/benches/database_lifecycle.rs
emturner: Dropped. The module docs now just say the registry hash is taken on every commit and re-derived on every checkout.

durable-storage/benches/database_lifecycle.rs
emturner: Moved `PREPOPULATED_MANIFEST_FILE` up beside the others, so the constants are one block above `env_usize`.

durable-storage/benches/database_lifecycle.rs
emturner: Taken verbatim.

The manifest is JSON now, `prepopulated_lifecycle.json`. It was cheaper than expected: `Key` and `CommitId` already have serde impls under `cfg(test_utils)` that render as hex strings, and `serde`/`serde_json` are already dev-dependencies of `durable-storage`, so it came down to swapping the derives for `serde_json::{from_slice, to_vec}`. Smoke-checked all three paths at 2x500 keys: populate-and-save, reuse ("Reusing prepopulated registry…"), and rebuild on a config mismatch.

durable-storage/benches/database_lifecycle.rs
emturner: Now just "populating all of them at once may exhaust memory".

durable-storage/benches/database_lifecycle.rs
emturner: Taken as a general pass rather than a local fix: the module docs are down from 66 lines to 54, every scenario comment is shorter, and this one is two lines.

It also turned up `modify_n` wearing `clone_modified_spread`'s doc comment on top of its own — each has the one that describes it now.

durable-storage/benches/database_lifecycle.rs
emturner: Done, though split by weight rather than only per operation: each of copy/move/clear was ~30 lines of the 464-line commit, so per-operation alone would have left ~380 + 30 + 30. The bulk was the prepopulation harness.

The branch is now six commits — harness + checkout (295 lines), then copy (+83), move (+33), clear (+32), commit (+164), hash (+40). The harness is paired with checkout so the copy/move/clear commits stay purely per-operation; checkout needs only the repo and the snapshot commit, no modification helpers.

One knock-on worth flagging: the numbered scenario list became a named one (`// Scenario 4c:` → `// The *second* commit.`), because commits introducing scenarios in a different order would otherwise renumber the list every time. Same reason "Settling the commit's write application" generalised to "Settling the writes" — the copy commit needs it before any commit scenario exists.

durable-storage/benches/database_lifecycle.rs
```diff
@@ -0,0 +1,644 @@
+// SPDX-FileCopyrightText: 2026 TriliTech <contact@trili.tech>
+//
+// SPDX-License-Identifier: MIT
+
+//! Focused lifecycle benchmarks for the durable-storage [`Registry`].
+//!
+//! The PVM's durable storage is a [`Registry`] owning a vector of
```
victor-dumitrescu:
```suggestion
//! The durable storage is a [`Registry`] owning a vector of
```

😅

victor-dumitrescu (approved):

durable-storage/benches/database_lifecycle.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/74e8a95f1217654c1fca858c44bd984243ed1b31..2dbebd878637c5ed58ef39ba5577b42d5b8db9b5)

thanks! 😆

PR #1187, opened 2026-08-25
feat(nds): reduce balance factor to `i8` from `i64`

emturner:
- Part of TZX-196

# What

Balance factor type updated from `i64` to `i8`.

# Why

The only allowed values for encoding/decoding 'in theory' (not yet enforced) should be `-1..=1`. Values of `-2 & 2` are temporarily encountered during insert/delete + rebalancing. 

We will reduce this to an enum - but first, it is clear that we can reduce drastically the amount we encode in the proof per balance factor (reducing by 7 bytes per balance factor).

## Impact on database sizes

Seven bytes per balance factor is seven bytes per node on the search path, which buys exactly one extra level of AVL depth for every operation. Because minimal AVL node counts are Fibonacci, one level is roughly 1.6x the number of keys:

```
┌─────────────────────────────────────────┬────────────────────────┬────────────────────────┐
│                operation                │         before         │         after          │
├─────────────────────────────────────────┼────────────────────────┼────────────────────────┤
│ Delete (existing)                       │ d=52 → 139,583,862,443 │ d=53 → 225,851,433,715 │
├─────────────────────────────────────────┼────────────────────────┼────────────────────────┤
│ Set / Write (fresh insert)              │ d=41 → 701,408,731     │ d=42 → 1,134,903,168   │
├─────────────────────────────────────────┼────────────────────────┼────────────────────────┤
│ Read / Write (max chunk, 64 MiB value)  │ d=34 → 24,157,815      │ d=35 → 39,088,167      │
├─────────────────────────────────────────┼────────────────────────┼────────────────────────┤
│ Set (replacing a 64 MiB value)          │ d=39 → 267,914,294     │ d=40 → 433,494,435     │
├─────────────────────────────────────────┼────────────────────────┼────────────────────────┤
│ ValueLength / Exists / Delete (missing) │ d=48 → 20,365,011,072  │ d=49 → 32,951,280,097  │
├─────────────────────────────────────────┼────────────────────────┼────────────────────────┤
│ Hash                                    │ unbounded              │ unbounded              │
├─────────────────────────────────────────┼────────────────────────┼────────────────────────┤
│ all operations                          │ d=34 → 24,157,815      │ d=35 → 39,088,167      │
└─────────────────────────────────────────┴────────────────────────┴────────────────────────┘
```

# Manually Testing

```
make all
```

# Regressions

Regenerated as proofs as smaller and hashes updated.

# Tasks for the Author


durable-storage/src/test_helpers/counterfeit.rs
```diff
@@ -139,13 +139,21 @@ mod tests {
...
+    use crate::test_helpers::proof_size::LEN_LEAF;
+    use crate::test_helpers::proof_size::TAG_BYTES;
+    use crate::test_helpers::proof_size::TREE_WRAP;
+    use crate::test_helpers::proof_size::avl_path_node;
+    use crate::test_helpers::proof_size::key_leaf;
+
+    /// Every key in these tests is [`COUNTERFEIT_KEY_LEN`] bytes, to make the size of the
+    /// generated proofs predictable.
     ///
     /// This will become unnecessary once the merkle layer switches to using hashed keys.
```
jobjo: As we don't plan on this - let's remove. 
```suggestion
```

jobjo (approved):

durable-storage/src/test_helpers/counterfeit.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/3bda7b75f437372f1bfc5bbbd975041c47b541fa..2bf977c829a25c85ea4302bf11a759b635648194)

PR #1188, opened 2026-08-25
feat(ci): report gc metrics

emturner:
- Part of TZX-209

# What

Track the actual gc metrics in CI (like we do for RISC-V performance). (Though not in datadog - just as PR comments). Tracking in datadog in future is something we _could_ consider

# Why

Ensure that GC actually does the job - improving sharing and/or reducing dead-bytes

# Manually Testing

```
make all
```

# Tasks for the Author

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -14,9 +14,11 @@ on:
 
   workflow_dispatch:
 
-# Only one job may reserve the reference machine at a time
+# Only one job may reserve the reference machine at a time. The group is a literal rather than
+# `${{ github.workflow }}` so that it is shared with the Durable Storage Benchmark workflow, which
+# reserves the same machine: a per-workflow group would let the two run concurrently.
 concurrency:
-  group: ${{ github.workflow }}
+  group: reference-machine
   cancel-in-progress: false
 
 jobs:
--- a/.github/workflows/durable-space-metrics.yml
+++ b/.github/workflows/durable-space-metrics.yml
@@ -1,0 +1,130 @@
+name: Durable Storage Space Metrics
+
+on:
+  # Trigger this workflow on PRs. A pull request touching nothing but prose cannot move the
+  # figures, and this is not a required check, so skipping it there is safe.
+  pull_request:
+    paths-ignore:
+      - "**.md"
+      - "docs/**"
+
+  # Trigger this workflow for Merge Queue items
+  merge_group:
+
+  # Trigger this workflow on pushes to these branches
+  push:
+    branches: [main]
+
+# Its own workflow rather than a job in Continuous Integration, so that the run list for this
+# file holds metrics runs and nothing else: the comment below links the default branch's figures,
+# and that link is only useful if every run behind it measured something.
+concurrency:
+  # Cancel previous workflow runs for the same ref (e.g. same PR or branch)
+  group: ${{ github.workflow }}-${{ github.ref }}
+
+  # Cancel only if it is not the main branch
+  cancel-in-progress: ${{ github.ref_name != 'main' }}
+
+env:
+  # Make Cargo output pretty
+  CARGO_TERM_COLOR: "always"
+
+  # Enable Sccache
+  RUSTC_WRAPPER: "sccache"
+
+  # Make sure Sccache automatically uses the GitHub Actions cache
+  SCCACHE_GHA_ENABLED: "true"
+
+  # Only write Sccache on the main branch
+  SCCACHE_GHA_RW_MODE: ${{ github.ref_name == 'main' && 'READ_WRITE' || 'READ_ONLY' }}
+
+jobs:
+  # Reports how much of the storage is garbage, what retaining a commit costs, and how much of a
+  # commit's bytes are shared with the previous one. Small scale on a pull request, so it catches a
+  # regression in those before it merges; the reference-scale run tracks the trend.
+  space-metrics:
+    name: Durable Storage Space Metrics
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      pull-requests: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Set up build environment
+        uses: ./.github/actions/setup
+        with:
+          # This job saves no cache, so there is nothing for a prefetch to fill: the Continuous
+          # Integration matrix populates the same keys from the same environment, and the Actions
+          # cache is shared across workflows.
+          save-nix-cache: "false"
+          prefetch-cargo: "false"
+
+      - name: Run the space metrics
+        run: |
+          # A pull request gets the short scale, to keep the check quick. The merge queue and the
+          # default branch get the longer one, which runs enough commits for the retention and
+          # sharing figures to say something the short run cannot.
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            target=durable/space-metrics
+          else
+            target=durable/space-metrics-long
+          fi
+
+          # The harness reports on stderr and 'bench-metrics --markdown' writes the table on
+          # stdout, so keeping the two apart separates the headline figures from their detail
+          # without parsing either. Cargo says nothing, to keep its progress out of the report,
+          # and make says nothing about the directory it recurses into, which lands on stdout and
+          # would otherwise be quoted into the comment above the table.
+          export CARGO_TERM_QUIET=true
+
+          make --no-print-directory "$target" MARKDOWN=--markdown \
+            > metrics.md 2> report.txt || status=$?
+
+          # The report is the log for this job, whether the run succeeded or not.
+          cat report.txt
+
+          exit "${status:-0}"
+
+      - name: Compose the report
+        run: |
+          {
+            echo "### Durable storage space metrics"
+            echo
+            cat metrics.md
+            echo
+            echo '```'
+            cat report.txt
+            echo '```'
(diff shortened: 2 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -105,3 +105,85 @@ jobs:
...
+            echo "</details>"
+            echo
+            echo "These figures are for catching a regression before it merges. The tracked series"
+            echo "is the reference-scale run on the [default branch](${{ github.server_url }}/${{ github.repository }}/actions/workflows/durable-bench.yml?query=branch%3Amain)."
+          } > pr-comment.md
+
+      - name: Comment metrics on PR
+        # A pull request from a fork gets a read-only token, so there is no comment to post; the
+        # metrics still run and the check still passes.
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
```
emturner: a pull request from a fork should not run anything on the benchmark machine at all! This is because this _could_ accidentally expose the secrets used to connect

.github/workflows/durable-bench.yml
```diff
@@ -0,0 +1,48 @@
+name: Durable Storage Benchmark
+
+on:
+  # Run every day at ~03:24 UTC, after the nightly tests and away from the start of the hour
+  schedule:
+    - cron: "24 3 * * *"
```
emturner: probably worth removing the nightly line

.github/workflows/durable-bench.yml
```diff
@@ -0,0 +1,48 @@
...
+  cancel-in-progress: false
+
+jobs:
+  benchmark:
+    name: Durable Storage Benchmark
+    runs-on: ubuntu-latest
+
+    # The reference machine does the work; this job only drives it. Prepopulating four
+    # million-key databases dominates the run.
+    timeout-minutes: 180
```
emturner: will this correctly re-use the database(s) if prepopulated? Is there a fallback to re-populating in the case that some format has changed on disk and so the run fails?

.github/workflows/ci.yml
emturner: Agreed on the principle, and it already holds here — but the comment I wrote invited exactly this doubt, so I have reworded it.

This job runs on `ubuntu-latest` and never touches the reference machine: it runs `make durable/space-metrics` on the runner itself and needs no secret at all. The fork condition on this step is only about *posting the comment* — a fork PR's `GITHUB_TOKEN` is read-only, so without the guard the step would fail and redden an otherwise green check.

On the wider point: the reference machine is reached only by `bench.yml` and `durable-bench.yml`, and both trigger on `pull_request` rather than `pull_request_target`. GitHub withholds secrets from fork PRs on `pull_request`, so `BENCH_SSH_KEY` is empty there and the ssh-agent step fails closed — annoying (a red Benchmark check on every fork PR) but not a disclosure. If you want that annoyance gone too, `bench.yml`'s job `if:` could take `&& github.event.pull_request.head.repo.full_name == github.repository`; say the word and I will add it, though it is outside this PR.

The SQUASH ME also now deletes `durable-bench.yml` outright (see the reply on the nightly comment), so after squashing this PR creates nothing that touches the reference machine.

.github/workflows/durable-bench.yml
emturner: Gone — and the SQUASH ME goes further than the one line, which is worth checking you want.

Once the space metrics run in `ci.yml` at a longer scale on the merge queue and the default branch, this workflow has nothing left to do at this commit: the space figures are byte counts and machine-independent (the script says so itself), and they only lived on the reference machine because the scale wanted gigabytes of scratch — which the setup action's cleanup already frees on a shared runner. Measured, the long scale is 88s and 2 GB.

So the SQUASH ME deletes `durable-bench.yml` and `ci-durable-bench.sh` here, and a second one recreates them on `ci: report registry lifecycle latencies` driving the latencies alone, which are the part that genuinely needs the machine. Net effect: nightly gone, one run per commit to main, and this PR no longer creates anything that touches the reference machine.

If you would rather keep this commit's shape and only drop the cron, say so and I will cut it back to that one line — it just leaves a reference-machine run doing work a shared runner now does better.

.github/workflows/durable-bench.yml
emturner: Checked both harnesses. Short answer: **in CI nothing is reused**, so there is nothing stale to fall back from — but by hand the two harnesses behave differently, and one of them is missing exactly the fallback you describe.

**In CI, neither reuses anything.** `gc_space` reuses a base state only when `--repo-dir` is passed; without it the CLI makes "a fresh temporary directory … removed on exit along with the base state", and no make target passes it. The lifecycle bench reuses only when `LIFECYCLE_REGISTRY_DIR` is set; unset, it uses an ephemeral `TestableTmpdir` and populates every run. The script sets neither, so both prepopulate from scratch each time. That is also why the job is slow, and why the timeout comment reads the way it does.

**By hand, the lifecycle bench does what you are asking for.** `try_load_prepopulated` returns `Option`, and every failure path is `.ok()?` — an unreadable manifest, a manifest that no longer deserialises (i.e. a format change), a config mismatch, or a failing `Reg::checkout` all yield `None`, and the caller repopulates. A stale on-disk format therefore costs a repopulate, not a failed run.

**By hand, `gc_space` does not.** `recorded_base` compares only the *shape*: a mismatch is reported and rebuilt, and `reset_to_base` clears the commits the stale base left (covered by `space_harness_restricted_rebuilds_mismatched_base`). But if the shape matches and the format has moved, `base_registry` does `Reg::checkout(...).context("checking out the base state")?` and the run aborts. A base-state file that no longer parses aborts even earlier, in `recorded_base` itself. No fallback.

Worth fixing, and small — treat a failed load or checkout as "no recorded base", loudly, the way the shape mismatch already does. That is a change to `gc_space`, which is already on main, so it wants its own commit rather than a squash into this PR. Want me to write it?

victor-dumitrescu:
```
 commit    blob MiB    live MiB    dead MiB   dead %   value MiB    disk MiB  commit ms  commits MiB     new MiB  shared %  rewrote MiB
      0        24.8        24.8         0.0     0.0%        15.4        42.6          0         42.5           -         -            -
      1        26.3        24.8         1.5     5.6%        15.4        44.3        124         44.2         1.7     96.1%          0.0
      2        27.7        24.8         2.9    10.6%        15.4        47.1        121         47.0         2.7     94.2%          0.0
      3        29.2        24.8         4.4    15.2%        15.4        97.8        140         50.6         3.6     92.9%          0.0
      4        30.7        24.8         5.9    19.2%        15.4       102.2        147        102.1        51.5      0.0%         50.5
      5        32.2        24.8         7.4    22.9%        15.4       107.3        164        107.2         5.1     91.0%          0.0
      6        33.7        24.8         8.9    26.3%        15.4       113.0        182        112.9         5.7     90.8%          0.0
      7        35.1        24.8        10.3    29.4%        15.4       156.6        195        119.7         6.7     90.1%          0.5
      8        36.6        24.8        11.8    32.3%        15.4       163.4        192        163.2        43.5     28.4%         51.2
      9        38.1        24.8        13.3    34.9%        15.4       170.7        206        170.5         7.3     89.3%          0.0
     10        39.6        24.8        14.8    37.4%        15.4       195.1        226        178.3         7.8     89.8%          0.0
```
  
The big swings in the new bytes at certain commits are due to compaction I imagine? Won't this make the average hard to compare? Or ar compactions expected to happen in exactly the same way given it's always the same scenario?

victor-dumitrescu (commented):
LGTM otherwise!

emturner:
> ```
>  commit    blob MiB    live MiB    dead MiB   dead %   value MiB    disk MiB  commit ms  commits MiB     new MiB  shared %  rewrote MiB
>       0        24.8        24.8         0.0     0.0%        15.4        42.6          0         42.5           -         -            -
>       1        26.3        24.8         1.5     5.6%        15.4        44.3        124         44.2         1.7     96.1%          0.0
>       2        27.7        24.8         2.9    10.6%        15.4        47.1        121         47.0         2.7     94.2%          0.0
>       3        29.2        24.8         4.4    15.2%        15.4        97.8        140         50.6         3.6     92.9%          0.0
>       4        30.7        24.8         5.9    19.2%        15.4       102.2        147        102.1        51.5      0.0%         50.5
>       5        32.2        24.8         7.4    22.9%        15.4       107.3        164        107.2         5.1     91.0%          0.0
>       6        33.7        24.8         8.9    26.3%        15.4       113.0        182        112.9         5.7     90.8%          0.0
>       7        35.1        24.8        10.3    29.4%        15.4       156.6        195        119.7         6.7     90.1%          0.5
>       8        36.6        24.8        11.8    32.3%        15.4       163.4        192        163.2        43.5     28.4%         51.2
>       9        38.1        24.8        13.3    34.9%        15.4       170.7        206        170.5         7.3     89.3%          0.0
>      10        39.6        24.8        14.8    37.4%        15.4       195.1        226        178.3         7.8     89.8%          0.0
> ```
>   
> The big swings in the new bytes at certain commits are due to compaction I imagine? Won't this make the average hard to compare? Or ar compactions expected to happen in exactly the same way given it's always the same scenario?

Yes, exactly! It always happens at exactly the same point - it's why there's two different averages too.

One of the things we do in the GC stack is to reduce the frequency of compaction to promote sharing, but ofc it does come with a degree of of performance implications (one of the various reasons to have the lifecycle benchmarks)

victor-dumitrescu (approved):

PR #1190, opened 2026-08-28
feat(nds): update registry arity to 2

emturner:
- Closes TZX-211

# What

Ensure registry is included in database proof size limit, and update arity of vector to 2.

# Why

Bundling these three changes together keeps the net effect on the number of keys/values we support in proofs (as reported by `proof_size.rs`) visible in one place. As the bounds are now tighter and more accurate, the overall effect is to raise confidence in the figures.

## Impact on database sizes

A database proof never ships on its own, so it is now measured against `DATABASE_MAX_PROOF_SIZE` (15288 B) rather than the full 16 KiB `MAXIMUM_PROOF_SIZE`: the difference is the registry envelope that wraps it (1096 B for the deepest registry `registry_resize` admits).

Against that reservation, the binding operation gets 677 B cheaper to model — 644 B from capping `value_open` at `MAX_PROOF_LENGTH`, and 33 B from no longer charging the terminal search-path node's blinded data child when the value is opened. The arity change takes a further 546 B off the envelope itself. That leaves 419 B unpaid, so the binding operation loses one level of AVL depth; because minimal AVL node counts are Fibonacci, a level is roughly 1.6x the number of keys:

```
┌─────────────────────────────────────────┬────────────────────────┬───────────────────────┐
│                operation                │         before         │         after         │
├─────────────────────────────────────────┼────────────────────────┼───────────────────────┤
│ Delete (existing)                       │ d=53 → 225,851,433,715 │ d=50 → 53,316,291,171 │
├─────────────────────────────────────────┼────────────────────────┼───────────────────────┤
│ Set / Write (fresh insert)              │ d=48 → 20,365,011,072  │ d=45 → 4,807,526,974  │
├─────────────────────────────────────────┼────────────────────────┼───────────────────────┤
│ Read / Write (max chunk, 64 MiB value)  │   d=35 → 39,088,167    │   d=34 → 24,157,815   │
├─────────────────────────────────────────┼────────────────────────┼───────────────────────┤
│ Set (replacing a 64 MiB value)          │   d=40 → 433,494,435   │   d=36 → 63,245,984   │
├─────────────────────────────────────────┼────────────────────────┼───────────────────────┤
│ ValueLength / Exists / Delete (missing) │ d=49 → 32,951,280,097  │ d=46 → 7,778,742,047  │
├─────────────────────────────────────────┼────────────────────────┼───────────────────────┤
│ Hash                                    │       unbounded        │       unbounded       │
├─────────────────────────────────────────┼────────────────────────┼───────────────────────┤
│ all operations                          │   d=35 → 39,088,167    │   d=34 → 24,157,815   │
└─────────────────────────────────────────┴────────────────────────┴───────────────────────┘
```

The reduction is a correction rather than a loss of capability. The `before` column was measured two ways that both overstated it: against a ceiling that ignored the registry wrapper the proof has to fit inside, and with a `value_open` that charged every touched page a full root-to-leaf branch of blinded siblings. `MAX_PROOF_LENGTH` (4097 B) is the worst case actually derivable for a three-page read or write over a 64 MiB leaf, sharing included.

For the record, d=34 now clears its ceiling by 5 bytes, and reaching d=35 would need another 324 B — out of the 1096 B envelope, or out of the 329 B per search-path level, which is dominated by the 255-byte key leaf.

# How

Tighten the bounds, by taking into account the true maximum size of `Byte` proofs, include the registry proof in the database proof-size limit, and reduce the arity of the registry.

# Manually Testing

```
make all
```

# Regressions

Updated due to arity change of `Vector`.

# Tasks for the Author


durable-storage/src/test_helpers/proof_size.rs
```diff
@@ -181,10 +185,48 @@ fn value_byte_ranges(op: &DatabaseOperation, value_len: usize) -> [Range<usize>;
...
+/// [`MAX_REGISTRY_LEN`] in that base.
+const MAX_REGISTRY_DEPTH: usize = MAX_REGISTRY_LEN
+    .next_power_of_two() // required as `ilog2` rounds down
+    .ilog2()
+    .div_ceil(VECTOR_NODE_ARITY.ilog2()) as usize;
+
+/// Ceiling on the serialised size of a whole registry proof: the shape that
+/// ships for any operation on a database held in a registry slot. Proofs above
+/// it are reported by [`assert_registry_proof_size`].
 const MAXIMUM_PROOF_SIZE: usize = 16 * 1024;
```
victor-dumitrescu: Maybe worth renaming this to `REGISTRY_MAX_PROOF_SIZE` for consistency with the newly introduced `DATABASE_MAX_PROOF_SIZE`.

durable-storage/tests/integration_test.proptest-regressions
victor-dumitrescu, line 1: What is this regression for?

data/src/components/vector.rs
```diff
@@ -873,7 +873,11 @@ impl<T: CloneState> CloneState for VerifyImpl<T> {
 }
 
 /// Arity of internal nodes in the Merkle tree
-pub const NODE_ARITY: usize = 4;
+///
+/// A binary tree keeps a single-item access cheap to prove: a proof of one item
+/// blinds one sibling per layer, so it carries `log2(len)` hashes rather than
+/// the `3 * log4(len)` a four-way tree needs.
```
victor-dumitrescu: Might be a bit confusing to include the comparison to a arity 4 tree, which now no longer exists?

durable-storage/src/test_helpers/proof_size.rs
```diff
@@ -181,10 +191,48 @@ fn value_byte_ranges(op: &DatabaseOperation, value_len: usize) -> [Range<usize>;
...
-/// Ceiling on serialised proof sizes. Proofs above it are reported by
-/// [`assert_proof_size`].
+/// One layer of a registry slot path: the node tag plus the blinded siblings of
+/// the child the path descends into.
+const REGISTRY_LAYER: usize = TAG_BYTES + (VECTOR_NODE_ARITY - 1) * BLIND_LEAF;
+
+/// The envelope a registry proof wraps around the proofs of the db's it
+/// touches. This includes the registry's `u64` length leaf, and, per touched
+/// slot, one [`REGISTRY_LAYER`] per layer of the tree plus the db's blinded
+/// content.
```
victor-dumitrescu: The blinded content isn't included, right?

durable-storage/src/test_helpers/proof_size.rs
```diff
@@ -181,10 +191,48 @@ fn value_byte_ranges(op: &DatabaseOperation, value_len: usize) -> [Range<usize>;
...
+/// [`MAX_REGISTRY_LEN`] in that base.
+const MAX_REGISTRY_DEPTH: usize = MAX_REGISTRY_LEN
+    .next_power_of_two() // required as `ilog2` rounds down
+    .ilog2()
+    .div_ceil(VECTOR_NODE_ARITY.ilog2()) as usize;
+
+/// Ceiling on the serialised size of a whole registry proof: the shape that
+/// ships for any operation on a database held in a registry slot. Proofs above
+/// it are reported by [`assert_registry_proof_size`].
 const MAXIMUM_PROOF_SIZE: usize = 16 * 1024;
```
victor-dumitrescu: Maybe worth renaming this to `REGISTRY_MAX_PROOF_SIZE` for consistency with the newly introduced `DATABASE_MAX_PROOF_SIZE`.

durable-storage/tests/integration_test.proptest-regressions
emturner: good catch, removed [diff](https://github.com/tezos/riscv-pvm/compare/9074bb13f3ac15445c11ef9efd8f084e678ab2c6..e4fa32d26573219394397784231e4360c4fbbb9f)

data/src/components/vector.rs
emturner: deleted [diff](https://github.com/tezos/riscv-pvm/compare/9074bb13f3ac15445c11ef9efd8f084e678ab2c6..e4fa32d26573219394397784231e4360c4fbbb9f)

durable-storage/src/test_helpers/proof_size.rs
emturner: ah sorry, this was misleading - it's more about accounting for the `final_hash`

I've reworked to make the distinction between the database/registry proof sizes, and the final hash in [diff](https://github.com/tezos/riscv-pvm/compare/9074bb13f3ac15445c11ef9efd8f084e678ab2c6..e4fa32d26573219394397784231e4360c4fbbb9f)

durable-storage/src/test_helpers/proof_size.rs
emturner: constants reworked in [diff](https://github.com/tezos/riscv-pvm/compare/9074bb13f3ac15445c11ef9efd8f084e678ab2c6..e4fa32d26573219394397784231e4360c4fbbb9f)

victor-dumitrescu (approved):

PR #1197, opened 2026-09-02
fix(nds): ensure database copies are truly independent

emturner:
- Closes TZX-212
- Part of RV-998

# What

Fix the 'copy pollution' bug recorded as TZX-212: a read on a database copy bled a value back into
the original, moving the root the original commits to.

# Why

Source of non-determinism. A layer's committed root stopped describing its own contents, so two
nodes that had seen the same writes could disagree on their root.

# How

`MerkleLayer::checkout` resolves the root eagerly, and `clone_with` clones the already-initialised
identifiers inside it - so a source and its copy share tree structure while each holds a store of
its own. Resolving one of those shared identifiers filled it from whichever store the resolving
layer held, and the identifier then folded by hashing those bytes, discarding the hash the node was
committed with. Nothing is mutated on that path, so copy-on-write never engaged: a read was enough.

1. We now load data lazily, only as required. As used by `Registry/Database` (normal mode) - this is only for `set` and `write`
2. We additionally enforce that for `read` on the  merkle layer directly, that the value loaded is consistent (which prevents incorrect values being loaded into the tree)

# Manually Testing

```
```

To see the bug rather than take its absence on trust, check out the first commit and run the
layer-level test with `--ignored`; it fails there and passes at the tip:

```
cargo test -p octez-riscv-durable-storage test_reading_a_copy_does_not_change_the_source_root -- --ignored
```

It's also worth noting that

```
AVL cold set/delete (persisted, lazy resolution)
                        time:   [23.739 ms 23.868 ms 24.041 ms]
                        change: [−37.244% −36.294% −35.300%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Improves a lot here (versus main) - while we do now do an extra hash on `read`, for `set/delete` this doesn't apply - and we load far fewer data values into the nodes

# Tasks for the Author

### DIFF
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -1089,7 +1089,7 @@ trait DataLoadable: Sized {
 
 impl DataLoadable for Bytes<Normal> {
     fn load(
-        _id: Hash,
+        id: Hash,
         key: &Key,
         store: &impl ReadableKeyValueStore,
     ) -> Result<Self, OperationalError> {
@@ -1103,6 +1103,19 @@ impl DataLoadable for Bytes<Normal> {
                     })?;
             Bytes::from(bytes.as_ref())
         };
+
+        // This value folds by hashing its own bytes, so an unchecked load would let whatever the
+        // store returned become the node's hash. Check it against the hash the node commits to
+        // for the same reason the lazy identifier does.
+        let found = Hash::from_foldable(&data);
+
+        if found != id {
+            return Err(OperationalError::CommitValueMismatch {
+                key: key.clone(),
+                expected: id,
+                found,
+            });
+        }
 
         Ok(data)
     }
--- a/durable-storage/src/avl/node.rs
+++ b/durable-storage/src/avl/node.rs
@@ -30,6 +30,7 @@ use perfect_derive::perfect_derive;
 
 use super::key::NodeKey;
 use super::key::NodeKeyMode;
+use super::resolver::DataResolver;
 use super::resolver::LazyDataId;
 use super::resolver::LazyTreeId;
 use super::resolver::ProveNode;
@@ -37,7 +38,6 @@ use super::resolver::TreeResolver;
 use super::tree::Tree;
 use crate::avl::resolver::AvlResolver;
 use crate::errors::Error;
-use crate::errors::InvalidArgumentError;
 use crate::errors::OperationalError;
 use crate::key::Key;
 use crate::storage::Loadable;
@@ -82,15 +82,24 @@ pub struct Node<TreeId, DataId, M: Mode> {
 
 impl Node<LazyTreeId, LazyDataId, Normal> {
     /// Converts the [`Node`] to prove mode.
-    pub fn into_proof(self) -> ProveNode {
-        Node {
+    ///
+    /// A node's value is not loaded when the node is, but a prove-mode value carries the bytes
+    /// it is proved against and their length, so it is resolved here - through the resolver
+    /// that owns the store this node came from - rather than assumed already present.
+    pub fn into_proof(
+        self,
+        resolver: &impl DataResolver<LazyDataId, Normal>,
+    ) -> Result<ProveNode, OperationalError> {
+        resolver.resolve_bytes(&self.data, self.key.as_ref())?;
+
+        Ok(Node {
             balance_factor: self.balance_factor.into_proof(),
             key: self.key.into_proof(),
             data: self.data.into_proof(),
             left: self.left.into_proof(),
             right: self.right.into_proof(),
             hash: self.hash.clone(),
-        }
+        })
     }
 
     #[cfg(test)]
@@ -880,7 +889,7 @@ impl<TreeId, DataId, M: AtomMode + NodeKeyMode> Node<TreeId, DataId, M> {
 
 impl<TreeId: Storable, DataId> Storable for Node<TreeId, DataId, Normal>
 where
-    DataId: Foldable<HashFold> + Borrow<[u8]>,
+    DataId: Foldable<HashFold> + DataStorable,
 {
     fn store(
         &self,
@@ -909,9 +918,7 @@ where
 
         // Are we in charge of writing the value data to the KV store?
         if options.node_data() {
-            let key: &[u8] = self.key.as_ref().as_ref();
-            let value: &[u8] = self.data.borrow();
-            store.set(key, value)?;
+            self.data.store_data(self.key.as_ref(), store)?;
         }
 
         self.left.store(store, options)?;
@@ -1022,6 +1029,53 @@ where
     }
 }
 
+/// Helper trait for writing a node's value to a [`WriteableKeyValueStore`].
+///
+/// A node loaded from a store carries only the hash its value was committed with, and there is
+/// nothing to write for such a node: the store this layer resolves from already holds that value,
+/// and it is the store being written. `Borrow<[u8]>` cannot express "nothing to write" - it would
+/// have to produce bytes that are not there - so storing goes through this trait instead.
+///
+/// This makes `MerkleLayer::commit` writing the layer's own persistence a precondition rather than
+/// a coincidence: committing a checked-out layer into some other store would leave the values it
+/// never loaded behind.
+trait DataStorable {
+    /// Write the value for `key` to `store`.
+    fn store_data(
+        &self,
+        key: &Key,
+        store: &impl WriteableKeyValueStore,
+    ) -> Result<(), OperationalError>;
+}
+
+impl DataStorable for Bytes<Normal> {
+    fn store_data(
+        &self,
+        key: &Key,
+        store: &impl WriteableKeyValueStore,
+    ) -> Result<(), OperationalError> {
+        let value: &[u8] = self.borrow();
+        store.set(key, value)
+    }
+}
+
+impl DataStorable for LazyDataId {
+    fn store_data(
+        &self,
+        key: &Key,
+        store: &impl WriteableKeyValueStore,
+    ) -> Result<(), OperationalError> {
+        // A value still held as a hash was never mutated through this layer, so the store this
+        // layer resolves from already holds it - and that is the store being written. Loading it
+        // would read bytes back only to write the same ones again.
+        let Some(value) = self.loaded() else {
+            return Ok(());
+        };
+
+        store.set(key, Borrow::<[u8]>::borrow(value))
+    }
+}
+
 /// Helper trait for node data that can be reconstructed
 /// from a [`ReadableKeyValueStore`] by content hash and key.
 trait DataLoadable: Sized {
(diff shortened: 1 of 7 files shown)


durable-storage/src/merkle_layer.rs
```diff
@@ -1055,6 +1055,74 @@ mod tests {
         );
```
emturner, line 1055: remove claude-session in all commit messages

durable-storage/src/avl/node.rs
```diff
@@ -1022,6 +1029,50 @@ where
...
+impl StorableData for LazyDataId {
+    fn store_data(
+        &self,
+        key: &Key,
+        store: &impl WriteableKeyValueStore,
+    ) -> Result<(), OperationalError> {
+        // A layer commits to the store it was checked out from, so a value still held as a hash
+        // is read back from the store that holds it. A store that cannot answer is a missing
+        // value to report rather than a state to assume away.
+        let value: &[u8] = self.try_get(key, store)?.borrow();
```
emturner: no, we should not attempt to load it at this point, if it's not loaded already

durable-storage/src/avl/resolver.rs
```diff
@@ -393,11 +393,21 @@ impl LazyDataId {
...
+    /// hash is dropped only once the value is in hand, so a load that fails leaves the identifier
+    /// still folding by the hash the node commits to.
     fn try_get_mut(
         &mut self,
         key: &Key,
         store: &impl ReadableKeyValueStore,
     ) -> Result<&mut Bytes<Normal>, OperationalError> {
+        if self.0.inner.get().is_none() {
+            let bytes = self.read(key, store)?;
+            let _ = self.0.inner.set(bytes);
```
emturner: correct me if I'm wrong, but we should have mutable/exclusive access to this lock at this point, no? 

Could we have unreachable on the error case here?

durable-storage/src/avl/resolver.rs
```diff
@@ -412,35 +422,65 @@ impl LazyDataId {
             return Ok(unsafe { &mut *temp });
         }
 
-        self.try_load_inner(key, store)?;
-        let bytes = self.0.inner.get_mut().expect("Try load succeeded");
-
-        Ok(bytes)
+        Err(OperationalError::ResolverInvariantViolated)
```
emturner: should this be `unreachable!`? It looks like the above `get_mut` could be flipped into a `let` expression, rather than an `if let`

durable-storage/src/merkle_layer.rs
emturner: at this point, in the third commit, we can assert that this matches the expected error case? Instead of having to duplicate an additional test down below

durable-storage/src/avl/resolver.rs
```diff
@@ -394,54 +401,93 @@ impl LazyDataId {
...
+    /// loaded separately from the KV store.
     ///
-    /// If this succeeds, the inner lazy-id is guaranteed to be
-    /// initialised.
-    fn try_load_inner(
+    /// Unchecked because the only caller that wants it that way is about to mutate the value, and
+    /// so has no committed hash left to check against. Every other load goes through
+    /// [`Self::try_load_inner`], which does check.
+    fn read(
         &self,
```
emturner, line 444: Perhaps then the name of the function could also be improved? 

As is - this looks very confusing for a reviewer. perhaps a `read_unchecked`? The section about 'only caller' feels misleading, as `try_load_inner` also uses this function

durable-storage/src/merkle_layer.rs
```diff
@@ -1055,6 +1055,172 @@ mod tests {
...
+            .get(&rhs)
+            .expect("Reading the source should succeed")
+            .expect("The source holds a value at `rhs`");
+
+        let mut read = vec![0u8; value.len()];
+        value.read(0, &mut read);
+        assert_eq!(read.as_slice(), b"value-2");
+
+        assert_eq!(merkle_0.hash(), merkle_2.hash());
+    });
```
emturner, line 1140: double check these various tests in `merkle_layer` aren't overlapping/checking identical things?

durable-storage/src/avl/node.rs
```diff
@@ -1022,6 +1029,53 @@ where
...
+///
+/// A node loaded from a store carries only the hash its value was committed with, and there is
+/// nothing to write for such a node: the store this layer resolves from already holds that value,
+/// and it is the store being written. `Borrow<[u8]>` cannot express "nothing to write" - it would
+/// have to produce bytes that are not there - so storing goes through this trait instead.
+///
+/// This makes `MerkleLayer::commit` writing the layer's own persistence a precondition rather than
+/// a coincidence: committing a checked-out layer into some other store would leave the values it
+/// never loaded behind.
+trait StorableData {
```
victor-dumitrescu: Nit: perhaps rename to `DataStorable` for consistency with existing `DataLoadable` (or the other way around)

durable-storage/src/merkle_layer.rs
```diff
@@ -1055,6 +1055,172 @@ mod tests {
...
+        assert!(
+            merkle.write(&key, 0, b"WITH").is_err(),
+            "Writing a value that cannot be loaded should fail"
+        );
+
+        // Still folds by the hash the node commits to, rather than panicking on an identifier
+        // left holding neither a hash nor a value.
+        assert_eq!(merkle.hash(), *commit.as_hash());
+    });
+
```
victor-dumitrescu, line 1239: `StorableData` + lazy loading introduce the possibility that values which are expected to be to be written to a store never are. For example, this (AI-generated) test which commits one key with `with_node_data`, checks the layer out without reading it, clones it onto a second empty store with `clone_with`, commits there, then checks that commit out of the second store and reads the key back. It passes on main but fails here.

```rust
    // `with_node_data` promises the nodes are persisted completely, and `clone_with` says the
    // clone commits to the store it is given. Together those say a committed layer's values are
    // readable from that store afterwards, whether or not this layer ever loaded them.
    kv_test!(test_commit_into_a_foreign_store_after_checkout, KV: PersistentKeyValueStore, {
        let key = Key::new(b"orphan").expect("Sizes less than KEY_MAX_SIZE");

        let options = super::StoreOptions::default().with_node_data();

        let source = Arc::new(KV::new(&repo).unwrap());
        let mut merkle = MerkleLayer::new(source.clone());
        merkle.set(&key, b"the value left behind").unwrap();

        let commit = merkle.commit(&options).expect("Committing should succeed");
        source.commit(&repo, &commit).unwrap();

        // Nothing reads this layer, so its values are never loaded from the store either.
        let checked_out_persistence = Arc::new(KV::checkout(&repo, &commit).unwrap());
        let checked_out = MerkleLayer::checkout(checked_out_persistence, commit)
            .expect("Checkout should succeed");

        // A store of its own, sharing nothing with the one the layer resolves from.
        let foreign = Arc::new(KV::new(&repo).unwrap());
        let mut copy = checked_out.clone_with(foreign.clone());

        assert_eq!(
            copy.commit(&options).expect("Committing into a foreign store should succeed"),
            commit
        );

        // The tree that store now holds commits to this value, so the store must hold it too.
        let committed = MerkleLayer::checkout(foreign, commit)
            .expect("The foreign store holds the tree, so it checks out");

        let value = committed
            .get(&key)
            .expect("Reading a committed value should succeed")
            .expect("The commit holds a value at `key`");

        let mut read = vec![0u8; value.len()];
        value.read(0, &mut read);
        assert_eq!(read.as_slice(), b"the value left behind");
    });

```

durable-storage/src/registry.rs
```diff
@@ -1183,7 +1183,7 @@ pub(super) mod tests {
         registry.verify_manifest(&root_commit, &expected_db_hashes);
     });
```
victor-dumitrescu, line 1185: Prove-mode `copy_database` can now fail: Destination gets source `ProveNodeIds` with unloaded values, so `into_proof` loads by key from the destination's separate store and fails. This (AI-generated) test passes on main but fails here:

```rust
    // A copy in prove mode hands the destination the source's working tree, and the values in it
    // are readable: the source resolves them, and the copy is what the destination now holds.
    kv_test!(test_prove_copy_after_checkout_reads_the_copied_value, KV: BackgroundPersistentKeyValueStore, {
        let mut registry = setup_size_2_registry::<KV>(repo.clone());

        let read_key = Key::new(&[2]).expect("Size less than KEY_MAX_SIZE");
        let probe_key = Key::new(&[1]).expect("Size less than KEY_MAX_SIZE");

        populate_database_with_key_value::<KV>(&mut registry, 0, &[1], b"probe");
        populate_database_with_key_value::<KV>(&mut registry, 0, &[2], b"read");

        let commit = registry.commit().expect("Commit should succeed");
        let checked_out = Registry::<KV, Normal>::checkout(repo, commit)
            .expect("Checkout should succeed");

        // Reads one key, which walks the node holding the other and leaves it cached.
        checked_out.databases[0].assert_database_value(&read_key, b"read");

        let mut prove = checked_out
            .try_start_proof()
            .expect("Starting a proof should succeed");

        prove.copy_database(0, 1).expect("Copying should succeed");

        prove
            .database(1)
            .expect("Database should exist.")
            .assert_database_value(&probe_key, b"probe");
    });

```

durable-storage/src/merkle_layer.rs
```diff
@@ -1055,6 +1055,172 @@ mod tests {
...
+        assert!(persistence.get(&key).is_err(), "The store should no longer hold the value");
+
+        assert!(
+            merkle.write(&key, 0, b"WITH").is_err(),
+            "Writing a value that cannot be loaded should fail"
+        );
+
+        // Still folds by the hash the node commits to, rather than panicking on an identifier
+        // left holding neither a hash nor a value.
+        assert_eq!(merkle.hash(), *commit.as_hash());
```
victor-dumitrescu, line 1225: This (AI-generated) test now fails in Prove mode too:

```suggestion
        assert_eq!(merkle.hash(), *commit.as_hash());
        
        let prove = merkle.start_proof();
        let _ = prove.get(&key);

        let proof = MerkleProof::from_foldable(&prove);
        assert_eq!(proof.root_hash(), *commit.as_hash());
```

durable-storage/src/merkle_layer.rs
emturner: I think this is absolutely fine and intended even. Yes, it's true that _fewer_ values will end up in the new store; but that was already the case, for every single node that was unloaded in the first place! Either way, if `foreign` is a brand new store, the committed + copied tree will be partial, it just slightly changes the 'shape' of the partialness

durable-storage/src/registry.rs
emturner: This is a known issue already - in that the way that the registry does not support multiple proof-steps, is _precisely_ the interaction between `copy/move` and database operations on the `copy/move`.

While this test succeeds on `main`, in fact - if you actually try and generate the full proof, you find that there are inconsistencies in the hashes anyway.

emturner: there's a comment above `Prove::copy_databse` that explains this too (with a `TODO TZX-173`

durable-storage/src/avl/node.rs
emturner: done in [diff](https://github.com/tezos/riscv-pvm/compare/5263a08dd7e4e69025534a89688f4b6abf527555..0c6c861c36184d645d570e5140acc2cf695099cb)

durable-storage/src/merkle_layer.rs
emturner: added a doc-comment covering this also, in [diff](https://github.com/tezos/riscv-pvm/compare/5263a08dd7e4e69025534a89688f4b6abf527555..0c6c861c36184d645d570e5140acc2cf695099cb)

durable-storage/src/merkle_layer.rs
emturner: good catch - technically it's ok (as `OperationalError` does semantically mean the state may be inconsistent. But given that we made it consistent in normal mode, worth doing in prove mode also.

See [diff](https://github.com/tezos/riscv-pvm/compare/5263a08dd7e4e69025534a89688f4b6abf527555..0c6c861c36184d645d570e5140acc2cf695099cb)

durable-storage/src/merkle_layer.rs
```diff
@@ -233,6 +233,12 @@ impl<KV> FromProof for MerkleLayer<KV, Verify> {
 
 impl<KV, M: MerkleLayerMode> MerkleLayer<KV, M> {
     /// Clone the Merkle layer. The new layer will commit to the provided persistence layer.
+    ///
+    /// The clone shares the source's tree, the values it has not loaded included - those are still
+    /// held as the hash they were committed with, and committing writes nothing for them. So
+    /// `persistence` has to already hold what the source never loaded: a clone of the source's
+    /// store does, an empty store does not, and committing into the latter leaves those values
+    /// behind.
```
victor-dumitrescu: I'm not sure I understand what the last part of this comment is trying to say.

victor-dumitrescu (approved):

durable-storage/src/merkle_layer.rs
emturner: reworded in [diff](https://github.com/tezos/riscv-pvm/compare/e398cddf5f9864c3d4494e97885b99682d8e403c..55adeaad84e738be1140ba8ecd4fbfbbf461b902)

ie, that if the merkle layer is a clone, for it to behave the same as the `src`, the persistence layer must also be a clone

PR #1198, opened 2026-09-03
feat(nds): omit overwritten pages from Bytes proofs

victor-dumitrescu:
Closes TZX-197.

# What

Set proofs no longer carry any page of the value being replaced, and Write proofs blind pages which are fully overwritten.

# Why

Optimise proof size. This data is not needed as it is replaced in full during Verify mode execution.

# How

- `set` is now a `BytesMode` method. In Prove mode it records the new data as one write over the whole array and records no resize boundary read. In Verify mode it replaces the data. When folding, `ProveImpl::page_presence` decides minimum presence for every page:
    - read pages are always `Present`
    - fully written pages are `MayBlind`, unless every other page is fully written, in which case they are all `MayOmit`
    - partially touched pages are `Present`
    - untouched pages are `MayOmit`
- `BytesMutOp::Set` added to Bytes proptests, and the Database regression trace gains a Set over an existing multi-page value; unit tests pin the proof shapes for set after a read and for whole-page versus partial writes.
- `proof_size.rs` is updated to reflect these changes

# Manually Testing

```
make all
```

# Regressions

The only changes are Set and Write proofs decreasing in size.

# Tasks for the Author


data/src/components/bytes.rs
```diff
@@ -890,6 +892,17 @@ impl BytesMode for Prove<'_> {
         len
     }
 
+    fn set(this: &mut Bytes<Self>, buffer: &[u8]) {
+        // Verification needs the length leaf in the proof: it derives the depth of the original
+        // page tree from it, in order to relate the tree it recomputes to the one the proof holds.
+        this.bytes.did_access_length.set(true);
```
emturner, line 900: is this strictly needed for `set`? Possibly it could be useful to have a follow up to remove the need to access length at all? 

I _think_ it's technically not required - as in theory we should be able to return the hash from just `blinded` - but might need additional support

durable-storage/src/test_helpers/proof_size.rs
```diff
@@ -23,6 +23,11 @@
 //! (an empty AVL child) is emitted as a present node rather than being blinded.
 //! Both encodings are shorter than a tag plus a digest, so the model stays
 //! conservative when it charges [`BLIND_LEAF`] for them.
+//!
+//! A page a `Write` covers in full is likewise blinded by the prover, but is
+//! still charged its content here. Tightening that would make the `Write` bound
+//! depend on how the write aligns with page boundaries; the looser charge is
+//! still an upper bound.
```
emturner, line 34: it could be possible, though, to update how we clamp the bytes component depending on the operation? (e.g. currently we clamp with `MAX_PROOF_SIZE` for bytes, but for writes we could clamp it with the tighter `MAX_WRITE_PROOF_SIZE` - but probably overkill

emturner (approved):

data/src/components/bytes.rs
victor-dumitrescu: Indeed it shouldn't be needed logically but it is currently. Opened a follow-up with an explanation and added todo [here](https://github.com/tezos/riscv-pvm/compare/52959fe5d8ff63078a4a1f861c968a4e960ea365..ed70e11c725359bce4d0354aa040f2f4901a783b).

durable-storage/src/test_helpers/proof_size.rs
victor-dumitrescu: Yeah my instinct was that also complicating the model to get a tighter upper bound for this specific case would be overkill.

emturner (approved):

PR #1199, opened 2026-09-03
feat(nds): page AVL keys so an ordering costs a page, not a key

emturner:
Closes TZX-189

# What

A key stored in an AVL node used to fold as a single leaf holding the whole key, so any proof
that touched a node carried that node's key in full. It now folds as a node of the key's `u8`
length and a tree of 64-byte pages, and comparisons are settled against as little of that as
possible:

- **Equality** is settled against the hash of the key as a whole, so the key stays blinded.
- **An ordering** reveals the single page on which the two keys differ. The pages before it are
  proven equal by the hashes of the subtrees spanning them, the pages after it never enter the
  comparison, and keys that agree on every page they share are ordered by their lengths.

# Why

The key was the dominant per-node cost of a search path, and it was paid in full at every node
the traversal stepped through, for information the comparison did not need. A proof only ever
needs enough of a key to justify the branch the traversal took - one page, and evidence that
the pages before it matched.

That cost is what bounds the tree: the deepest AVL tree whose proofs still fit
`DATABASE_MAX_PROOF_SIZE` is set by what each node on the search path costs, so shrinking the
key contribution deepens the tree the storage layer can hold. The binding operation is
unchanged - a max-chunk read or write over a 64 MiB value.

Measured at this commit, `key_ordering` charges a `KEY_MAX_SIZE` key on a search path 137 bytes
where the single leaf cost 257. Short keys pay slightly more than they did, the length leaf and
the page tree's nodes being fixed overhead a single leaf did not have.

# Impact on database sizes

That per-node saving is what deepens the tree. Minimal AVL node counts are Fibonacci, so each
level of depth is roughly 1.6x the keys, and the worst-case number of key/value pairs a database
can hold moves per operation as:

```
┌─────────────────────────────────────────┬───────────────────────┬──────────────────────────────┐
│                operation                │        before         │            after             │
├─────────────────────────────────────────┼───────────────────────┼──────────────────────────────┤
│ Delete (existing)                       │ d=50 → 53,316,291,171 │ d=63 → 27,777,890,035,286    │
├─────────────────────────────────────────┼───────────────────────┼──────────────────────────────┤
│ Set / Write (fresh insert)              │ d=45 → 4,807,526,974  │ d=71 → 1,304,969,544,928,655 │
├─────────────────────────────────────────┼───────────────────────┼──────────────────────────────┤
│ Read / Write (max chunk, 64 MiB value)  │ d=34 → 24,157,815     │ d=54 → 365,435,296,160       │
├─────────────────────────────────────────┼───────────────────────┼──────────────────────────────┤
│ Set (replacing a 64 MiB value)          │ d=46 → 7,778,742,047  │ d=73 → 3,416,454,622,906,705 │
├─────────────────────────────────────────┼───────────────────────┼──────────────────────────────┤
│ ValueLength / Exists / Delete (missing) │ d=46 → 7,778,742,047  │ d=72 → 2,111,485,077,978,048 │
├─────────────────────────────────────────┼───────────────────────┼──────────────────────────────┤
│ Hash                                    │ unbounded             │ unbounded                    │
├─────────────────────────────────────────┼───────────────────────┼──────────────────────────────┤
│ all operations                          │ d=34 → 24,157,815     │ d=54 → 365,435,296,160       │
└─────────────────────────────────────────┴───────────────────────┴──────────────────────────────┘
```

# How

Three commits, each self-contained:

1. `equality` checks can be settled by hash
2. turn `NodeKey` into a tree of length + `64 byte` pages. Equal pages can be blinded, so only one-or-zero pages must be included in all proofs
3. compare subtrees that are equal by hash, this ensures that, in the worst case, a `Key` only contains one page, one subtree hash, and one page hash (sibling of the included page). This is where the choice of `64 byte` pages comes from, as `32 byte` pages would force the inclusion of more 'sibling' hashes due to a second level in the tree

Verify mode keeps the page tree rather than flattening it to a list of pages, because a blinded
subtree's hash is exactly what the check needs and flattening loses it.

# Manually Testing

```
make all
```

# Regressions

Key hashes change, so the captures move with them. Which captures move, and why, differs per
commit:

- **Commit 1** touches only `*.proof-trace`. Recording equality as a comparison by hash lets the
  proof carry the key blinded, which changes the shape of the proof but no hash.
- **Commit 2** touches `*.trace`, `digests.out` and `*.proof-trace`. A key now folds as a node of
  its length and a page tree rather than as one leaf, so its hash changes, and every state hash
  derived from it changes with it.
- **Commit 3** touches no captures at all. It only changes which subtrees a proof blinds, and for
  the key shapes these captures exercise - keys that differ in their first page, none a
  page-aligned prefix of another - it marks exactly what commit 2 marked. The sampled keys were
  widened in this commit precisely because that was true of the property tests too.

*NB* not every proof becomes smaller (e.g. a key < 64 bytes now carries additional length/tag bytes), but the worst case is massively improved.

# Tasks for the Author


durable-storage/src/avl/key/tests.rs
```diff
@@ -134,22 +141,103 @@ fn verify_absent() {
     assert_eq!(cmp_value, None);
 }
 
-/// Test pinning NodeKey initial behaviour onto the same behaviour
-/// as Atom<Key>
+/// A blinded key still answers equality, in both directions, by hashing the key it is
+/// compared against.
 #[test]
-fn verify_blinded() {
+fn verify_blinded_eq() {
```
emturner, line 170: could be a proptest?

durable-storage/src/avl/key/tests.rs
```diff
@@ -134,22 +141,103 @@ fn verify_absent() {
...
-    let node_key = NodeKey::from_proof(ProofTree::present(&proof))
-        .expect("Should create absent NodeKey")
+    assert_eq!(catch_not_found(|| node_key.eq(&key)).ok(), Some(true));
+    assert_eq!(catch_not_found(|| node_key.eq(&other)).ok(), Some(false));
+}
+
+/// A blinded key can settle the [`Ordering::Equal`] case of a comparison, but not the
+/// orderings that need the key bytes.
+#[test]
+fn verify_blinded_cmp() {
```
emturner, line 185: ditto

durable-storage/src/avl/key/tests.rs
```diff
@@ -134,22 +141,103 @@ fn verify_absent() {
...
+
+    let key_normal = NodeKey::new(key.clone());
+
+    let key_prove = key_normal.start_proof();
+    assert!(!key_prove.eq(&other));
+
+    let merkle_proof = MerkleProof::from_foldable(&key_prove);
+
+    // The key is not in the proof - only its hash is.
+    assert_eq!(
```
emturner, line 281: surely the better place is to assert on the `merkle_proof` structure directly?

emturner: so serialisation would not be required

durable-storage/src/avl/key/tests.rs
```diff
@@ -134,22 +141,103 @@ fn verify_absent() {
...
+    assert_eq!(catch_not_found(|| key_verify.eq(&other)).ok(), Some(false));
+    assert_eq!(catch_not_found(|| key_verify.eq(&key)).ok(), Some(true));
+}
 
-    let cmp_value = catch_not_found(|| node_key.cmp(&key)).ok();
-    assert_eq!(cmp_value, None);
+/// A comparison that had to look at the key bytes keeps the key in the proof.
+#[test]
+fn unequal_cmp_keeps_the_key_in_the_proof() {
+    let key = Key::new(&[1; KEY_MAX_SIZE]).unwrap();
```
emturner, line 354: could be a proptest again?

durable-storage/src/avl/key/tests.rs
```diff
@@ -134,22 +141,103 @@ fn verify_absent() {
...
+
+    let key_normal = NodeKey::new(key.clone());
+
+    let key_prove = key_normal.start_proof();
+    assert_eq!(key_prove.cmp(&other), Ordering::Less);
+
+    let merkle_proof = MerkleProof::from_foldable(&key_prove);
+
+    assert!(
+        serialise(&merkle_proof)
```
emturner: ditto - assert on the proof directly, no?

durable-storage/src/avl/key.rs
```diff
@@ -7,6 +7,13 @@
 //! Specifically, this contains an optimised [`PartialOrd`] representation -
 //! that compares by hashes as much as possible (in [`Prove`] and [`Verify`] modes).
 //!
+//! A key's leaf encoding is length-prefixed, and therefore injective: two keys hash
```
emturner: this is restated in the commit message - which could be removed from the commit message entirely

durable-storage/src/avl/key.rs
```diff
@@ -72,9 +79,9 @@ impl<M: NodeKeyMode> NodeKey<M> {
 }
 
 impl<'normal> NodeKey<Prove<'normal>> {
-    /// Was the value accessed (read or written) during proof generation?
-    fn was_accessed(&self) -> bool {
-        self.key.read.get()
+    /// How was the key accessed during proof generation?
+    fn access(&self) -> KeyAccess {
```
emturner:
```suggestion
fn accessed(&self) ->  KeyAccess {
```
to make it clear this method doesn't access the key, but reports how it was?

durable-storage/src/avl/key/tests.rs
```diff
@@ -11,20 +11,54 @@ use octez_riscv_data::hash::PartialHash;
...
+
+/// Serialised size of the proof leaf holding a [`KEY_MAX_SIZE`] key's last page. It is the
+/// only page of that key that is not full, as [`KEY_MAX_SIZE`] is not a multiple of
+/// [`PAGE_SIZE`].
+const LAST_PAGE_LEAF_SIZE: usize = 1 + 1 + KEY_MAX_SIZE % PAGE_SIZE;
+
+/// Serialised size of a proof for the cheapest ordering of two [`KEY_MAX_SIZE`] keys: they
+/// differ in their first page, so the pages after it are never compared and the subtree
+/// holding them blinds away.
+const BEST_CASE_CMP_PROOF_SIZE: usize = 1
```
emturner: why is there a difference in proof size in first vs last page? surely the page tree is symmetric so it shouldn't matter?

durable-storage/src/avl/key/tests.rs
```diff
@@ -300,10 +302,10 @@ fn cmp_of_a_prefix_is_decided_by_length() {
     );
 }
 
-/// The comparison a proof was generated for is the only one it has to answer. A key that
-/// diverges from this one in a page the proof left out cannot be compared against it at
-/// all - not even for equality - and that is a missing-state failure rather than a wrong
-/// answer.
+/// The comparison a proof was generated for is the only one it has to order. A key that
```
emturner, line 291: I think the middle commit gets very confusing. It might be better to have whatever infrastructure required to support subtree hashing in place first, and then do the full key/cmp by subtree hash directly

durable-storage/src/avl/key/tests.rs
emturner: Done — a proptest over two arbitrary keys now. The unequal direction asserts `Some(key == other)` rather than filtering the equal case out, so the equal case is covered by the same run.

In `SQUASH ME -> settle AVL key equality against the key hash` (`qzyvqpno` / `3fed4245`), and again in `SQUASH ME -> prove a key's equal pages by their subtree hashes` (`kznwntrx` / `e8ea9670`) for the version this test grows into — the test is still keyed on `[1; 42]` at the tip of the branch too.

durable-storage/src/avl/key/tests.rs
emturner: Same treatment: proptest over two arbitrary keys, with the second comparison expecting `(key == other).then_some(Ordering::Equal)` — a blinded key settles `Equal` and nothing else.

durable-storage/src/avl/key/tests.rs
emturner: Agreed — and it comes out stronger than the size check, not just serialisation-free:

```rust
assert_eq!(
    merkle_proof,
    MerkleProof::leaf_blind(Hash::from_foldable(&key_normal))
);
```

The whole proof is that one hash, so `BLIND_LEAF_SIZE` goes away with it at this commit.

Checked it bisects: flipping `KeyAccess::Hashed` to `MinimumPresence::Present` fails this test and only this test.

At the tip of the branch the same two blinding tests (`eq_blinds_the_key_in_the_proof`, `equal_cmp_blinds_the_key_in_the_proof`) assert `merkle_proof == blinded_proof(&key)`. The three tests that assert a *size* there — early/late divergence and the prefix case — keep asserting the size, since the size is what they claim.

durable-storage/src/avl/key/tests.rs
emturner: Done — proptest over two keys filtered to be different, expecting `key.cmp(&other)` rather than a hardcoded `Ordering::Less`.

durable-storage/src/avl/key/tests.rs
emturner: Also done, and structurally rather than by size:

```rust
let leaf = <Key as LeafEncode<Bincode>>::leaf_encode(&key)?;
prop_assert_eq!(&merkle_proof, &MerkleProof::leaf_read(leaf));
```

That pins the proof to *exactly* the key's leaf, where the old assertion only said "longer than `KEY_MAX_SIZE` bytes".

Bisects: setting `KeyAccess::Read` to `MayBlind` fails it (and `node_key_cmp_verifies_correctly` with it).

durable-storage/src/avl/key.rs
emturner: Dropped from the commit message — the module doc is where it belongs. The first paragraph becomes:

> `NodeKey` now records an equality comparison as a comparison by hash rather than
> as a read, which lets the proof carry the key blinded, and `Verify` mode answers
> `eq` against a blinded key by hashing the key it is compared against.

The rest of the message is unchanged. I've left this one to you rather than rewriting the commit.

durable-storage/src/avl/key.rs
emturner: Applied, with the doc saying the same thing rather than asking a question:

```rust
/// How the key was accessed during proof generation - this does not itself
/// count as an access.
fn accessed(&self) -> KeyAccess {
```

Worth noting it is short-lived: paging records an access per page, so there is no whole-key accessor left by the next commit.

durable-storage/src/avl/key/tests.rs
emturner: The page tree is symmetric; the *access pattern* at this commit is not.

A divergence in the first page leaves pages 2-4 untouched, so the aligned subtree spanning them is never marked at all and collapses into a single blind leaf. The pages *before* a divergence are not untouched — they have to be proven equal — and at this commit each of them is blinded individually, so a late divergence pays a blind leaf per page plus the nodes above them that stay in the proof. That is the whole difference: 137 vs 170 bytes here, all of it "one hash per page" against "one hash for the subtree".

It is also exactly what the next commit removes: with the equal pages left omittable, the subtrees spanning them blind as units and the two cases come out at 137 and 136 bytes. The remaining byte is not structural either — `KEY_MAX_SIZE` is 255, which is not a multiple of `PAGE_SIZE`, so a late divergence reveals a 63-byte last page where an early one reveals a full 64-byte page.

Which does mean the constant's doc at the tip is wrong the other way round: `LATE_DIVERGENCE_CMP_PROOF_SIZE` is described as "the dearest a comparison gets" when it is a byte *cheaper* than an early divergence. Fixed in `kznwntrx` — it now says the cost is set by the page tree's depth, which is the same wherever the keys diverge.

durable-storage/src/avl/key/tests.rs
emturner: Agreed, and I've built it so you can see it rather than take my word for it: `emturner@key-paging-restructured` (tip `nqnrrkmn` / `93954a0b`), sitting on top of commit 1 — nothing of yours is rewritten, commits 2 and 3 are still where they are.

It is two commits instead of three, and the split is file-level:

1. `feat(data): expose the layout of an aligned chunk of an IndexableSeqAsTree` (`yutzppss` / `3d69240f`) — `data/src/foldable/seq_tree.rs` only: `chunk`, its argument checks, and its tests. Green on its own (12 `seq_tree` tests).
2. `feat(nds): page AVL keys so an ordering costs a page, not a key` (`nqnrrkmn` / `93954a0b`) — all of `durable-storage`, with subtree hashing from the start. Byte-identical tree to the current tip plus the test fixes from this review.

What it buys, beyond your point about the middle commit:

- The captures move **once**, in one commit, so the per-commit regression story in the description collapses to "commit 1 touches `*.proof-trace`, commit 2 moves the hashes".
- No commit ever describes the per-page blinding that only exists between commits 2 and 3 — which is also what made [the proof-size asymmetry above](https://github.com/tezos/riscv-pvm/pull/1199#discussion_r3932671670) confusing to read.
- The `chunk` argument-validation rationale sits on the commit that adds `chunk`.

The message on 2 is a merge of your two, minus the paragraphs about the interim state. Take it or leave it — if you'd rather keep the three commits, the review fixes are in the `SQUASH ME` pair (`qzyvqpno`, `kznwntrx`) and are independent of this branch.

durable-storage/src/avl/key/tests.rs
emturner: Taken further, per your follow-up: at the tip there is no longer a size to distinguish. There is one constant, `CMP_PROOF_SIZE` — what an ordering costs, set by the page tree's depth and so the same wherever the keys diverge — and `cmp_diverging_late_proves_the_earlier_pages_as_subtrees` now asserts that diverging in the last page costs no more than diverging in *any* other page, rather than asserting a size of its own:

```rust
let last = divergence_proof_size(pages - 1);

for page in 0..pages - 1 {
    assert!(
        last <= divergence_proof_size(page),
        "diverging in the last page should cost no more than diverging in page {page}"
    );
}
```

`LATE_DIVERGENCE_CMP_PROOF_SIZE` and `LAST_PAGE_LEAF_SIZE` are gone with it, and so is the "which of these is the dearest" question the two constants invited.

The assertion is not vacuous: the sizes across the four pages of a `KEY_MAX_SIZE` key are `[137, 137, 137, 136]`. Under the middle commit's per-page blinding the last would be 170, so it fails there — which is why the middle commit keeps both constants, now with a doc saying the page tree is symmetric and the asymmetry is only in which pages get blinded and how.

durable-storage/src/avl/key.rs
```diff
@@ -7,28 +7,55 @@
...
+//! rather than as one leaf holding the whole key. That shape lets a comparison
+//! put only what it needs into the proof:
+//!
+//! - An equality check is settled against the hash of the key as a whole, which
+//!   lets the key be blinded outright.
+//! - Two different keys are ordered within the first page on which they differ:
+//!   the subtrees before it are equal, which the verifier re-checks against the
+//!   subtree hash, and the subtree after it is never compared. Only that
+//!   one page has to be fully included in the proof, so an ordering costs a that
+//!   page, plus the hash of a precending/subsequent subtree/page.
```
victor-dumitrescu, line 20:
```suggestion
//!   one page has to be fully included in the proof, so an ordering costs
//!   a page, plus the hash of a preceeding/subsequent subtree/page.
```

durable-storage/src/avl/key.rs
```diff
@@ -282,32 +750,127 @@ impl private::NodeKeyImpl for Normal {
...
+    Read,
+}
+
+impl KeyAccess {
+    /// The presence the proof must give a leaf accessed this way.
+    fn presence(self) -> MinimumPresence {
+        match self {
+            KeyAccess::None => MinimumPresence::MayOmit,
+            KeyAccess::Hashed => MinimumPresence::MayBlind,
+            KeyAccess::Read => MinimumPresence::Present,
```
victor-dumitrescu, line 776: `KeyAccess` is equivalent to `MinimumPresence`, I guess the reason to introduce it is readability for the rest of this module?

durable-storage/src/test_helpers/database.rs
```diff
@@ -131,9 +132,38 @@ pub fn make_database_operations(
...
+/// reaches the subtree hashes that prove the pages before a divergence equal, nor the branches
+/// that order keys agreeing on every page they share. Half the samples are therefore built from
+/// a fixed run of whole pages, so that keys drawn independently of one another still agree page
+/// for page - and, where the remainder is empty, are page-aligned prefixes of one another.
+///
+/// The empty key gets an arm of its own rather than being left to turn up in the length range,
+/// where it would be one draw in `KEY_MAX_SIZE`. Nothing rejects it - [`Key`] validation only
+/// bounds the length from above - and it is the one key with no page for a comparison to mark,
+/// so its page tree blinds whole rather than opening down to a page.
 pub(crate) fn key_strategy() -> impl Strategy<Value = Key> {
```
victor-dumitrescu, line 147: Might be good to reset regressions so they also capture this.

durable-storage/src/avl/key.rs
```diff
@@ -125,83 +244,432 @@ impl<'normal> Provable<'normal> for NodeKey<Normal> {
...
+
+    let mut node = proof.into_node()?;
+
+    match node.presence() {
+        Partial::Absent => return node.done(PageTree::Absent),
+        Partial::Blinded(hash) => {
+            return node.done(PageTree::Blinded {
+                hash,
+                hash_chunk: chunk_hash::<C>,
+            });
```
victor-dumitrescu, line 563: Minor, but seeing how this is duplicated in `leaf` might be worth pulling into a separate function?

durable-storage/src/avl/key.rs
```diff
@@ -282,32 +750,127 @@ impl private::NodeKeyImpl for Normal {
...
+        // Only the boundary page is marked. The path down to it stays in the
+        // proof, which leaves the subtrees spanning the pages before it blinded as
+        // units - one hash to check per layer rather than one per page.
+        match first_differing_page(bytes, rhs.as_ref()) {
+            Some(page) => self.record_page(page, KeyAccess::Read),
+
+            // The keys agree on every shared page, so their lengths order them -
+            // but the shared pages still have to be checkable, and hashing the
+            // first page past them holds the path open.
+            None if shared < pages => self.record_page(shared, KeyAccess::Hashed),
```
victor-dumitrescu, line 832:
```suggestion
            None if shared < pages && shared > 0 => self.record_page(shared, KeyAccess::Hashed),
```
I think? Slight optimisation for when one of the keys is empty.

durable-storage/src/avl/key.rs
```diff
@@ -317,41 +880,164 @@ impl<'normal> private::NodeKeyImpl for Prove<'normal> {
...
-            inner: Partial::Present(key),
-        };
+        let bytes = key.as_ref();
 
-        NodeKey { key: verify_impl }
+        NodeKey {
+            key: VerifyImpl::Tree {
+                length: Partial::Present(key_length(&key)),
+                pages: PageTree::present(bytes, Chunk::root(page_count(bytes.len()))),
+            },
```
victor-dumitrescu, line 1020:
```suggestion
            key: VerifyImpl::Full(key)
```

You don't really need to page the key when it's full, could instead store it whole as a new variant, which would also remove the need for `PageTree::present`. `key_hash` will still hash it as fi it were paged, which I think is enough?

emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/70af4c0cc6ea0bcb44680f535f017a2ef438b3b4..52019ebb95b9610f6962a53d247beb473f8f1606)

durable-storage/src/avl/key.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/70af4c0cc6ea0bcb44680f535f017a2ef438b3b4..52019ebb95b9610f6962a53d247beb473f8f1606)

durable-storage/src/avl/key.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/70af4c0cc6ea0bcb44680f535f017a2ef438b3b4..52019ebb95b9610f6962a53d247beb473f8f1606)

durable-storage/src/test_helpers/database.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/70af4c0cc6ea0bcb44680f535f017a2ef438b3b4..52019ebb95b9610f6962a53d247beb473f8f1606)

durable-storage/src/avl/key.rs
emturner: Yes exactly, it's mainly aimed at readability - rather than introducing any alternate variants.

Explained in [diff](https://github.com/tezos/riscv-pvm/compare/70af4c0cc6ea0bcb44680f535f017a2ef438b3b4..52019ebb95b9610f6962a53d247beb473f8f1606)

durable-storage/src/avl/key.rs
emturner: done [diff](https://github.com/tezos/riscv-pvm/compare/70af4c0cc6ea0bcb44680f535f017a2ef438b3b4..52019ebb95b9610f6962a53d247beb473f8f1606)

victor-dumitrescu (approved):

PR #1207, opened 2026-09-08
feat(nds): garbage collect commit directories older than a given registry commit

emturner:
Part of TZX-209

# What

Record the order of commits made, in an append only journal. Then allow for commits to be collected that are older than a given root (and not referenced since). Finally, bound the registry long test with that collection, in place of the copy of it the test kept.

# Why

`octez-smart-rollup-node` needs the storage to reclaim disk, and today it reclaims
nothing: a repository grows with the total number of commits over its lifetime.

Currently, a commit id is just the registry root hash, so contains no information about when the commit occurred. Two unrelated commits could share the same id, and ancestry is unavailable.

Simply removing directories removes a large proportion of what a repository holds. As commit directories are sets of hard links, removing the last directory containing a given hard link frees it. This is sufficient for garbage collection of the 'key-value' column family.

# How

## Commit Journal 

Each committed root is recorded as a sequence number that increases with
every commit. Identity is untouched, so committing the same state twice still produces
the same id and the same manifest; only the journal grows. Retention follows the recorded
order rather than ancestry, which keeps states whose provenance is unclear. Collecting at
a root keeps every root recorded at or after it, and a root committed more than once
keeps its highest position, so re-committing an old state can only extend how long it
survives. The target is read the other way about: its *earliest* position is the floor,
so collecting at a state that has been committed twice drops only what preceded the first
time it was reached, and never the states committed in between. Both directions err
towards keeping a state.

## Collection 

Database commits are content-addressed and therefore shared — unrelated
registry commits name the same one whenever their states agree — so their removal follows
reachability rather than the order commits were made in.

It is safe to interrupt and repeat. It prunes the journal to the retained roots, removes
the manifests it dropped, then removes the unreachable database commits, in that order, so
everything retained is intact at each point in between. Each step after the first
enumerates what is present rather than what the journal says should be, so a round left
unfinished is picked up by the next one.

Both sit on the repository trait, so the in-memory backend records and collects the same
way the on-disk one does.

For now, collection must not overlap a commit or another collection; that is the caller's
responsibility for now, and is stated in the docs page.

## Long Test Retention

The registry long test advances a shared base once per epoch, then runs a proptest of many
cases against that base. A case checks the base out and applies operations with proofs; it
never commits. Only the epoch advance does. So the only commits a run records are epoch
bases, in the order the epochs ran.

Bounding the repository over a long run therefore means dropping old epoch bases, and the
test used to do that with its own copy of collection: a window of recent bases, and on
eviction a reachability walk over the retained manifests, then database directory and
manifest removals, mirrored by hand onto both backends. That is the mechanism this PR
ships, written a second time, so the retention a run exercised was never quite the
retention that ships — and the two could drift apart without anything noticing.

Now the window is all the test keeps. It trims to the `keep` most recent bases and calls
`collect()` at the oldest one still in the window, once per backend. Retention by recorded
order coincides with a window over epoch bases exactly because nothing between them is ever
recorded, so "every root recorded at or after this one" and "the bases still in the window"
are the same set.

Three things follow from the swap. A long randomised run now exercises the real
`collect()`, including the journal pruning the copy never touched. The bound is tighter,
because the copy only ever considered the database commits of the base it was evicting,
leaving anything else unreachable on disk. And it is the only place the in-memory
repository's side of collection runs at all — see the note under Manually Testing.

Ordering is safe: collection must not overlap a commit, and the call sits between epochs,
after the proptest for that epoch has finished, so nothing else is touching the repository.

# Manually Testing

```
make -C durable-storage check
make -C durable-storage test
```

The space harness now drives the real `collect()` rather than a copy of it, so the
reclaim figures it reports measure what ships:

```
cargo run --features unstable-test-utils --bin gc_space -- --keys 200000 --commits 20 --collect
```

So does the registry long test, on both backends:

```
make -C durable-storage run-registry-long-test
```

That run is also the only coverage the in-memory repository has for its side of collection.
The `collect` tests reach the trait through `DirectoryManager`, and the crate's two test
flavours are the same build in this respect — the self dev-dependency takes default
features, and `default = ["rocksdb"]`, so `--no-default-features` does not turn rocksdb off
for the test binary — which left nothing running `InMemoryRepo`'s journal and removal
methods.

# Regressions

None. Nothing is deleted by the first commit, and the second only removes commit
directories and manifests that are no longer retained.

# Tasks for the Author

### DIFF
--- a/durable-storage/src/bin/gc_space.rs
+++ b/durable-storage/src/bin/gc_space.rs
@@ -75,11 +75,13 @@ struct Cli {
     #[arg(long)]
     json_out: Option<PathBuf>,
 
-    /// After the last commit, delete every commit not reachable from it and measure again.
+    /// After the last commit, collect at it and measure again.
+    ///
+    /// Collection at the final commit deletes all commits not reachable from it.
     ///
     /// Shows what collecting at directory granularity reclaims, and what dead node data survives it.
     #[arg(long)]
-    simulate_dir_gc: bool,
+    collect: bool,
 }
 
 fn main() -> Result<()> {
@@ -114,7 +116,7 @@ fn main() -> Result<()> {
         seed: cli.seed,
         repo_dir: cli.repo_dir,
         json_out: cli.json_out,
-        simulate_dir_gc: cli.simulate_dir_gc,
+        collect: cli.collect,
     }
     .run()
 }
--- a/durable-storage/src/collect.rs
+++ b/durable-storage/src/collect.rs
@@ -1,0 +1,352 @@
+// SPDX-FileCopyrightText: 2026 Trilitech <contact@trili.tech>
+//
+// SPDX-License-Identifier: MIT
+
+//! Reclaiming the commits a repository no longer needs.
+//!
+//! Collecting at a root drops every registry commit recorded before it, together with the database
+//! commits none of the surviving registry commits still reference. Database commits are shared -
+//! they are content-addressed, so unrelated registry commits name the same one whenever their
+//! states agree - which is why removal follows reachability rather than the order commits were
+//! made in.
+//!
+//! This reclaims the value side of a commit. A database commit directory is a set of hard links,
+//! so unlinking the last directory that references a file frees it, and values are keyed by user
+//! key, so overwrites become obsolete versions that compaction discards. Merkle nodes are
+//! content-addressed and every version of a node is a distinct live key, so no amount of directory
+//! removal reclaims them; that needs deletion of the node keys themselves and is a separate
+//! mechanism.
+//!
+//! # Interruption
+//!
+//! Collection is safe to interrupt and repeat. It runs in three steps, ordered so that everything
+//! retained is intact at every point in between:
+//!
+//! 1. prune the journal to the retained roots,
+//! 2. remove the manifests of registry commits that were dropped,
+//! 3. remove the database commits no retained manifest reaches.
+//!
+//! Pruning first is what makes a repeat safe rather than merely possible. A target older than one
+//! already collected at is no longer in the journal, so it is refused instead of being treated as
+//! a floor that retains states whose data has already gone. Each later step enumerates what is
+//! actually present rather than what the journal says should be, so work left unfinished by an
+//! interruption is picked up by the next round.
+//!
+//! # Concurrency
+//!
+//! Collection must not run while the repository is being committed to, and two collections must
+//! not overlap. Neither is enforced here.
+
+use std::collections::HashSet;
+
+use crate::commit::CommitId;
+use crate::errors::GcError;
+use crate::errors::OperationalError;
+use crate::journal;
+use crate::registry;
+use crate::repo::RegistryRepo;
+
+/// What a collection round reclaimed.
+#[derive(Debug, Default, PartialEq, Eq, Clone, Copy)]
+pub struct Collected {
+    /// Registry commits whose manifests were removed.
+    pub registry_commits: usize,
+
+    /// Database commits that were removed.
+    pub database_commits: usize,
+}
+
+/// Drop everything the repository no longer needs to serve `target` and the commits after it.
+///
+/// Retains every registry commit recorded at or after `target`, and every database commit one of
+/// those still references. Fails with [`CollectionTargetNotRecorded`] if the journal holds no
+/// entry for `target`, which is also how a target that an earlier round already collected past is
+/// refused.
+///
+/// [`CollectionTargetNotRecorded`]: crate::errors::GcArgumentError::CollectionTargetNotRecorded
+pub fn collect<Repo: RegistryRepo>(repo: &Repo, target: &CommitId) -> Result<Collected, GcError> {
+    let retained_roots = journal::roots_to_retain(&repo.read_commit_journal()?, target)?;
+
+    // Read before anything is removed. These manifests are all retained, so they would survive the
+    // steps below either way, but reading them first keeps the reachable set independent of how
+    // far a previous interrupted round got.
+    let reachable = reachable_database_commits(repo, &retained_roots)?;
+
+    repo.prune_journal(&retained_roots)?;
+
+    let mut collected = Collected::default();
+
+    for id in repo.registry_commits()? {
+        if retained_roots.contains(&id) {
+            continue;
+        }
+
+        repo.remove_registry_commit(&id)?;
+        collected.registry_commits += 1;
+    }
+
+    for id in repo.database_commits()? {
+        if reachable.contains(&id) {
+            continue;
+        }
+
+        repo.remove_database_commit(&id)?;
+        collected.database_commits += 1;
+    }
+
+    Ok(collected)
+}
+
+/// The database commits referenced by any of `roots`.
+///
+/// A root with no manifest is skipped rather than failing the round. A commit writes its manifest
+/// before recording its root, so an interruption leaves a manifest with no root rather than the
+/// other way about; what leaves a root with no manifest is a caller that recorded one itself
+/// without committing it. Skipping it keeps that from holding up the commits that do have
+/// manifests.
+fn reachable_database_commits<Repo: RegistryRepo>(
+    repo: &Repo,
+    roots: &HashSet<CommitId>,
+) -> Result<HashSet<CommitId>, OperationalError> {
+    let mut reachable = HashSet::new();
+
+    for root in roots {
+        match registry::database_commits(repo, root) {
+            Ok(databases) => reachable.extend(databases),
+            Err(OperationalError::CommitNotFound) => continue,
+            Err(error) => return Err(error),
+        }
+    }
+
+    Ok(reachable)
+}
+
+#[cfg(all(test, rocksdb_test_utils))]
+mod tests {
+    use bytes::Bytes;
+    use octez_riscv_data::mode::Normal;
+    use octez_riscv_test_utils::TestableTmpdir;
+
+    use super::*;
+    use crate::errors::GcArgumentError;
+    use crate::key::Key;
+    use crate::persistence_layer::PersistenceLayer;
+    use crate::registry::Registry;
+    use crate::repo::DirectoryManager;
+
+    /// A repository holding a registry of one database, committed once per call to `commit`.
+    struct Fixture {
+        _tmp: TestableTmpdir,
+        repo: DirectoryManager,
+        registry: Registry<PersistenceLayer, Normal>,
+    }
+
+    impl Fixture {
+        fn new() -> Self {
+            let tmp = TestableTmpdir::new();
+            let repo = DirectoryManager::new(tmp.path())
+                .expect("creating the directory manager should succeed");
+
+            let mut registry = Registry::<PersistenceLayer, Normal>::new(repo.clone());
+            registry
+                .resize_tick(1)
+                .expect("resizing the registry should succeed");
+
+            Self {
+                _tmp: tmp,
+                repo,
+                registry,
+            }
+        }
+
+        /// Set `key` to `value` in the registry's first database and commit.
+        fn commit(&mut self, key: &[u8], value: &[u8]) -> CommitId {
+            self.commit_to(0, key, value)
+        }
+
+        /// Set `key` to `value` in database `index` and commit.
+        fn commit_to(&mut self, index: usize, key: &[u8], value: &[u8]) -> CommitId {
+            let key = Key::new(key).expect("the key should be valid");
+            self.registry
+                .database_mut(index)
+                .expect("the database should exist")
+                .set(key, Bytes::copy_from_slice(value))
+                .expect("setting should succeed");
+
+            self.registry.commit().expect("committing should succeed")
+        }
+
+        fn registry_commits(&self) -> HashSet<CommitId> {
+            self.repo
+                .registry_commits()
+                .expect("listing registry commits should succeed")
+                .into_iter()
+                .collect()
+        }
+
+        fn database_commits(&self) -> HashSet<CommitId> {
+            self.repo
+                .database_commits()
+                .expect("listing database commits should succeed")
+                .into_iter()
+                .collect()
+        }
+    }
+
+    // Collecting at the newest root keeps it and drops the ones before it, and the retained root
+    // still checks out - which needs its database commits to have survived.
+    #[test]
+    fn collecting_at_the_newest_root_drops_the_rest() {
+        let mut fixture = Fixture::new();
+
+        let first = fixture.commit(b"a", b"1");
+        let second = fixture.commit(b"a", b"2");
+        let third = fixture.commit(b"a", b"3");
+
+        let collected = collect(&fixture.repo, &third).expect("collection should succeed");
+
+        assert_eq!(collected.registry_commits, 2);
+        assert_eq!(fixture.registry_commits(), HashSet::from([third]));
+        assert!(!fixture.registry_commits().contains(&first));
+        assert!(!fixture.registry_commits().contains(&second));
+
+        // Each commit rewrote the same key, so each has a database commit of its own and the two
+        // dropped roots take theirs with them.
+        assert_eq!(collected.database_commits, 2);
+
+        let retained: HashSet<CommitId> = registry::database_commits(&fixture.repo, &third)
+            .expect("reading the retained manifest should succeed")
+            .into_iter()
+            .collect();
+        assert_eq!(
+            fixture.database_commits(),
+            retained,
+            "the database commits left should be exactly the ones the retained root names"
+        );
+
+        Registry::<PersistenceLayer, Normal>::checkout(fixture.repo.clone(), third)
+            .expect("the retained root should still check out");
+    }
+
+    // A database commit shared with a retained registry commit is kept, even though the registry
+    // commit that also referenced it was dropped. Committing the registry without touching the
+    // database leaves the database commit unchanged, so both roots name it.
+    #[test]
+    fn a_shared_database_commit_survives() {
+        let mut fixture = Fixture::new();
+
+        let first = fixture.commit(b"a", b"1");
+        let shared: HashSet<CommitId> = registry::database_commits(&fixture.repo, &first)
+            .expect("reading the manifest should succeed")
+            .into_iter()
+            .collect();
+
+        // Write to a second database, so the registry root changes while the first database is
+        // untouched and keeps the commit id the dropped root referenced.
+        fixture
+            .registry
+            .resize_tick(2)
+            .expect("resizing should succeed");
+        let second = fixture.commit_to(1, b"b", b"2");
+
+        assert!(
+            registry::database_commits(&fixture.repo, &second)
+                .expect("reading the manifest should succeed")
+                .iter()
+                .any(|id| shared.contains(id)),
+            "the second root should still reference the first database's commit"
+        );
+
+        collect(&fixture.repo, &second).expect("collection should succeed");
+
+        let present = fixture.database_commits();
+        for id in &shared {
+            assert!(
+                present.contains(id),
+                "a database commit a retained root references should survive"
+            );
+        }
+
+        Registry::<PersistenceLayer, Normal>::checkout(fixture.repo.clone(), second)
+            .expect("the retained root should still check out");
+    }
+
+    // Collecting at the oldest root is a no-op, since everything is at or after it.
+    #[test]
+    fn collecting_at_the_oldest_root_removes_nothing() {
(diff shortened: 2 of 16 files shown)


durable-storage/src/storage/in_memory.rs
```diff
@@ -441,4 +445,31 @@ impl crate::repo::RegistryRepo for InMemoryRepo {
...
+
+        Ok(seq)
+    }
+
+    fn commit_journal(&self) -> Result<Vec<crate::journal::JournalEntry>, OperationalError> {
+        Ok(self
+            .journal
+            .read()
+            .map_err(|_| OperationalError::LockPoisoned)?
+            .clone())
```
emturner:
```suggestion
       self
            .journal
            .read()
            .map(Clone::clone)
            .map_err(|_| OperationalError::LockPoisoned)
```

durable-storage/src/journal.rs
```diff
@@ -0,0 +1,215 @@
...
+
+/// The roots to keep when collecting at `target`.
+///
+/// Every root recorded at or after `target` is retained, `target` itself included. Fails with
+/// [`OperationalError::CollectionTargetNotRecorded`] if the journal holds no entry for `target`,
+/// since without one there is no floor to compare against and collecting would drop everything.
+pub fn roots_to_retain(
+    entries: &[JournalEntry],
+    target: &CommitId,
+) -> Result<HashSet<CommitId>, OperationalError> {
```
emturner: possible nit - one might argue that `OperationalError` here for `CollectionTargetNotRecorded` should instead be some sort of `GcError::InvalidArgument(GcArgumentError)` instead?

durable-storage/src/registry.rs
```diff
@@ -235,6 +235,10 @@ where
             .repo
             .write_registry_commit(&registry_commit, &encoded)?;
 
+        // Recorded after the manifest, so that an interrupted commit leaves a manifest collection
+        // can reclaim rather than a recorded root with nothing behind it.
```
```suggestion
        // Recorded after the manifest, so that an interrupted commit leaves a manifest that collection
        // can reclaim, rather than a recorded root with nothing behind it.
```

durable-storage/src/repo.rs
```diff
@@ -112,6 +164,16 @@ pub trait RegistryRepo: Clone {
...
+
+    /// Record `root` as the most recently committed registry root.
+    ///
+    /// Called after the manifest is written, so a crash in between leaves an unrecorded manifest
+    /// rather than a recorded root with nothing behind it. The commit id was never returned to the
+    /// caller in that case, so nothing can reference it and collection is free to reclaim it.
+    fn record_commit(&self, root: &CommitId) -> Result<Seq, OperationalError>;
+
+    /// Every commit recorded by [`RegistryRepo::record_commit`], in the order they were recorded.
+    fn commit_journal(&self) -> Result<Vec<JournalEntry>, OperationalError>;
```
emturner:
```suggestion
fn read_commit_journal(&self) -> Result<Vec<JournalEntry>, OperationalError>;
```

perhaps? `commit_journal` sounds like the journal is being committed

durable-storage/src/repo.rs
```diff
@@ -131,4 +193,190 @@ impl RegistryRepo for DirectoryManager {
...
+            .truncate(false)
+            .open(self.journal_file())
+            .map_err(|error| OperationalError::FileWriteFailed { error })?;
+
+        // Write at the end of the last *whole* entry rather than at the end of the file. A torn
+        // write leaves a partial record that reading already ignores, but appending past it would
+        // put every later entry out of alignment permanently: the journal is read as fixed-width
+        // chunks from the start, so one stray byte turns the boundary between two real entries
+        // into a fabricated third and drops the tail. Truncating to the new end discards the torn
+        // bytes in the same step.
```
emturner:
```suggestion
        // Write at the end of the last *whole* entry rather than at the end of the file. A torn
        // write leaves a partial record that reading already ignores, but appending past it would
        // put every later entry out of alignment permanently.
```

durable-storage/src/storage/in_memory.rs
```diff
@@ -32,6 +32,10 @@ pub struct InMemoryRepo {
```
emturner, line 32: update commit message:

```
Collection needs to know which commits are newer than the one it is asked to
collect at. That cannot be read off the commits themselves: a commit id is the
registry root hash, so it says what a state contains and nothing about when it
was reached.

A repository now records each committed root in a journal beside the manifests,
as a sequence number that increases with every commit. Retention follows the 
recorded order rather than
ancestry. Collecting at a root therefore keeps every root recorded
at or after it, and a root committed more than once keeps its highest position,
so re-committing an old state extends how long it survives.

Deletion/collection follows in subsequent commits.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
```

durable-storage/src/bin/gc_space.rs
```diff
@@ -75,11 +75,11 @@ struct Cli {
     #[arg(long)]
     json_out: Option<PathBuf>,
 
-    /// After the last commit, delete every commit not reachable from it and measure again.
+    /// After the last commit, collect at it and measure again.
```
emturner, line 78:
```suggestion
/// After the last commit, collect at it and measure again.
///
/// Collection at the final commit deletes all commits not reachable from it.
```

durable-storage/src/gc_space/prune.rs
```diff
@@ -7,94 +7,37 @@
...
 ///
-/// This is collection at directory granularity: the reachable database commits are read from the
-/// retained registry manifest, and every other commit directory and manifest is removed. Blocks
-/// come back only where no surviving link remains, which is exactly the point — the dead node data
-/// inside the retained commit survives this untouched, because the retained commit still needs the
-/// files holding it.
+/// The collection itself is [`collect`], so this measures the shipped implementation rather than a
+/// stand-in for it. Blocks come back only where no surviving link remains, which is exactly the
+/// point — the dead node data inside a retained commit survives untouched, because that commit
+/// still needs the files holding it.
```
emturner:
```suggestion
/// Blocks come back only where no surviving link remains: the dead node data inside a retained commit survives untouched, because that commit
/// still needs the files holding it.
```

durable-storage/src/repo.rs
```diff
@@ -101,6 +116,45 @@ impl DirectoryManager {
...
+        let len = journal
+            .metadata()
+            .map_err(|error| OperationalError::FileReadFailed { error })?
+            .len();
+
+        // A trailing partial entry is a torn write, so the last whole entry is the one before it.
+        let whole = len / journal::ENTRY_BYTES as u64;
+        let Some(last) = whole.checked_sub(1) else {
+            return Ok(None);
+        };
```
victor-dumitrescu: Nit, mostly for readability: Should this be a separate `open_journal` function?

durable-storage/src/journal.rs
```diff
@@ -0,0 +1,215 @@
...
+/// Entries are fixed width so that the journal can be appended to without rewriting it, and so
+/// that a torn write at the end of the file is recognisable by length alone.
+pub(crate) const ENTRY_BYTES: usize = SEQ_BYTES + Hash::DIGEST_SIZE;
+
+/// Bytes used by the sequence number within an encoded entry.
+const SEQ_BYTES: usize = size_of::<u64>();
+
+impl JournalEntry {
+    /// Encode this entry as [`ENTRY_BYTES`] bytes.
+    pub(crate) fn encode(&self) -> [u8; ENTRY_BYTES] {
```
victor-dumitrescu, line 73: Are sequence numbers actually required to be part of the journal serialisation? Unless I'm missing something, it seems to me they're always the position of a commit id in the journal.

If there is a requirement for sequence numbers not to be reused after a collection, then storing only the first entry's sequence number and deriving the rest should work? Although at that point the more complicated logic would probably not be worth it.

durable-storage/src/long_test/registry/gc.rs
victor-dumitrescu, line 1: The dummy GC loop from the long tests should be replaced with an actual GC run, but perhaps as a follow up?

durable-storage/src/storage/in_memory.rs
```diff
@@ -441,4 +432,73 @@ impl crate::repo::RegistryRepo for InMemoryRepo {
         commits.insert(*id, bytes.to_vec());
         Ok(())
     }
+
+    fn record_commit(
```
victor-dumitrescu, line 436: The in-memory implementations aren't currently tested because all the tests in `collect.rs` only run as persistence layer tests.

durable-storage/src/collect.rs
```diff
@@ -0,0 +1,337 @@
...
+        collected.database_commits += 1;
+    }
+
+    Ok(collected)
+}
+
+/// The database commits referenced by any of `roots`.
+///
+/// A root with no manifest is skipped rather than failing the round. That happens when a commit
+/// was interrupted after its root was recorded but before its manifest was written, and when a
```
victor-dumitrescu: But the manifest is written before the commit is recorded right?

durable-storage/src/journal.rs
```diff
@@ -0,0 +1,215 @@
...
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Hash)]
+pub struct Seq(u64);
+
+impl Seq {
+    /// The sequence number given to the first commit a repository records.
+    pub const FIRST: Self = Self(0);
+
+    /// The sequence number after this one.
+    pub fn next(self) -> Self {
+        Self(self.0 + 1)
```
victor-dumitrescu: Obviously this would never overflow in practice, but perhaps it might if we're working off some kind of corrupted state. In that case, perhaps better to fail explicitly?

durable-storage/src/repo.rs
```diff
@@ -131,4 +225,260 @@ impl RegistryRepo for DirectoryManager {
...
+        let seq = match self.last_journal_entry()? {
+            Some(last) => last.seq.next(),
+            None => Seq::FIRST,
+        };
+
+        let entry = JournalEntry { seq, root: *root };
+
+        let mut journal = std::fs::OpenOptions::new()
+            .create(true)
+            .write(true)
```
victor-dumitrescu: Very minor, but the journal ends up being opened twice in a row: first above when reading `last_journal_entry` then here.

durable-storage/src/collect.rs
```diff
@@ -0,0 +1,337 @@
...
+    fn collecting_at_the_newest_root_drops_the_rest() {
+        let mut fixture = Fixture::new();
+
+        let first = fixture.commit(b"a", b"1");
+        let second = fixture.commit(b"a", b"2");
+        let third = fixture.commit(b"a", b"3");
+
+        let collected = collect(&fixture.repo, &third).expect("collection should succeed");
+
+        assert_eq!(collected.registry_commits, 2);
```
victor-dumitrescu, line 208: Should this test also check that the expected number of database commits was removed?

durable-storage/src/repo.rs
emturner: Done — [`open_journal`](https://github.com/tezos/riscv-pvm/blob/458c33835cf9a23f273d76ea34271bb5a36d4af4/durable-storage/src/repo.rs#L123-L129) now returns the handle, the last whole entry and the offset just past it, and [`record_commit`](https://github.com/tezos/riscv-pvm/blob/458c33835cf9a23f273d76ea34271bb5a36d4af4/durable-storage/src/repo.rs#L202-L205) is its only caller. That also removes the second open you spotted further down, so the two comments are one change.

[diff](https://github.com/tezos/riscv-pvm/compare/5ec82e6f13f99c9785213681b70632900457fc10..458c33835cf9a23f273d76ea34271bb5a36d4af4)

durable-storage/src/repo.rs
emturner: Fixed by the same change as the `open_journal` comment above: one `read(true).write(true)` open, the tail read through that handle, the entry written at the offset it hands back — [`open_journal`](https://github.com/tezos/riscv-pvm/blob/458c33835cf9a23f273d76ea34271bb5a36d4af4/durable-storage/src/repo.rs#L123-L129), [`record_commit`](https://github.com/tezos/riscv-pvm/blob/458c33835cf9a23f273d76ea34271bb5a36d4af4/durable-storage/src/repo.rs#L202-L215).

[diff](https://github.com/tezos/riscv-pvm/compare/5ec82e6f13f99c9785213681b70632900457fc10..458c33835cf9a23f273d76ea34271bb5a36d4af4)

durable-storage/src/journal.rs
emturner: You're right about this commit: as it stands a seq *is* the entry's index, and nothing here would notice if it were derived.

What needs it is the follow-up that collects. Pruning the journal drops entries and shifts everything after them, so an index stops being stable — while the number has to be, because the sweep compares stamps written under an earlier journal against a floor read from the current one. Renumbering there does not corrupt anything, but it makes stale stamps sit above new floors, so the sweep concludes nodes are still held and quietly stops reclaiming. Storing the first seq and deriving the rest goes the same way, and additionally breaks once a position is read before a commit that then fails: positions order commits and need not be contiguous.

The doc line was claiming the opposite of all this ("nothing outside it should persist one"), so it now says what is actually guaranteed: [journal.rs#L24-L31](https://github.com/tezos/riscv-pvm/blob/458c33835cf9a23f273d76ea34271bb5a36d4af4/durable-storage/src/journal.rs#L24-L31).

While in here, one behavioural change worth a second look since you were reading this function: [`roots_to_retain`](https://github.com/tezos/riscv-pvm/blob/458c33835cf9a23f273d76ea34271bb5a36d4af4/durable-storage/src/journal.rs#L124-L159) now takes the floor from the target's **earliest** position rather than its latest, so a target committed more than once collects from the first time it was reached and the states committed between its two positions are no longer dropped. Other roots are still judged by their latest position, so both directions err towards keeping a state.

[diff](https://github.com/tezos/riscv-pvm/compare/5ec82e6f13f99c9785213681b70632900457fc10..458c33835cf9a23f273d76ea34271bb5a36d4af4)

durable-storage/src/journal.rs
emturner: Made deliberate rather than incidental — it wraps, with the reasoning written down: [`Seq::next`](https://github.com/tezos/riscv-pvm/blob/458c33835cf9a23f273d76ea34271bb5a36d4af4/durable-storage/src/journal.rs#L37-L46). Reaching the wrap takes 2^64 commits or a journal corrupt enough to decode a number near it, and wrapping to zero costs nothing that matters: zero is at or below every position already recorded, so collecting at the wrapped commit retains everything. The failure mode is a round that reclaims nothing, not one that drops a state it should have kept.

Worth saying why not a hard failure: the journal has no integrity check at all, so a corrupt seq is the least of what a corrupt entry can do — a corrupt *root hash* is undetectable today and much worse. If we want to fail explicitly on a torn record, the thing to add is a checksum per entry, and trapping here would follow from that rather than standing on its own.

[diff](https://github.com/tezos/riscv-pvm/compare/5ec82e6f13f99c9785213681b70632900457fc10..458c33835cf9a23f273d76ea34271bb5a36d4af4)

durable-storage/src/collect.rs
emturner: It is, and the doc was wrong. `Registry::commit` writes the manifest and *then* records the root, so an interruption leaves a manifest with no root — never a root with no manifest.

Rewritten to say that, and to name the case that does produce one: a caller that records a root itself without committing it. [collect.rs#L100-L106](https://github.com/tezos/riscv-pvm/blob/2b8c75bb900106585120e4cf693f076516181c73/durable-storage/src/collect.rs#L100-L106). The inverse shape does occur in the long test, where `restore_base` registers manifests with `write_registry_commit` and never records them — collection reclaims those, which is what we want.

[diff](https://github.com/tezos/riscv-pvm/compare/24a0a20e935519f5115959b79c2b485a13a5aed6..2b8c75bb900106585120e4cf693f076516181c73)

durable-storage/src/collect.rs
emturner: Added — the count, and that what survives is exactly what the retained root names: [collect.rs#L213-L225](https://github.com/tezos/riscv-pvm/blob/2b8c75bb900106585120e4cf693f076516181c73/durable-storage/src/collect.rs#L213-L225). The three commits rewrite the same key, so each has a database commit of its own and two go with the dropped roots.

Worth having: no test asserted that a round removes *any* database commit, so a regression that made that loop a no-op passed the whole suite. Confirmed the new assertion catches it — disabling the removal fails it at `left: 0, right: 2`.

[diff](https://github.com/tezos/riscv-pvm/compare/24a0a20e935519f5115959b79c2b485a13a5aed6..2b8c75bb900106585120e4cf693f076516181c73)

durable-storage/src/long_test/registry/gc.rs
emturner: Done here rather than as a follow-up, since it turned out to be the answer to your in-memory comment as well: new commit `test(nds): retain the long test's snapshots with collect()`.

**Why the window translates.** A run advances the shared base once per epoch and then runs a proptest of many cases against that base. A case checks the base out, applies operations with proofs and compares roots — it never commits; only the epoch advance does, on both backends, asserting the two commit ids agree. So the only commits a run records are epoch bases, in epoch order. "Every root recorded at or after this one" and "the bases still in the window" are therefore the same set, which is what makes a window over epochs expressible as a single `collect()` call.

**What it replaces.** `prune` kept a `VecDeque` of recent bases and, on eviction, did its own reachability walk over the retained manifests, then removed the evicted base's unreferenced database directories and its manifest, mirroring both onto the in-memory repo by hand. That is the mechanism this PR ships, written a second time — so the retention a long run exercised was never quite the retention that ships, and the two could drift without anything failing.

**What it is now.** Trim to the `keep` most recent bases, then collect at the oldest one still in the window, once per backend:

```rust
while recent_commits.len() > keep.get() {
    recent_commits.pop_front();
}

let Some(oldest) = recent_commits.front() else {
    return Ok(());
};

collect(persistent_repo, oldest).context("collecting the persistent repository")?;
collect(in_memory_repo, oldest).context("collecting the in-memory repository")?;
```

Three consequences worth naming. A long randomised run now exercises the real `collect()`, including `prune_journal`, which the copy never touched at all. The bound is tighter, because the copy only considered the database commits of the base it was evicting and left anything else unreachable sitting on disk. And it is the only place `InMemoryRepo`'s side of collection runs — see the reply on `in_memory.rs`.

Ordering is safe: collection must not overlap a commit, and the call sits between epochs in the `Ok` branch, after that epoch's proptest has finished, so nothing else is touching the repository.

The tests either side of it: `prunes_unreachable_snapshots` now also asserts that both journals were pruned, so neither backend will collect at the evicted base again, while the retained base still checks out fully on both — which requires the database commits it shares with the evicted one to have survived. `keep_epochs_bounds_the_repo` still pins exactly one manifest on disk across a four-epoch run.

One thing I did not change: `restore_base` on the failure-reproduction path registers manifests with `write_registry_commit` and never records them, so a restored base is not a collection target. It never runs in the same loop as `prune`, but it is the reason the "root with no manifest" case in `collect.rs` exists at all.

The whole commit is the change, so: [diff](https://github.com/tezos/riscv-pvm/pull/1207/commits/ee4d1005948c26ab855eb52e1245deb4ce70bdc6)

durable-storage/src/storage/in_memory.rs
emturner: Correct, and the reason is worse than it looks: the crate's two test flavours are the same build. The self dev-dependency

```toml
octez-riscv-durable-storage = { path = ".", features = ["unstable-test-utils"] }
```

takes default features and `default = ["rocksdb"]`, so `--no-default-features` does not turn rocksdb off for the test binary: `rocksdb_test_utils` is set either way, and both flavours list an identical 242 tests. The `test` target's comment in `durable-storage/Makefile` claims otherwise and is inaccurate. So `InMemoryRepo`'s journal and removal methods were not merely untested by `collect.rs` — they had no coverage anywhere.

Covered now by driving the long test's retention through `collect()` on both repositories, plus an assertion that both journals were pruned, which is what exercises `read_commit_journal`/`prune_journal` on the in-memory side: [gc.rs#L119-L129](https://github.com/tezos/riscv-pvm/blob/ee4d1005948c26ab855eb52e1245deb4ce70bdc6/durable-storage/src/long_test/registry/gc.rs#L119-L129).

The whole commit is the change, so: [diff](https://github.com/tezos/riscv-pvm/pull/1207/commits/ee4d1005948c26ab855eb52e1245deb4ce70bdc6)

victor-dumitrescu (approved):

